PR #7, opened 2017-01-26
DAO & Mock Registry

eriknelson:
* Implemented some basic dao features for talking with etcd
* Filled out a mock-registry as a separate binary so we have something to pull data from in a real way.

### DIFF
--- a/pkg/ansibleapp/dev_registry.go
+++ b/pkg/ansibleapp/dev_registry.go
@@ -23,8 +23,8 @@ func (r *DevRegistry) Init(config RegistryConfig, log *logging.Logger) error {
 	return nil
 }
 
-func (r *DevRegistry) LoadApps() ([]*Spec, error) {
-	r.log.Debug("RHCCRegistry::LoadApps")
+func (r *DevRegistry) LoadSpecs() ([]*Spec, error) {
+	r.log.Debug("RHCCRegistry::LoadSpecs")
 
 	appsUrl := r.fullAppsPath()
 
--- a/pkg/ansibleapp/dev_registry.go
+++ b/pkg/ansibleapp/dev_registry.go
@@ -1,20 +1,63 @@
 package ansibleapp
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"github.com/op/go-logging"
+	"net/http"
+	"strings"
 )
+
+const APPS_PATH = "/ansibleapps"
 
 type DevRegistry struct {
 	config RegistryConfig
+	log    *logging.Logger
 }
 
-func (r *DevRegistry) Init(config RegistryConfig) error {
+func (r *DevRegistry) Init(config RegistryConfig, log *logging.Logger) error {
+	log.Debug("RHCCRegistry::Init")
 	r.config = config
-	fmt.Printf("DevRegistry::Init with url -> [ %s ] \n", r.config.Url)
+	r.log = log
 	return nil
 }
 
-func (r *DevRegistry) LoadApps() error {
-	fmt.Println("DevRegistry::LoadApps ")
-	return nil
+func (r *DevRegistry) LoadApps() ([]*Spec, error) {
+	r.log.Debug("RHCCRegistry::LoadApps")
+
+	appsUrl := r.fullAppsPath()
+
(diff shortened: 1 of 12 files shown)


pkg/ansibleapp/dev_registry.go
```diff
@@ -1,20 +1,63 @@
...
+	log    *logging.Logger
 }
 
-func (r *DevRegistry) Init(config RegistryConfig) error {
+func (r *DevRegistry) Init(config RegistryConfig, log *logging.Logger) error {
+	log.Debug("RHCCRegistry::Init")
 	r.config = config
-	fmt.Printf("DevRegistry::Init with url -> [ %s ] \n", r.config.Url)
+	r.log = log
 	return nil
```
jmrodri: Why return anything if we're just going to return nil?

pkg/ansibleapp/registry.go
```diff
@@ -22,7 +31,10 @@ func CreateRegistry(config RegistryConfig) Registry {
...
 	}
 
-	reg.Init(config)
+	err := reg.Init(config, log)
+	if err != nil {
+		return nil, err
+	}
 
-	return reg
+	return reg, err
```
jmrodri: Looks good.

pkg/ansibleapp/rhcc_registry.go
```diff
@@ -1,20 +1,22 @@
...
 	return nil
 }
 
-func (r *RHCCRegistry) LoadApps() error {
-	fmt.Println("RHCCRegistry::LoadApps ")
-	return nil
+func (r *RHCCRegistry) LoadApps() ([]*Spec, error) {
+	r.log.Debug("RHCCRegistry::LoadApps")
+	return []*Spec{}, nil
 }
```
jmrodri: I do like the pluggable registry mechanism.

pkg/app/app.go
```diff
@@ -1,24 +1,37 @@
...
+	registry ansibleapp.Registry
 }
 
 func CreateApp() App {
 	var err error
+
+	fmt.Println("============================================================")
+	fmt.Println("==           Starting Ansible Service Broker...           ==")
+	fmt.Println("============================================================")
+
```
jmrodri: NICE formatting. :tada: :trophy: :medal_sports:

pkg/app/app.go
```diff
@@ -36,11 +49,28 @@ func CreateApp() App {
...
+		os.Exit(1)
+	}
+
 	return app
 }
 
 func (a *App) Start() {
-	a.log.Info("Starting application")
+	a.log.Notice("Ansible Service Broker Started")
+	a.registry.LoadApps()
```
jmrodri: What if this fails?

pkg/dao/dao.go
```diff
@@ -1,10 +1,108 @@
...
+	}
+
+	return nil
+}
+
+////////////////////////////////////////////////////////////
+// Key generators
+////////////////////////////////////////////////////////////
+func specKey(id string) string {
+	return fmt.Sprintf("/spec/%s", id)
```
jmrodri: This file looks good too.

pkg/ansibleapp/dev_registry.go
eriknelson: Won't compile without it

eriknelson: I expect real registries to be doing more interesting things with failure possibility; this guys just kind of a test stub for us for now.

pkg/app/app.go
eriknelson: Good catch; I actually should remove this line for this PR. I'm picturing this start method being the one responsibly for starting the http mux listener. That will call into methods like `broker.Bootstrap()` (not written yet), which *in turn*, will be the guy that calls `registry.LoadApps()`.

So the full story has yet to be written here. This PR should be a lot of support without action in `Start()`. I'll try to make a note of this in `//TODOs` to explain myself a little better inline.

pkg/ansibleapp/dev_registry.go
jmrodri: Oh I get it now. The interface signature requires an error for REAL registries. For the dev one we don't care. Ok I forgot about that. +1

jmrodri:
ACK

PR #8, opened 2017-01-26
broker::Bootstrap & broker simplification

eriknelson:
* Adds `Bootstrap` method for loading specs from a registry and persisting them
* Simplifies broker organization until we need to start splitting things off


eriknelson:
Closing in favor of #10, which contains this as well as `broker::Catalog`.

PR #12, opened 2017-01-27
Provision support

eriknelson:
* `broker::Provision` implementation
* `ServiceInstance` type for instance bookkeeping
* Dao support for ServiceInstances
* Made {Dump,Load}JSON methods more generic for reuse across model types

### DIFF
--- a/pkg/ansibleapp/types.go
+++ b/pkg/ansibleapp/types.go
@@ -2,8 +2,10 @@ package ansibleapp
 
 import (
 	"encoding/json"
+	"github.com/pborman/uuid"
 )
 
+type Parameters map[string]interface{}
 type SpecManifest map[string]*Spec
 
 type Spec struct {
@@ -16,18 +18,22 @@ type Spec struct {
 	Async string `json:"async"`
 }
 
-func (s *Spec) LoadJSON(payload string) *Spec {
-	json.Unmarshal([]byte(payload), s)
-	return s
-}
-
-func (s *Spec) DumpJSON() string {
-	payload, err := json.Marshal(s)
+func LoadJSON(payload string, obj interface{}) error {
+	err := json.Unmarshal([]byte(payload), obj)
 	if err != nil {
-		panic(err)
+		return err
 	}
 
-	return string(payload)
+	return nil
+}
+
+func DumpJSON(obj interface{}) (string, error) {
+	payload, err := json.Marshal(obj)
+	if err != nil {
+		return "", err
+	}
+
+	return string(payload), nil
 }
 
 func NewSpecManifest(specs []*Spec) SpecManifest {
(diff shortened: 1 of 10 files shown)


test/catalog.sh
```diff
@@ -5,4 +5,5 @@
 curl \
   -H 'X-Broker-API-Version: 2.9' \
   -v \
-  http://cap.example.com:8000/v2/catalog
+  http://localhost:1338/v2/catalog
```
jmrodri: That's not a very l33t port.

pkg/ansibleapp/types.go
```diff
@@ -2,32 +2,45 @@ package ansibleapp
 
 import (
 	"encoding/json"
+	"github.com/pborman/uuid"
 )
 
+type Parameters map[string]interface{}
```
eriknelson: Reading this blog, `map[string]interface{}` is the generic type go uses for JSON objects in that the entire object should still be accessible by normal golang means: https://blog.golang.org/json-and-go

"The json package uses map[string]interface{} and []interface{} values to store arbitrary JSON objects and arrays; it will happily unmarshal any valid JSON blob into a plain interface{} value."

This type allows us to access stuff inside of it, as opposed to an opaque string blob.

pkg/ansibleapp/types.go
```diff
@@ -2,32 +2,45 @@ package ansibleapp
...
 type SpecManifest map[string]*Spec
 
+type AnswerDescriptor struct {
+	Name        string      `json:"name"`
+	Description string      `json:"description"`
+	Default     interface{} `json:"default"`
```
eriknelson: It could be a string, I debated this briefly when writing this and chose `interface{}` because of things like ports. This will serialize correctly into a json blob also, so it will show `{"default": 400}` instead of `{"default": "400"}`.

I can update if we don't see much of a benefit here, it just feels more correct?

pkg/ansibleapp/types_test.go
```diff
@@ -16,17 +16,20 @@ const SpecDescription = "A note taking webapp"
...
 {
-  "id": "%s",
+	"id": "%s",
 	"description": "%s",
-  "name": "%s",
-  "bindable": %t,
-  "async": "%s"
+	"name": "%s",
+	"bindable": %t,
+	"async": "%s"
```
eriknelson: Editor was complaining about mismatched indentation. It saw 2 spaces here in a file that is tabbed (since I copied this in from a json file where I had 2 space tab settings). Updated here for uniformity.

test/catalog.sh
eriknelson: I normally use 1337, but it's taken by my mock-registry :P

pkg/ansibleapp/types.go
jmrodri: Excellent source. Ok makes sense.

pkg/ansibleapp/types.go
jmrodri: Nope makes sense.

pkg/ansibleapp/types_test.go
jmrodri: Ok, makes sense. Usually these types of changes are mistakes, but this one was on purpose. +1

test/catalog.sh
jmrodri: So this is l33t +1 :) that's geeky enough. Accepted.

jmrodri:
ACK

PR #23, opened 2017-02-17
fix flake8, _headers wasn't doing anything

jmrodri:
flake8 said ```_header``` was unused, but I think it was to save the existing header values update them then replace with the modified set. If not, I can fix it.


eriknelson:
@jmrodri I screwed that up from an earlier iteration. Will push an update to this branch and merge. Thanks for the formatting!

eriknelson:
Used this branch for the demo, no problems. ACK

PR #28, opened 2017-02-24
Broker ansibleapp

eriknelson:
* Container assets
* Updated confs
* Properly parameterize targeted openshift cluster
* asbcli support for deploying broker

**NOTE:** This is a breaking change for asbcli, uses cli interface outlined in README.md


eriknelson:
Related: https://github.com/fusor/ansibleapp-library/pull/20

PR #29, opened 2017-02-27
Move ansibleapp files to service broker repo

eriknelson:
Moving the ansibleapp / Docker files here instead of ansibleapp-library so we can make sure new images are built for dockerhub on tagged updates of this repo.

* Also added a "--version" option to the command line so it will line up with tags.


eriknelson:
Related https://github.com/fusor/ansibleapp-library/pull/21

jmontleon:
ACK

PR #51, opened 2017-03-24
append version to setup command in rpm spec

jmontleon:


jmontleon:
I was able to reproduce the mock-scm process locally. The directory in the archive is %{name}-%{version}.

tar zxvf ../SOURCES/ansible-service-broker.tar.gz 
ansible-service-broker-0/

PR #57, opened 2017-03-30
Validate we are connected to etcd by checking the version

rthallisey:
We don't check if we are connected to etcd until we make a request.
Instead, after we make the etcd client we can ask for the version
to verify we have a valid connection.


rthallisey:
https://github.com/fusor/ansible-service-broker/pull/61

PR #60, opened 2017-04-04
[DO NOT MERGE] Exec into the container to get bind creds

rthallisey:
Instead of getting bind credentials from stdout, the broker will exec into the container and get them.

More details in this patch: https://github.com/fusor/ansible-playbook-bundle-library/pull/1

### DIFF
--- a/pkg/apb/bind.go
+++ b/pkg/apb/bind.go
@@ -51,22 +51,19 @@ func Bind(
 	log.Debug("Calling getPodName")
 	podname, _ := getPodName(output, log)
 	log.Debug("Calling monitorOutput on " + podname)
-	bindout, _ := monitorOutput(podname)
-	log.Info(string(bindout))
 
-	// HACK ALERT!
-	bd, err := buildBindData(bindout)
-	if err != nil {
-		// HACK: this is HORRIBLE. but there is definitely a time between a bind
-		// and when the container is up.
-		time.Sleep(5 * time.Second)
-		log.Warning("Trying to monitor the output again on " + podname)
-		bindout, _ = monitorOutput(podname)
-		log.Info(string(bindout))
-		return buildBindData(bindout)
+	out := make(chan []byte)
+	go func() {
+		monitorOutput(podname, out, log)
+	}()
+
+	var bindout []byte
+	for msg := range out {
+		log.Info(string(msg))
+		bindout = msg
 	}
 
-	return bd, err
+	return buildBindData(bindout)
 }
 
 // HACK: this really is a crappy way of getting output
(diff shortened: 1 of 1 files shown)


pkg/apb/bind.go
```diff
@@ -86,8 +83,24 @@ func getPodName(output []byte, log *logging.Logger) (string, error) {
...
-	return runCommand("oc", "logs", "-f", podname)
+func monitorOutput(podname string, mon chan []byte, log *logging.Logger) {
+	// TODO: Error handling here
+	// It would also be nice to gather the script output that exec runs
+	// instead of only getting the credentials
+	retries := 5
+	for r := 1; r <= retries; r++ {
+		output, _ := runCommand("oc", "exec", podname, "broker-bind-creds")
+
+		if strings.Contains(string(output), "BIND_CREDENTIALS") {
```
jmrodri: Didn't `Contains` existed. Not sure how I missed that before. +1

pkg/apb/bind.go
```diff
@@ -86,8 +83,24 @@ func getPodName(output []byte, log *logging.Logger) (string, error) {
...
+			mon <- output
+			close(mon)
+			return
+		}
+
+		time.Sleep(time.Duration(r) * time.Second)
+		log.Warning("Retry attempt %d: exec into %s failed", r, podname)
```
jmrodri: Good log.

pkg/apb/bind.go
```diff
@@ -51,22 +51,19 @@ func Bind(
...
-		return buildBindData(bindout)
+	out := make(chan []byte)
+	go func() {
+		monitorOutput(podname, out, log)
+	}()
+
+	var bindout []byte
+	for msg := range out {
+		log.Info(string(msg))
+		bindout = msg
```
jmrodri: So we monitor the output in a go routine, then we wait for there to be something in the channel to parse. I like it.

jmrodri: If the monitor fails to write anything to the channel, what happens? Does this hang? does it quit?

rthallisey: If the monitor writes nothing to the channel we'll hang.

pkg/apb/bind.go
```diff
@@ -86,8 +83,24 @@ func getPodName(output []byte, log *logging.Logger) (string, error) {
...
+		if strings.Contains(string(output), "BIND_CREDENTIALS") {
+			mon <- output
+			close(mon)
+			return
+		}
+
+		time.Sleep(time.Duration(r) * time.Second)
+		log.Warning("Retry attempt %d: exec into %s failed", r, podname)
+	}
+	return
```
rthallisey: Need to send a timeout error message to the channel and ```close(mon)```. Return here isn't enough.

jmrodri:
@rthallisey we're going to hold off on this PR because of the work @eriknelson has been doing recently things have changed. I want to keep this open because we want some of this work but it will not be a clean merge.

rthallisey:
No rush.  I'll rebase when we're ready to merge.

jmrodri:
Now I don't know how this will be affected with all the new `ext_creds.go` refactor we did earlier.

jmrodri:
@rthallisey rebase this and we can take another look at it. We also are getting bind credentials from the output of a **provision** call now.

eriknelson:
Because credentials can now come out of a provision or a bind call, we need to chat about the various scenarios that could arise and how to react to each. Ex:

```
Credentials returned:
prov: T, bind: F -> pass back prov creds from bind, we do this today.
prov: F, bind: T -> pass back bind creds, basically the same as the above.
prov: T, bind: T -> First, is this okay? If so, how do we handle this? Merge one onto the other? Which 
  takes precendence if there is a key conflict?
prov: F, bind F -> Is this okay? I lean no. If a bind is basically defined by the return of credentials, if 
  there are none available, this is bad. Can we think of a case where a bind makes sense that 
  *doesn't* return anything to the consuming application? NOTE: This currently segfaults. Needs to get 
   fixed and tested against our decision here.
```

rthallisey:
Ya let's revisit this method.

eriknelson:
Related #104

rthallisey:
https://github.com/fusor/ansible-service-broker/pull/124

PR #64, opened 2017-04-06
ExtractedCredentials Refactor

eriknelson:
Bind tests are all broken, need updating before merge.

### DIFF
--- a/pkg/apb/bind.go
+++ b/pkg/apb/bind.go
@@ -1,14 +1,6 @@
 package apb
 
 import (
-	"encoding/base64"
-	"encoding/json"
-	"errors"
-	"fmt"
-	"regexp"
-	"strings"
-	"time"
-
 	logging "github.com/op/go-logging"
 )
 
@@ -21,11 +13,11 @@ func Bind(
 	instance *ServiceInstance,
 	parameters *Parameters,
 	clusterConfig ClusterConfig, log *logging.Logger,
-) (*BindData, error) {
+) (*ExtractedCredentials, error) {
 	log.Notice("============================================================")
 	log.Notice("                       BINDING                              ")
 	log.Notice("============================================================")
-	log.Notice(fmt.Sprintf("Parameters: %v", parameters))
+	log.Notice("Parameters: %v", parameters)
 	log.Notice("============================================================")
 
 	var client *Client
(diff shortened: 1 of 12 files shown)


pkg/apb/bind.go
```diff
@@ -21,11 +13,11 @@ func Bind(
 	instance *ServiceInstance,
 	parameters *Parameters,
 	clusterConfig ClusterConfig, log *logging.Logger,
-) (*BindData, error) {
+) (*ExtractedCredentials, error) {
```
jmrodri: acceptable name :smile:

pkg/apb/bind.go
```diff
@@ -21,11 +13,11 @@ func Bind(
...
 	log.Notice("============================================================")
 	log.Notice("                       BINDING                              ")
 	log.Notice("============================================================")
-	log.Notice(fmt.Sprintf("Parameters: %v", parameters))
+	log.Notice("Parameters: %v", parameters)
```
jmrodri: do we not need the `fmt.Sprint` anymore?

pkg/apb/bind_test.go
```diff
@@ -1,13 +1,13 @@
...
 import (
-	"fmt"
 	"testing"
 
 	ft "github.com/fusor/ansible-service-broker/pkg/fusortest"
 )
 
-func TestDecodeOutput(t *testing.T) {
+func TestBind(t *testing.T) {
+	t.Skip("skipping bind until we can pass in a mock client")
```
jmrodri: I'm purposely skipping this test because Bind uses a docker client that we need to mock. Would be nice to be able to pass in a mock client and set an expectation of what commands should get passed to runCommand, etc.

pkg/apb/ext_creds.go
```diff
@@ -0,0 +1,119 @@
...
+
+	log.Debug("Calling getPodName")
+	podname, _ := getPodName(output, log)
+	log.Debug("Calling monitorOutput on " + podname)
+	bindout, _ := monitorOutput(podname)
+	log.Info(string(bindout))
+
+	var creds *ExtractedCredentials
+	var err error
+	creds, err = buildExtractedCredentials(bindout)
```
jmrodri: this made it hard to test because the output given to the extractCredentials is only used to get the podname. then we run oc get logs and use THAT output. So we really need a way to mock out the runCommand to have it return testable output. For now, I tested buildExtractedCredentials separately.

pkg/apb/ext_creds.go
```diff
@@ -0,0 +1,119 @@
...
+		startIdx = strings.Index(str, "<BIND_ERROR>")
+		startOffset := startIdx + len("<BIND_ERROR>")
+		endIdx := strings.Index(str, "</BIND_ERROR>")
+		if startIdx > -1 && endIdx > -1 {
+			// Case 3, error reported
+			return nil, errors.New(str[startOffset:endIdx])
+		}
+
+		// Case 1, no creds found, no errors occurred
+		return nil, nil
```
jmrodri: we already talked about this.

pkg/apb/provision.go
```diff
@@ -14,37 +15,47 @@ import (
 func Provision(
 	spec *Spec, parameters *Parameters,
 	clusterConfig ClusterConfig, log *logging.Logger,
-) error {
+) (*ExtractedCredentials, error) {
```
jmrodri: man that feels dirty :)

eriknelson: Formatting wise? I've been favoring this when signatures start to blow up, kinda like it :)

pkg/apb/bind.go
eriknelson: Nope, realized this recently so I'm updating these when I come across them.

pkg/apb/provision.go
jmrodri: hah, no returning credentials from provision :) hehe.

PR #67, opened 2017-04-07
Convert YAML to JSON before using JSON tags

rthallisey:
First translate YAML to JSON before unmarshaling to a structure with JSON tags. This is less prone to error.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 89d00acd05cddf68463ac64bdfb0511be82b906e2aa3e3855c7edd97dfce8884
-updated: 2017-05-18T07:10:44.527708169-04:00
+hash: 9fb582bdb1e6fa2a3d56603f7d184dc1bac46188924b40fb5bcea300ee270a9a
+updated: 2017-06-05T15:07:10.74174955-04:00
 imports:
 - name: github.com/Azure/go-ansiterm
   version: 70b2c90b260171e829f1ebd7c17f600c11858dbe
@@ -61,7 +61,7 @@ imports:
   - storage
   - storageversion
 - name: github.com/coreos/etcd
-  version: 43b75072bfaca5a7c35c718179defbcabd9a0886
+  version: d267ca9c184e953554257d0acdd1dc9c47d38229
   subpackages:
   - client
   - pkg/pathutil
@@ -143,7 +143,7 @@ imports:
 - name: github.com/fsouza/go-dockerclient
   version: 94aeff6a741494ddf6e5e66b2b0881828468b1df
 - name: github.com/ghodss/yaml
-  version: 73d445a93680fa1a78ae23a5839bad48f32ba1ee
+  version: 0ca9ea5df5451ffdf184b4428c902747c2c11cd7
 - name: github.com/go-openapi/jsonpointer
   version: 46af16f9f7b149af66e5d1bd010e3574dc06de98
 - name: github.com/go-openapi/jsonreference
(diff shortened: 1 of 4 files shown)


jmrodri:
What was the reason behind this PR? IIRC the decoded string is in YAML. We are able to decode yaml into structures quite easily. We can even add ``yaml:"fieldname"`` similar to how we do the json stuff.

rthallisey:
I was passing around a large set of yaml as a config file and found this issue.  I think the YAML to JSON has trouble with complex YAML structures so this was to guarantee we never have that issue.

rthallisey:
@jmrodri does that address your question?

jmrodri:
@rthallisey what large yaml were you passing around? How large of a yaml file was it to cause the problem? Right now things are working as is, unless you can show me what problem this fixes.

jmrodri:
NACK

eriknelson:
Maybe we should log an issue with the problematic yaml and investigate that directly?

rthallisey:
Let me dig up what I was using.  It wasn't bad yaml, but the information was nested 3 or 4 layers down and was never being gathered.   All I'm doing is converting all incoming data, yaml and json, and decoding with the json decoder. Kubernetes does the same thing.  That's where I got the idea from.

https://github.com/kubernetes/apimachinery/blob/4500b7d1a0fc94ffb6d03f4019b2968b47e165bf/pkg/runtime/serializer/json/json.go#L87

shawn-hurley:
Have we seen an issue during the last iteration with this? 
[This](http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/) makes it seem like the real problem is attempting to serialize json and yaml into the same struct. Specs are taken in as YAML and output as JSON. Because of this, I think it might make sense to convert it to all one format to interact with the struct.

jmrodri:
I see the comments in the blog, but I struggle because we've been running for months without a problem in any of the areas where we do yaml conversion. If there was a problem we saw consistently I would be much less of a pain with this simple PR.

Now if we really want to do this then, there are quite a few uses of `yaml.Unmarshal` that need to be replaced then. 

```
$ git grep "yaml.Unmarshal"         
cmd/mock-registry/mock_registry.go:     yaml.Unmarshal(dat, &parsedDat)            
pkg/apb/mock_registry.go:       err = yaml.Unmarshal(specYaml, &parsedData)        
pkg/apb/types.go:       if err = yaml.Unmarshal([]byte(payload), obj); err != nil {
pkg/app/config.go:      if err = yaml.Unmarshal(dat, &config); err != nil {        
```

jmrodri:
@eriknelson thoughts?

jmrodri:
this also needs rebasing since there are conflicts.

jmrodri:
@jmontleon this PR would add a new dependency.

eriknelson:
Sorry guys, needed to catch up on this one.

My current feelings are, I'm still averse to moving from yaml -> json -> struct. golang (or whatever lang) *should* be entirely agnostic after that text runs through a processor. Maybe that's not fair, because [yaml is a technical superset of json](https://stackoverflow.com/a/1729545), but we're not doing anything in our yaml specs that deserve loss in a conversion to json.

Fundamentally, I don't think the data rep should matter from a broker perspective, and it shouldn't dictate broker code.

I'm okay with adopting it if it necessary, but I want to capture limitations of the yaml parser and log issues where they need to be logged, because yaml alone should be sufficient if it's a true superset of json.

eriknelson:
@rthallisey Do we want to close this and reopen if we come across the problem again, or were you working on that so we should leave this open?

rthallisey:
I wanted to see if I could generate the yaml that gave me the issue. Since what we have works, let's stick with it and we can always come back to this if there's a problem.

PR #70, opened 2017-04-11
Work engine

jmrodri:


jmrodri:
A lot of these changes came from the force-async-prov branch. The bulk of the work I did was around provision_job, provision_subscriber, and work-engine (which I brought in from our work on cap-server). 

The gist behind async is that a provision request comes in with a `accepts_incomplete=true` flag. We create a provision job that runs in the background writing the output to a channel. While the job is running, the provision call returns 202 ACCEPTED to the catalog. When the provision job finishes, it will write its output to the channel which the subscriber pulls the message from it and updates etcd with either failed or succeeded and extracted_credentials.

rthallisey:
@jmrodri how many of those patches are only relevant to the async work? There's also a few conflicts.

pkg/broker/provision_subscriber.go
```diff
@@ -0,0 +1,41 @@
...
+	return &ProvisionWorkSubscriber{dao: dao}
+}
+
+func (p *ProvisionWorkSubscriber) Subscribe(msgBuffer <-chan WorkMsg) {
+	p.msgBuffer = msgBuffer
+
+	var pmsg *ProvisionMsg
+	var extCreds *apb.ExtractedCredentials
+	go func() {
+		for {
```
rthallisey: How does this for loop end and go routine exit?

pkg/broker/broker.go
```diff
@@ -156,30 +160,47 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest)
...
-	}
+	var token string
+
+	if async {
+		a.log.Info("ASYNC provisioning in progress")
+		// asyncronously provision and return the token for the lastoperation
+		pjob := NewProvisionJob(spec, parameters, a.clusterConfig, a.log)
+
+		// HACK: wow this feels dirty
+		a.engine.AttachSubscriber(NewProvisionWorkSubscriber(a.dao))
```
rthallisey: This seems like a lot of effort to pass around a token, spec id, and message.

pkg/broker/provision_subscriber.go
rthallisey:
```
defer close(msgBuffer)
for n := range msgBuffer {
...
}
```

pkg/broker/util.go
```diff
@@ -53,3 +53,16 @@ func SpecToService(spec *apb.Spec) Service {
 	copy(retSvc.Tags, spec.Tags)
 	return retSvc
 }
+
+func StateToLastOperation(state apb.State) LastOperationState {
+	switch state {
+	case apb.StateInProgress:
```
rthallisey: Since the strings for apb.StateInProgress and LastOperationsStateInProgress are the same you could just return apb.StateInProgress.

pkg/handler/handler_test.go
```diff
@@ -52,103 +52,94 @@ func (m MockBroker) Unbind(uuid.UUID, uuid.UUID) error {
...
+	state := broker.LastOperationStateInProgress
+	return &broker.LastOperationResponse{State: state, Description: ""}, nil
+}
+
+var b MockBroker
+var dahandler handler
+
+func init() {
+	b = MockBroker{Name: "testbroker"}
+	dahandler = handler{*mux.NewRouter(), b}
```
rthallisey: would ```dahandler := handler{*mux.NewRouter(), b}``` work in the init? If so you could drop the variable creation above.

nit: could make the variable name ```testhandler``` to be more descriptive.

pkg/broker/work_subscriber.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (s *StdoutWorkSubscriber) Subscribe(msgBuffer <-chan WorkMsg) {
+	// Always drain the buffer if there's a message waiting.
+	// Here we're just forwarding to stdout, but of course, the message
+	// destination could be anything (ultimate websockets!)
+	// NOTE: DON'T FORGET TO GOROUTINE THIS, OR WILL YOU CHOKE THE MAIN PROCESSOR
+	s.msgBuffer = msgBuffer
+	go func() {
+		for {
```
rthallisey: Same as above. How does the goroute exit?

rthallisey: Looks like there are two go routines reading from the channel?  One prints to stdout the other saves state.

pkg/apb/types.go
```diff
@@ -50,6 +50,16 @@ type ExtractedCredentials struct {
 	// might be more one day
 }
 
+// HACK: created these State types because I couldn't import broker into the
```
jmrodri: I'm going to remove the HACK keyword. After using these states, I'm ok with them.

pkg/broker/broker.go
jmrodri: It was something we used in our cap-server prototype, so we brought it over from there since it worked ok. Our initial use was to get current updates to a websocket on the UI. I think ultimately that would be something we'd desire is to get more fine grained updates that we can write output to.

pkg/broker/provision_subscriber.go
jmrodri: Yeah, I think I got bit by the cargo cult. This was a long lived subscriber that wrote to websocket. I wrote a quick prototype to use this and sure enough it'll hang forever. :( Working on a fix. Good catch.

pkg/broker/work_subscriber.go
jmrodri: I've got a PR to remove this, it's unused.

pkg/handler/handler_test.go
jmrodri: `dahandler := handler...` won't work because then it isn't visible throughout the test. I get an error. But I was able to remove the `var b MockBroker` by doding the `b := MockBroker`. 

And `dahandler` was my gangsta variable name :) like da popo. I'll make it testhandler though hehe

rthallisey: :)

PR #71, opened 2017-04-12
Add Kubernetes and OpenShift compatible templates

rthallisey:
Create some yaml templates that will create the broker resources with 'oc create -f' or 'kubectl create -f'.

I'm going to also integrate this into catasb.

### DIFF
--- a/templates/broker-deployment.yaml
+++ b/templates/broker-deployment.yaml
@@ -1,0 +1,37 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: asb
+  namespace: ansible-service-broker
+  labels:
+    app: ansible-service-broker
+    service: asb
+spec:
+  strategy:
+    type: Recreate
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: ansible-service-broker
+        service: asb
+    spec:
+      restartPolicy: Always
+      containers:
+        - image: "ansibleapp/ansible-service-broker-asb:latest"
+          name: main
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: 1338
+              protocol: TCP
+          env:
+            - name: DOCKERHUB_PASS
+              value: "{{dockerhub_pass}}"
+            - name: DOCKERHUB_USER
+              value: "{{dockerhub_user}"
+            - name: OPENSHIFT_PASS
+              value: "{{openshift_pass}}"
+            - name: OPENSHIFT_TARGET
+              value: "{{openshift_target}}"
+            - name: OPENSHIFT_USER
+              value: "{{openshift_user}}"
--- a/templates/etcd-deployment.yaml
+++ b/templates/etcd-deployment.yaml
@@ -1,0 +1,35 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: etcd
+  namespace: ansible-service-broker
+  labels:
+    app: ansible-service-broker
+    service: etcd
+spec:
+  strategy:
+    type: Recreate
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: ansible-service-broker
+        service: etcd
+    spec:
+      restartPolicy: Always
+      containers:
+        - image: ansibleapp/ansible-service-broker-etcd:latest
+          name: main
+          imagePullPolicy: IfNotPresent
+          workingDir: /etcd
+          args:
+            - ./etcd
+            - --data-dir=/data
+            - --listen-client-urls=http://0.0.0.0:2379
+            - --advertise-client-urls=http://0.0.0.0:2379
+          ports:
+          - containerPort: 2379
+            protocol: TCP
+          env:
+          - name: ETCDCTL_API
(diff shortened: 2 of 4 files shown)


jmrodri:
@rthallisey what's the best way to test this? Pretend I'm a n00b, like how do I create a kubernetes cluster so I can test this.

rthallisey:
Since Kubernetes is a subset of Openshift, as long as templates don't have openshift specific fields in them they are compatible.  That's a good smoke test.  Setting up Kubernetes I use: https://kubernetes.io/docs/getting-started-guides/kubeadm.  Spawn a plain centos7 vm in vagrant and run those commands on it.

jmrodri:
TEST 
-------
Tested with OpenShift. 
* Bring up an OC cluster: `oc cluster up`
* Create a user using [catasb stage2 script](https://github.com/fusor/catasb/blob/master/archived/stage2/create_cluster_user.sh): 
* Create the services first `oc create -f services.yaml`
* Deploy etcd `oc create -f etcd-deployment.yaml`
* Deploy broker: `OPENSHIFT_TARGET=127.0.0.1:8443 OPENSHIFT_PASS=zeus OPENSHIFT_USER=zeus oc create -f broker-deployment.yaml`
* Create routes for broker `oc create -f route.yaml`

broker-deployment.yaml is failing. Trying to figure out how to pass in the parameters to run it.

templates/broker-deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+              protocol: TCP
+          env:
+            - name: DOCKERHUB_PASS
+              value: "{{dockerhub_pass}}"
+            - name: DOCKERHUB_USER
+              value: "{{dockerhub_user}"
+            - name: OPENSHIFT_PASS
+              value: "{{openshift_pass}}"
+            - name: OPENSHIFT_TARGET
+              value: "{{openshift_target}}"
```
jmrodri: How do I supply these? I tried a few `oc create -f` variations and none seemed to work. I also saw reference to these in your other branch: https://github.com/fusor/ansible-service-broker/tree/broker-build-process

Are these related?

jmrodri: A README.md in this directory would be useful for n00bs to knwo what order to run these in. A little trial and error I seemed to figure out something that worked.

rthallisey: Agreed, this change needs a doc.  You populate those fields by hand, but I also have a patch that will render those files based on the info in etc/dev.config.yaml. The patches you need for that is in this branch broker-build-process. I'll turn those commits in that branch into PRs: https://github.com/fusor/ansible-service-broker/tree/broker-build-process

jmrodri:
Tried to run it through oc process but that failed too. I think the syntax of the variables is weird.
```
oc process -f templates/broker-deployment.yaml | oc create -f -
error: unable to parse "templates/broker-deployment.yaml", not a valid Template but *extensions.Deployment
error: no objects passed to create
```

jmrodri:
NACK can't figure out how these work :( I may just be doing it wrong.

rthallisey:
```oc create -f templates/broker-deployment.yaml``` with the template values filled in.

**Comment by rthallisey** [comment]
Sorry there are 4 templates. You need to do this in order since services.yaml creates the project.
```oc create -f services.yaml```
```oc create -f route.yaml```
```oc create -f etcd-deployment.yaml```
```oc create -f broker-deployment.yaml```

**Comment by jmrodri** [comment]
@rthallisey thanks for the order, you confirmed what I ended up trying. I was not aware I had to edit the files. That's the part I was missing.

**Comment by jmrodri** [comment]
@rthallisey if I fill in those values I'm able to launch the broker using these templates. I don't like having to update the files manually.

**Comment by jmrodri** [comment]
I spoke with Todd to figure out what the goal for the broker templates are. The idea is that we create an OpenShift template that a user can use to launch a broker from a released image.

```
oc process /path/to/template | oc create -f -
```
Similar to the [service catalog template](https://github.com/fusor/catasb/blob/master/archived/stage2/service-catalog.templ.yaml) we created in catasb.

I spoke with Ben Parees about how template broker is deployed. It is a special case, it is built into OpenShift and you just enable it in a config. So there is no template involved. He suggested we look at the OpenShift docs: https://docs.openshift.org/latest/dev_guide/templates.html#writing-templates

For this PR, we should look to creating a single OpenShift template to launch a released broker image.

rthallisey:
@jmrodri I don't like having to manually do it either. In the broker-build-process branch there are three more patches on top of this one that complete the build and deploy pipeline by integrating it with the Makefile.  I'll turn those into PRs.

Ok I'll check out that doc. My target for this patch was to re work the build and deploy process to streamline it for a developer trying to work on it.  Perhaps that template is an another patch after this one.  Either way, this path gets us part of the way and I can follow up with a patch that adds that functionality.

jmrodri:
Can we close this PR then and just get the build branch into shape and make a single PR with those changes? I'm still hesitant to merge this as is.

rthallisey:
Sure. The branch is more developed. Closing this since in favor of reviewing the broker-build-process branch.

PR #83, opened 2017-04-27
add ability to build images

jmrodri:
Introductory make target to build asb images. It's not perfect, but it seems to work and I don't have to worry about missing a step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+REGISTRY         = docker.io/ansibleplaybookbundle
+TAG              = latest
+BROKER_APB_IMAGE = $(REGISTRY)/ansible-service-broker-apb
+
 build: $(shell find cmd pkg)
 	# HACK: Unless docker's vendor directory is removed, we end up with a
 	# duplicate symbol error from the linker that prevents compilation.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,4 +27,16 @@ vendor:
 test: vendor
 	go test ./pkg/...
 
+images: asb-image
+
+asb-image:
```
rthallisey: Add asb-images to .PHONY at the bottom.

Makefile
```diff
@@ -23,4 +27,16 @@ vendor:
 test: vendor
 	go test ./pkg/...
 
+images: asb-image
```
rthallisey: Since ```make images``` calls the ```asb-images``` recipe, we might as well just remove images and do ```make asb-images```

Makefile
rthallisey: How about changing the name to ```ansible-container```?  This is specific to ansible-container and later we'll merge the PR that adds building without ansible-container.  That way this recipe reflects how the apb is being built.

Makefile
```diff
@@ -1,3 +1,7 @@
+REGISTRY         = docker.io/ansibleplaybookbundle
+TAG              = latest
```
rthallisey: Instead of setting these values, let's set them as defaults and let the user override these vars by doing: ```make apb-images TAG=old```. This would look like: ```TAG ?= latest```.

Makefile
```diff
@@ -1,3 +1,7 @@
+REGISTRY         = docker.io/ansibleplaybookbundle
```
rthallisey: Can you break up these two into ```$(Registry)/$(Project)```.  That will allow anyone to point this at any registry and change the name if they want.

jmrodri: Done. I also added `?=` to allow it to default the values if not overridden.

Makefile
jmrodri: Done, made TAG ?= latest

Makefile
jmrodri: I can do that, I was going to just use `make images` and for a second thought we might create more than one. 

What about 

```
images: asb-images apb-image

asb-images:
    ansible-container stuff

apb-image:
    docker build stuff

.PHONY asb-images apb-image
```

jmrodri: I left it as is since your branch will ultimately be where we go.

Makefile
jmrodri: Since you already have a branch for a more long term build process, I decided not to make this perfect. It builds what we have today, and I'm all for moving this to a better process.

rthallisey: Wfm. Let's get this merged since it's a nice improvement.

rthallisey:
+1

PR #95, opened 2017-05-02
Make bind spec compliant

jmrodri:


jmrodri:
Need to run this live a few times.

pkg/broker/broker.go
```diff
@@ -339,12 +339,38 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
...
+		if bi.Id.String() == bindingInstance.Id.String() &&
+			reflect.DeepEqual(bi.Parameters, bindingInstance.Parameters) {
+
+			a.log.Debug("already have this binding instance returning 200")
+			return &BindResponse{}, ErrorAlreadyProvisioned
+		} else if bi.Id.String() == bindingInstance.Id.String() &&
+			!reflect.DeepEqual(bi.Parameters, bindingInstance.Parameters) {
+
+			// TODO: remove these debug statements at some point
+			a.log.Debug("Existing binding parameters")
```
rthallisey: If you aren't using these  for something else, maybe we should drop them so we don't log any bind params.  We should keep that info secret between the broker and apb.

pkg/broker/broker.go
```diff
@@ -339,12 +339,38 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
...
+	// if err is not nil, we will just bubble that up
+	//
 	// if binding instance exists, and the parameters are the same return: 200.
 	// if binding instance exists, and the parameters are different return: 409.
 	//
 	// return 201 when we're done.
-	//
-	// once we create the binding instance, we call apb.Bind
+	if bi, err := a.dao.GetBindInstance(bindingUUID.String()); err == nil {
+		if bi.Id.String() == bindingInstance.Id.String() &&
```
rthallisey: I think there's a slight improvement you can make here. What if you check if the uuid matches first, then inside that if block do a single if block ```if reflect.DeepEqual(bi.Parameters, bindingInstance.Parameters) { ... } else { ... }```.  If you do it that way you will only have to run the DeepEqual function once instead of twice. That should make this slightly more efficient since DeepEqual does a bunch of loops to compare values on the provided data structure.

pkg/handler/handler.go
```diff
@@ -183,9 +184,23 @@ func (h handler) bind(w http.ResponseWriter, r *http.Request, params map[string]
...
 
+	// process binding request
 	resp, err := h.broker.Bind(instanceUUID, bindingUUID, req)
 
-	writeDefaultResponse(w, http.StatusCreated, resp, err)
+	if err != nil {
+		switch err {
+		case broker.ErrorDuplicate:
+			writeResponse(w, http.StatusConflict, broker.BindResponse{})
+		case broker.ErrorAlreadyProvisioned:
```
rthallisey: Nit pick: Is this really an error if we're returning ```http.StatusOK```? You don't have to change this now, but maybe the variable should be renamed.

pkg/broker/broker.go
rthallisey: Missed the title as WIP. Ignore my comment.

pkg/handler/handler.go
jmrodri: That's a hard one because there's no good way to returned what seems like an ok state since there are 3 good states, OK, ACCEPTED, & CREATED. So you *could* look at it as a good case is we go to create one, and trying to provision the SAME instance is an error case since we already did that. The 200 OK is per the spec. That's the thought process I went with since it is an exceptional case of provision.

pkg/broker/broker.go
jmrodri: Yeah I can remove those. I would've forgotten if you didn't mention it :)

pkg/broker/broker.go
jmrodri: DUH! not sure what I was thinking there. I'll fix that.

jmrodri: I need to fix that on provision as a separate Issue #99

PR #97, opened 2017-05-02
make sure jq and etcd are called out

jmrodri:


dymurray:
ACK, looks good to me.

Issue #99, opened 2017-05-02
optimization: simplify the provision duplication detection code


jmrodri:
Provision could call DeepEqual twice which can be expensive. 

https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L197-L217

Also remove the logging of parameters which could contain sensitive information.

jmrodri:
Basically make it look lie this.
https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L351-L362

PR #101, opened 2017-05-04
Removing duplicate calls to reflect.DeepEquals in broker

shawn-hurley:
fixing issue #99 

* Removing debug of Parameters
* Reducing the deep equals calls
* Adding helper method to check if service instance's are equal

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -2,6 +2,7 @@ package apb
 
 import (
 	"encoding/json"
+	"reflect"
 
 	logging "github.com/op/go-logging"
 	"github.com/pborman/uuid"
@@ -101,6 +102,14 @@ type ServiceInstance struct {
 	Id         uuid.UUID   `json:"id"`
 	Spec       *Spec       `json:"spec"`
 	Parameters *Parameters `json:"parameters"`
+}
+
+//IsEqual - Compares the Id and Parameters of the instance to determine if SericeInstance is equal to the instance passed in.
+func (s ServiceInstance) IsEqual(si *ServiceInstance) bool {
+	if s.Id.String() != si.Id.String() {
+		return false
+	}
+	return reflect.DeepEqual(s.Parameters, si.Parameters)
 }
 
 type BindInstance struct {
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -2,7 +2,6 @@ package apb
 
 import (
 	"encoding/json"
-	"reflect"
 
 	logging "github.com/op/go-logging"
 	"github.com/pborman/uuid"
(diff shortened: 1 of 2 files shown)


pkg/apb/types.go
```diff
@@ -103,6 +104,14 @@ type ServiceInstance struct {
...
 }
 
+//IsEqual - Compares the Id and Parameters of the instance to determine if SericeInstance is equal to the instance passed in.
+func (s ServiceInstance) IsEqual(si *ServiceInstance) bool {
+	if s.Id.String() != si.Id.String() {
+		return false
+	}
+	return reflect.DeepEqual(s.Parameters, si.Parameters)
+}
+
```
jmrodri: While this is technically correct, it starts to set a precedence for having IsEqual methods on our structs. I'd prefer not to add it to the ServiceInstance.

pkg/broker/broker.go
```diff
@@ -193,26 +193,13 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
...
-				a.log.Debug("%s = %s", k, v)
-			}
-			a.log.Debug("Incoming parameters")
-			for k, v := range map[string]interface{}(*serviceInstance.Parameters) {
-				a.log.Debug(fmt.Sprintf("%s = %s", k, v))
-			}
-
-			a.log.Info("we have a duplicate instance with identical parameters, returning 409 conflict")
+		} else if si.Id.String() == serviceInstance.Id.String() {
+			a.log.Info("we have a duplicate instance with parameters that differ, returning 409 conflict")
```
jmrodri: I was leaning towards just copying what we did on bind.

https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L351-L362

```
	if bi, err := a.dao.GetBindInstance(bindingUUID.String()); err == nil {
		if bi.Id.String() == bindingInstance.Id.String() {
			if reflect.DeepEqual(bi.Parameters, bindingInstance.Parameters) {
				a.log.Debug("already have this binding instance, returning 200")
				return &BindResponse{}, ErrorAlreadyProvisioned
			}

			// parameters are different
			a.log.Info("duplicate binding instance diff params, returning 409 conflict")
			return nil, ErrorDuplicate
		}
	}
```

This also IMO makes it easier to compare to the spec. Which lists out the conditions for returning. I.e. if the ids are the same and the parameters are the same return 200, else if the ids are the same but the parameters are different, 409.

jmrodri:
For posterity, this PR fixes #99 .

pkg/broker/broker.go
shawn-hurley: That sounds good, makes sense about the reasoning, What is your feeling about changing the comparison of Id to the [uuid.Equals]( https://godoc.org/github.com/pborman/uuid#Equal). 

I think this would prevent issues in the future if the library ever went to a something other than what the string function could compare? 
ex.
``` if uuid.Equals(si.Id, serviceInstance.Id) {```

jmrodri: :+1: to  [uuid.Equals]( https://godoc.org/github.com/pborman/uuid#Equal).  Good find, I didn't even think to look at that.

shawn-hurley: should I also change other places in the broker.go? Other places could be changed as they are run into?

jmrodri: Change it in this particular case AND in the bind case. Then create a *new* issue for converting the other places to `uuid.Equals`.

Issue #63, opened 2017-04-06
Clean up dao redundancy


eriknelson:
Most of the getters/setters in the dao are doing the same thing varying by type. Would be really nice to introduce some kind of "serializable" interface and do the work in one place. The gopher stole my generics.

jmrodri:
Fixed by PR #105

PR #105, opened 2017-05-11
Removing redundancy in dao for issue #63

shawn-hurley:
Fix for issue #63 

* Also fixing golint things like comments on export methods and structs

### DIFF
--- a/pkg/dao/dao.go
+++ b/pkg/dao/dao.go
@@ -10,11 +10,13 @@ import (
 	logging "github.com/op/go-logging"
 )
 
+//Config - confg holds etcd host and port.
 type Config struct {
 	EtcdHost string `yaml:"etcd_host"`
 	EtcdPort string `yaml:"etcd_port"`
 }
 
+//Dao - retreive, create, and update  objects from etcd
 type Dao struct {
 	config    Config
 	log       *logging.Logger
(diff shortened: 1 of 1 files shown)


pkg/dao/dao.go
```diff
@@ -10,11 +10,13 @@ import (
 	logging "github.com/op/go-logging"
 )
 
+//Config - confg holds etcd host and port.
```
jmrodri: Add a space after the `//` not a golint check, but a zeus check. I prefer the space :smile:

pkg/dao/dao.go
```diff
@@ -105,26 +111,22 @@ func (d *Dao) BatchGetRaw(dir string) (*[]string, error) {
 	return &payloads, nil
 }
 
+//GetSpec - Retrieve the spec for the kvp API.
 func (d *Dao) GetSpec(id string) (*apb.Spec, error) {
-	raw, err := d.GetRaw(specKey(id))
+	spec := &apb.Spec{}
+	err := d.getObject(specKey(id), spec)
```
jmrodri: Let's also change it to include the error check in the same line.
```
if err := d.getObject(specKey(id), spec); err != nil {
   return nil, err
}
```
I know it will result in more changes, but I think it looks better than what we had.

pkg/dao/dao.go
```diff
@@ -155,114 +157,97 @@ func (d *Dao) BatchGetSpecs(dir string) ([]*apb.Spec, error) {
...
+}
+
+func (d *Dao) setObject(key string, data interface{}) error {
+	payload, err := apb.DumpJSON(data)
+	if err != nil {
+		return err
+	}
+	return d.SetRaw(key, payload)
+}
+
```
jmrodri: +1

PR #111, opened 2017-05-15
adding ability to continue bootstraping when a spec fails

shawn-hurley:
* When specs fail, print out the error to the log.
* Only include correctly created specs in the completed list.

### DIFF
--- a/pkg/apb/dev_registry.go
+++ b/pkg/apb/dev_registry.go
@@ -10,13 +10,16 @@ import (
 	logging "github.com/op/go-logging"
 )
 
+// AppsPath - path to bundles for dev registry.
 const AppsPath = "/bundles"
 
+// DevRegistry - dev registry
 type DevRegistry struct {
 	config RegistryConfig
 	log    *logging.Logger
 }
 
+// Init - initialize the DevRegistry
 func (r *DevRegistry) Init(config RegistryConfig, log *logging.Logger) error {
 	log.Debug("DevRegistry::Init")
 	r.config = config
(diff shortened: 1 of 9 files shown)


shawn-hurley:
I am wondering if the BootstrapResponse should have ImageCount and SpecCount. That way we can say created X specs of Y images to the clients? We could also name -> bool object for each image if it was created?

pkg/apb/dockerhub_registry.go
```diff
@@ -74,7 +72,7 @@ func (r *DockerHubRegistry) createSpecs(
 		}
 
 		if _err = LoadYAML(string(decodedSpecYaml), _spec); _err != nil {
-			r.log.Error("Something went wrong loading decoded spec yaml")
+			r.log.Error("Something went wrong loading decoded spec yaml - %v - %v", string(decodedSpecYaml), _err)
```
jmrodri: :+1: I like this makes it much easier to debug.

jmrodri:
@shawn-hurley I would be ok with the `BootstrapResponse` changes you propose. Having imported count would be useful.

Go from this:

```
type BootstrapResponse struct {
    SpecCount int
}
```
to:
```
type BootstrapResponse struct {
    ImportedCount int
    SpecCount int
}
```

Something like that, might need a better name.

pkg/apb/dev_registry.go
```diff
@@ -10,30 +10,34 @@ import (
...
+// AppsPath - path to bundles for dev registry.
 const AppsPath = "/bundles"
 
+// DevRegistry - dev registry
 type DevRegistry struct {
 	config RegistryConfig
 	log    *logging.Logger
 }
 
+// Init - initialize the DevRegistry
```
jmrodri: :heart: that you add comments :)

pkg/apb/dev_registry.go
```diff
@@ -10,30 +10,34 @@ import (
...
 	return nil
 }
 
-func (r *DevRegistry) LoadSpecs() ([]*Spec, error) {
+// LoadSpecs - Load Specs from the DevRegistry
+func (r *DevRegistry) LoadSpecs() ([]*Spec, int, error) {
 	r.log.Debug("DevRegistry::LoadSpecs")
 
-	appsUrl := r.fullAppsPath()
+	appsURL := r.fullAppsPath()
```
jmrodri: Heh following good guidelines :) using `URL` instead of `Url`.

jmrodri:
@shawn-hurley the tests failed. `make test` will run all the tests. To debug and work on a single test, in this case I believe it is `handler_test.go`. Do the following workflow, it makes things easier IMO:

```
cd pkg/handler
go test -v
```
Runs just the tests in `handler_test.go` making it easier to figure out what went wrong. I suspect the `BootstrapResponse` we create in the test no longer matches the change you made in this PR.

jmrodri:
TEST
------
Bootstrap returns the image count and the spec count now. :+1: 
```
$ curl -X POST http://localhost:1338/v2/bootstrap
{
  "spec_count": 7,
  "image_count": 9
}
```

And I see in the log the bad spec.

jmrodri:
ACK, fix the tests, and I'll merge.

jmrodri:
ACK

PR #112, opened 2017-05-15
Flatten glide dependencies

### LINKED ISSUE #100: Fix up glide.yaml to flatten dependecies
As part of make build we remove docker vendor dependency instead of having glide flatten it ```rm -rf ${GOPATH}/src/github.com/fusor/ansible-service-broker/vendor/github.com/docker/docker/vendor```.  We should get glide.yaml & glide.lock into good shape so we don't have to do this anymore and it avoid this coming up in the future.

rthallisey:
Add the appropriate pks to glide.yaml so that the vendors
directory can be properly flattened when running `glide install -v`.

Flattening the vendor directory removes duplicate pkgs that will
conflict with pkgs used by the broker.

Partially fixes #100

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -46,3 +46,18 @@ import:
   version: 7da10c8c50cad14494ec818dcdfb6506265c0086
 - package: github.com/gorilla/handlers
   version: ~1.2.0
+- package: github.com/coreos/go-oidc
+- package: github.com/opencontainers/image-spec
+  version: v1.0.0-rc4
+- package: github.com/Sirupsen/logrus
+  version: v0.11.4
+- package: github.com/howeyc/gopass
+- package: github.com/opencontainers/go-digest
+  version: ^1.0.0-rc0
+- package: github.com/pkg/errors
+  version: ^0.8.0
+- package: github.com/blang/semver
+  version: ^3.5.0
+- package: k8s.io/client-go
+  version: ^3.0.0-beta.0
+- package: k8s.io/apimachinery


jmrodri:
@rthallisey I'm not quite sure what this accomplishes. We can talk tomorrow. I'm missing something I think :)

rthallisey:
@jmrodri one of the issues I have right now when adding the client is the duplicate pkgs that are in the vendor directory and that are required by the broker.  What this patch does is add the necessary pks to glide so you can run ```glide -v``` removing the nested vendor directories like from the Makefile: ```rm -rf ${GOPATH}/src/github.com/fusor/ansible-service-broker/vendor/github.com/docker/docker/vendor``` so the broker will compile.

Since I'm including ```k8s.io/apimachinery```, I think we're going to need to do ```glide install -v``` everywhere.   We also can remove ```rm -rf ${GOPATH}/src/github.com/fusor/ansible-service-broker/vendor/github.com/docker/docker/vendor``` from the Makefile. Those will be follow up patches.

jmrodri:
I **think** I understand. So there are packages that we use that are not in glide.yaml but are getting pulled in through one of the vendored packages. So by adding them in we can have glide strip off all vendored from the deps

rthallisey:
You're super close.

All the packages the broker uses are in glide.yaml.  So adding packages into glide.yaml adds that package to the vendor directory. But, some of those packages, the larger ones like docker/docker, have dependencies.  Those dependencies get added to a vendor directory that's nested in the broker's vendor directory -> ```/src/github.com/fusor/ansible-service-broker/```**vendor**```/github.com/docker/docker/```**vendor**```/ ...```.  Then, if we have some code in the broker that includes a package, but that same package gets included also as a dependency of another package, we will get a conflict because there will be two sources the compiler can use. For example, the source of client-go could be ```/src/github.com/fusor/ansible-service-broker/```**vendor**```/client-go``` and ```/src/github.com/fusor/ansible-service-broker/```**vendor**```/github.com/docker/docker/```**vendor**```/client-go```. And glide.yaml would include both client-go and docker as packages used by the broker.

PR #114, opened 2017-05-16
Update glide.lock file

### LINKED ISSUE #100: Fix up glide.yaml to flatten dependecies

rthallisey:
We need to update the glide.lock file to reflect the recent changes to glide.yaml.

Partially fixes #100

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,7 @@ BROKER_APB_IMAGE = $(REGISTRY)/$(PROJECT)/ansible-service-broker-apb
 BUILD_DIR        = "${GOPATH}/src/github.com/fusor/ansible-service-broker/build"
 
 install: $(shell find cmd pkg)
-	# HACK: Unless docker's vendor directory is removed, we end up with a
-	# duplicate symbol error from the linker that prevents compilation.
-	rm -rf ${GOPATH}/src/github.com/fusor/ansible-service-broker/vendor/github.com/docker/docker/vendor && \
-		go install -ldflags="-s -w" ./cmd/broker
+	go install -ldflags="-s -w" ./cmd/broker
 
 ${GOPATH}/bin/mock-registry: $(shell find cmd/mock-registry)
 	go install ./cmd/mock-registry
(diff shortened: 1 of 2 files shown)


jmrodri:
TEST INSTRUCTIONS
-------------------------
* add `-v` option to end of `@glide install` in the `vendor` target of the `Makefile`
* comment out the `rm -rf` of docker vendor directory from the `install` target

Then do the following:
1. remove the vendor directory: `rm -rf vendor`
2. clear out your `glide` cache: `glide cache-clear`
3. recreate the vendor directory: `make vendor`
4. rebuild broker: `make install`
5. PROFIT: :heavy_dollar_sign: :heavy_dollar_sign: :heavy_dollar_sign:

rthallisey:
@jmrodri Did this work for you? I added in the stuff you mentioned in your test comment.

TBH I really think we should just remove glide.lock all together... Maybe in a follow up patch.

jmrodri:
The glide.lock is there for a reason. The purpose is to lock down the actual deps. On rails and javascript files, we would check in the lock files. So I would vote no to *NOT* checking in the glide.lock file.

rthallisey:
Ya but glide.yaml creates glide.lock based on the version that set in there for each pkg. I remember reading somewhere that we're not suppose to touch it. I'll check into it.

jmrodri:
PROFIT: :heavy_dollar_sign: :heavy_dollar_sign: :heavy_dollar_sign: 

The test past. I'll see why travis is having issues.

jmrodri:
Fixed travis. Our travis script was doing a normal `glide install` outside the `Makefile`. So that was creating the vendor directory WITH the vendor subdirs. So when our new Makefile never ran the `glide install -v`. Now that this is fixed. I'll verify this PR tomorrow morning, don't want to commit it tonight.

Issue #73, opened 2017-04-18
Deprovision should return 410 when instanceID is not found


jmrodri:
Change Deprovision in broker.go to return a NotFound error instead of generic etcd error. The `handler.go` file seems to look at the error to see if it `IsNotFound` but that doesn't seem to be an `http.StatusNotFound`. We either update `handler.go` to use `http` codes or use the `k8s` codes.

PR #117, opened 2017-05-17
Deprovison spec compliance

shawn-hurley:
Fixes issue #73

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -44,7 +44,7 @@ type AnsibleBroker struct {
 	engine        *WorkEngine
 }
 
-// NewAnsibleBroker - creates a new ansbile broker
+// NewAnsibleBroker - creates a new ansible broker
 func NewAnsibleBroker(
 	dao *dao.Dao,
 	log *logging.Logger,
@@ -312,18 +312,17 @@ func (a AnsibleBroker) Deprovision(instanceUUID uuid.UUID) (*DeprovisionResponse
 	}
 	// bubble up  error
 	if err != nil {
-		a.log.Error("unknown etcd error - %#v", err)
+		a.log.Error("Error from etcd - %#v", err)
 		return nil, err
 	}
 	err = apb.Deprovision(instance, a.log)
-	// TODO: Right now the client we are using does not expose bb
 	if err == docker.ErrNoSuchImage {
 		a.log.Debug("unable to find service instance - %#v", err)
 		return nil, ErrorNotFound
 	}
 	// bubble up error.
 	if err != nil {
-		a.log.Error("unknown etcd error - %#v", err)
+		a.log.Error("error from deprovision - %#v", err)
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


jmrodri:
NACK needs work. Adding this comment in case folks don't see the review above.

jmrodri:
These errors 

https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L272-L281

need to be ones we defined in this file:

https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L15-L19

Because as it stands now, line 272 will return an etcd error, line 281 would return something from the apb, and line 273 something different as well. So basically using the errors we defined (feel free to add a new one if necessary) return one of those as appropriate.

Then in handler.go, check the specific errors and return the correct http error. Let me know if you need more clarification.

pkg/broker/broker.go
```diff
@@ -265,23 +276,35 @@ func (a AnsibleBroker) Deprovision(instanceUUID uuid.UUID) (*DeprovisionResponse
...
+	// bubble up  error
+	if err != nil {
+		a.log.Error("unknown etcd error - %#v", err)
 		return nil, err
 	}
-
-	if err = apb.Deprovision(instance, a.log); err != nil {
+	err = apb.Deprovision(instance, a.log)
+	// TODO: Right now the client we are using does not expose bb
+	if err == docker.ErrNoSuchImage {
```
shawn-hurley: If you follow the path of the docker client, it could come back with the image this error.

jmrodri:
@shawn-hurley for this PR I'd like to see 2 things

1) rebase it so we no longer have conflicts
2) go to the Provision and Bind calls and change those to be consistent with your new changes. i.e. in Provision we are looking at the error string for "100".

Alternatively, change deprovision to use the same handling as provision. But I think using the clients error validation is better.

pkg/broker/broker.go
```diff
@@ -265,23 +275,35 @@ func (a AnsibleBroker) Deprovision(instanceUUID uuid.UUID) (*DeprovisionResponse
...
-	if instance, err = a.dao.GetServiceInstance(instanceId); err != nil {
+	instance, err := a.dao.GetServiceInstance(instanceID)
+	/// Handle case where service is not found in etcd
+	if client.IsKeyNotFound(err) {
+		a.log.Debug("unable to find service instance - %#v", err)
+		return nil, ErrorNotFound
+	}
+	// bubble up  error
+	if err != nil {
+		a.log.Error("unknown etcd error - %#v", err)
```
rthallisey: Even though we aren't catching the error explicitly, I think etcd will know the error.  Maybe it should be more like ```Error from etcd - ...```.

pkg/broker/broker.go
```diff
@@ -265,23 +275,35 @@ func (a AnsibleBroker) Deprovision(instanceUUID uuid.UUID) (*DeprovisionResponse
...
 		return nil, err
 	}
-
-	if err = apb.Deprovision(instance, a.log); err != nil {
+	err = apb.Deprovision(instance, a.log)
+	// TODO: Right now the client we are using does not expose bb
```
shawn-hurley: I can't remember, so removing.

pkg/broker/broker.go
```diff
@@ -37,6 +43,7 @@ type AnsibleBroker struct {
 	engine        *WorkEngine
 }
 
+// NewAnsibleBroker - creates a new ansbile broker
```
jmrodri: typo: ansbile -> ansible

rthallisey:
@shawn-hurley LGTM. Just needs a rebase.

PR #121, opened 2017-05-18
Replace get_images_from_org

rthallisey:
Currently we gather the list of images in a bash script. Let's pull that logic into the broker.
- Running the broker locally and in a cluster will use the same function instead
of looking for the script in <path>.
- One less file to track between runtime environments.
- Adding the logic to the broker will give us more flexibility to implement async pulls
in the future
- I finally figured out why my apbs were never loaded into the broker because we limit
the search to 100 just like the dockerhub!

TODO: Remove the param in the broker that accepts the script directory as a path. (The scripts
directory may still be used. In that case leave it in)


jmrodri:
This needs rebasing now since I merged your other PR.

jmrodri:
I was able to run this against our dockerhub registry, it worked.

rthallisey:
@jmrodri rebased

jmrodri:
TEST
-------
* make clean
* make install
* make run
* curl -X POST http://localhost:1338/v2/bootstrap

I was able to load up my docker hub org as well as ansibleplaybookbundle. 

**PASS**

Issue #104, opened 2017-05-09
Bind request seg-faults when no credentials are available to the broker for return


eriknelson:
If a bind request is made against an instance, and no credentials have been returned from the apb during the provision or bind, we end up without a credentials object in etcd (as we should), and the broker actually segfaults somewhere around here:

https://github.com/fusor/ansible-service-broker/blob/master/pkg/broker/broker.go#L386

The segfault is absolutely a bug, because my intent as written was to handle this case, so it's not behaving as expected.

Is it okay for the catalog to request a bind against an instance that hasn't given the broker any credentials, either by choice or because of an error in the apb? How should the broker handle this?

eriknelson:
From a comment on #60, because it's probably more appropriate to discuss here and I don't want it lost:

```
Credentials returned:
prov: T, bind: F -> pass back prov creds from bind, we do this today

prov: F, bind: T -> pass back bind creds, basically the same as the above

prov: T, bind: T -> First, is this okay? If so, how do we handle this? Merge one onto the other? Which 
  takes precendence if there is a key conflict?

prov: F, bind F -> Is this okay? I lean no. If a bind is basically defined by the return of credentials, if 
  there are none available, this is bad. Can we think of a case where a bind makes sense that 
  *doesn't* return anything to the consuming application? NOTE: This currently segfaults. Needs to get 
   fixed and tested against our decision here.
```

@rthallisey @jmrodri @jwmatthews thoughts?

rthallisey:
Since we don't want to block on a bind, we could create a go routine that monitors the apb for the bind creds.  Then, we can return a 201 with the assumption that the bind operation was executed and will be completed some time in the future. 
```
go func() {
   if err != nil {
      monitorOutput(...)
    }()
return 201
```
https://github.com/fusor/ansible-service-broker/pull/60/files#diff-354ba5b1a51aebd68d32f9772b760a2aR57

eriknelson:
[...]

PR #124, opened 2017-05-24
Broker bind output rework

rthallisey:
Change the broker bind output from trailing the logs of
the provision or bind container to exec'ing into the container.

### DIFF
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -22,7 +22,7 @@ func ExtractCredentials(
 	credOut := make(chan []byte)
 
 	log.Debug("Calling monitorOutput on " + podname)
-	go monitorOutput(podname, credOut, log)
+	go monitorOutput(namespace, podname, credOut, log)
 
 	var bindOutput []byte
 	for msg := range credOut {
@@ -32,14 +32,14 @@ func ExtractCredentials(
 	return buildExtractedCredentials(bindOutput)
 }
 
-func monitorOutput(podname string, mon chan []byte, log *logging.Logger) {
+func monitorOutput(namespace string, podname string, mon chan []byte, log *logging.Logger) {
 	// TODO: Error handling here
 	// It would also be nice to gather the script output that exec runs
 	// instead of only getting the credentials
 	retries := 20
 
 	for r := 1; r <= retries; r++ {
-		output, _ := RunCommand("oc", "exec", podname, "broker-bind-creds")
+		output, _ := RunCommand("oc", "exec", podname, "broker-bind-creds", "--namespace="+namespace)
 
 		stillWaiting := strings.Contains(string(output), "ContainerCreating") ||
 			strings.Contains(string(output), "NotFound") ||
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -11,7 +11,6 @@ import (
 	logging "github.com/op/go-logging"
 )
 
-var stillWaitingError = "status: still waiting to start"
 var timeoutFreq = 6    // Seconds
 var totalTimeout = 900 // 15min
 
@@ -44,8 +43,17 @@ func monitorOutput(namespace string, podname string, mon chan []byte, log *loggi
 		stillWaiting := strings.Contains(string(output), "ContainerCreating") ||
 			strings.Contains(string(output), "NotFound") ||
 			strings.Contains(string(output), "container not found")
+		podCompleted := strings.Contains(string(output), "current phase is Succeeded") ||
+			strings.Contains(string(output), "cannot exec into a container in a completed pod")
+
+		// TODO: Replace the string parsing by passing around the pod
+		// object and checking its status
 		if stillWaiting {
 			log.Warning("Retry attempt %d: Waiting for container to start", r)
+		} else if podCompleted {
+			close(mon)
+			log.Notice("APB completed")
+			return
 		} else if strings.Contains(string(output), "BIND_CREDENTIALS") {
 			mon <- output
 			close(mon)
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"strings"
 	"time"
 
@@ -28,6 +27,10 @@ func ExtractCredentials(
 		bindOutput = msg
 	}
 
+	if bindOutput == nil {
+		return nil, nil
+	}
+
 	return buildExtractedCredentials(bindOutput)
 }
 
@@ -35,10 +38,9 @@ func monitorOutput(namespace string, podname string, mon chan []byte, log *loggi
 	// TODO: Error handling here
 	// It would also be nice to gather the script output that exec runs
 	// instead of only getting the credentials
-	retries := 20
 
-	for r := 1; r <= retries; r++ {
-		output, _ := RunCommand("oc", "exec", podname, "broker-bind-creds", "--namespace="+namespace)
+	for r := 1; r <= CredentialRetries; r++ {
+		output, _ := RunCommand("oc", "exec", podname, GatherCredentialsCMD, "--namespace="+namespace)
 
 		stillWaiting := strings.Contains(string(output), "ContainerCreating") ||
 			strings.Contains(string(output), "NotFound") ||
@@ -49,26 +51,22 @@ func monitorOutput(namespace string, podname string, mon chan []byte, log *loggi
 		// TODO: Replace the string parsing by passing around the pod
 		// object and checking its status
 		if stillWaiting {
-			log.Warning("Retry attempt %d: Waiting for container to start", r)
+			log.Warning("[%s] Retry attempt %d: Waiting for container to start", podname, r)
 		} else if podCompleted {
 			close(mon)
-			log.Notice("APB completed")
+			log.Notice("[%s] APB completed", podname)
 			return
 		} else if strings.Contains(string(output), "BIND_CREDENTIALS") {
 			mon <- output
 			close(mon)
-			log.Notice("Bind credentials found")
+			log.Notice("[%s] Bind credentials found", podname)
 			return
-		} else {
-			log.Warning("Retry attempt %d: exec into %s failed", r, podname)
 		}
 
-		log.Debug(string(output))
+		log.Warning("[%s] Retry attempt %d: exec into %s failed", podname, r, podname)
 		time.Sleep(time.Duration(r) * time.Second)
-
 	}
-	t := fmt.Sprintf("ExecTimeout: Failed to gather bind credentials after %d retries", retries)
-	mon <- []byte(t)
+	log.Errorf("[%s] ExecTimeout: Failed to gather bind credentials after %d retries", podname, CredentialRetries)
 	close(mon)
 	return
 }
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


jmrodri:
oh this is going to be trying to merge this with my changes :cry: your PR makes sense. The changes I had to make was to move getPodName out of extractCredentials call. because I need to get the pod name before the job finishes.

jmrodri:
Good work. I'll look at it probably not until later this week.

jmrodri:
actually doesn't look too bad for me.

rthallisey:
Testing this will require: https://github.com/fusor/apb-examples/pull/25

rthallisey:
@jmrodri rebased https://github.com/fusor/apb-examples/pull/25. So this should be ok to test.

jwmatthews:
I'm apprehensive of merging this until we have a working broker with new service catalog.
My preference is to debug issues with broker<->service catalog, after those are resolved, test this PR and determine if we merge.
Main concern is introducing another variable into the mix when things are not working.

@rthallisey would you help me understand the benefits of this approach with oc exec and reading file opposed to parsing oc logs? 

@rthallisey what problems do you see this resolving that we may have faced with the oc logs usage?

rthallisey:
@jwmatthews When we send the credentials to stdout we're making the creds readable by anyone who can view that pod.  If we move to exec, the pod will have to be visible and we can use the container environment to secure who can look at the credentials. That could be things like selinux, users, passwords, or tokens to read the credentials that we can't use to protect stdout.

eriknelson:
I'm a little undecided on this. I'm convinced it's more secure than our current implementation given @rthallisey's argument, so even if we merged as is, there's value to be gained. On the other hand, it's not quite one of the three options I remember discussing when credential exfiltration was originally discussed. IIRC we settled on 3 options:

1) Some kind of pod exit status that was yet to be implemented, but soon to be available?
2) Broker opens a webhook, networking concerns here.
3) Some kind of shared volume that the broker and pod can both read/write from.

This PR isn't really any of those options. I don't think that disqualifies it though; is it a 4th option that accomplishes everything we're looking for?

Regardless of the answer, this is definitely better than what we're doing right now.

rthallisey:
@eriknelson Ya it's not one of the original three, but here's my thinking.

> Some kind of pod exit status that was yet to be implemented, but soon to be available?
>
I interpreted this as reading the pod exit status from something like an ```oc describe```. I didn't understand this one because this seems like it would be the same as what we're doing now.

I'm going to look in the API code and see how a pod can return an error code and maybe that's what was meant here.

> Broker opens a webhook, networking concerns here.
>
Networking concerns.

> Some kind of shared volume that the broker and pod can both read/write from.
>
This is nearly identical to exec. In the container we don't allow anyone but the broker to read from the directory that's attached to the volume. But, now we have to create and destroy resources on the broker side.  Exec doesn't need that.

jwmatthews:
In regard to this PR, I'd like to chat with some of the AOS folks to run this approach by them and get feedback before we merge.   

Plan to address feedback in next sprint, tracking in card: https://trello.com/c/PZJfBeph

jwmatthews:
@rthallisey please rebase this when you can.

I spoke to Clayton and confirmed he's good with this approach.

rthallisey:
Depends on #242

pkg/apb/ext_creds.go
```diff
@@ -14,78 +15,56 @@ var stillWaitingError = "status: still waiting to start"
 var timeoutFreq = 6    // Seconds
 var totalTimeout = 900 // 15min
 
-// ExtractCredentials - Extract credentials from pod in a certain namespace.
```
shawn-hurley: Please keep this line in. I am pretty sure this is why the build is failing.

pkg/apb/ext_creds.go
```diff
@@ -15,77 +16,56 @@ var timeoutFreq = 6    // Seconds
...
-}
+func monitorOutput(podname string, mon chan []byte, log *logging.Logger) {
+	// TODO: Error handling here
+	// It would also be nice to gather the script output that exec runs
+	// instead of only getting the credentials
+	retries := 20
+
+	for r := 1; r <= retries; r++ {
+		output, _ := RunCommand("oc", "exec", podname, "broker-bind-creds")
+
```
shawn-hurley: I don't think that we should throw this error away. I would prefer to check it and output it contents so if oc command fails we have a better chance at debugging imo

pkg/apb/ext_creds.go
```diff
@@ -15,77 +16,56 @@ var timeoutFreq = 6    // Seconds
...
-	log.Debug("Calling monitorOutput on " + podname)
-	credOut, _ := monitorOutput(podname, namespace)
-	log.Debug("oc log output: %s", string(credOut))
+	credOut := make(chan []byte)
 
-	var creds *ExtractedCredentials
-	var err error
-	creds, err = buildExtractedCredentials(credOut)
+	log.Debug("Calling monitorOutput on " + podname)
+	go monitorOutput(podname, credOut, log)
```
eriknelson: I'm okay with the goroutine here if it doesn't introduce complexity for maintenance purposes. Realistically, we don't need this here in the next few months, and in that time frame, we're definitely going to need to introduce async binds with the `WorkEngine`, so we'll have a strategic way to parallelize that work. Just wanted to comment so people are aware, this isn't needed when we spin off bind work.

pkg/apb/ext_creds.go
jmrodri: I've been bit by not knowing what failed during a command run.

eriknelson:
Sent an email to the list about this, plan is the following:

>    Test the PRs for apb-examples/ansible-service-broker prior to merging
>    Build broker locally, build apb-base, mediawiki/postgres and push to
> private dockerhub
>    Test a provision/bind workflow with mediawiki and postgres
>    If all looks good proceed with merge to master and rebuild.

Will do this first thing tomorrow.

eriknelson:
# Test
1) Rebuilt broker from master with this patch applied
2) Applied https://github.com/fusor/apb-examples/pull/25.patch to apb-examples master
3) Rebuilt apb-base as eriknelson/apb-base:bind-output-test.
4) Edited mediawiki, rhscl-postgresql, and hello-world apb.yml files to use eriknelson org instead of ansibleplaybookbundle in the image field. Ran apb prepare on each to update metadata block on dockerfiles. Edited docker files to build from eriknelson/apb-base:bind-output-test. Pushed all images to eriknelson org and ran docker rmi on local apb images.
5) Updated catasb group_vars to deploy docker.io/eriknelson/ansible-service-broker:bind-output-test as the broker image. Edited my_vars.yaml to use eriknelson dockerhub org instead of ansibleplaybook bundle.

Looks like the apb pod never completes, but the broker thinks things completed and was returning "success" from last_operation endpoint. Broker can't seem to find the running apb despite it being available and running in namespace "foo". First thing I would check is to make sure ext_creds is looking in the correct namespace to find the pod.

Logs: https://gist.github.com/eriknelson/1debd19fd27efb11675f1b70655fbc17

pkg/apb/ext_creds.go
rthallisey: output will contain the error. I use it on line 45.  Getting the other param will get ```exit status 1``` from the command. I think it's because we use CombinedOutput ```output, err := exec.Command(cmd, args...).CombinedOutput()```

eriknelson:
# TEST: failure
Trying to deploy apbs that do not output bind credentials causes the apb to never finish:
https://gist.github.com/eriknelson/b8b9f7ae0d05dbb5eb1579f31fca114c

Media wiki actually successfully deployed, but the apb pod is never finishing.

bind-creds is touched, but nothing ever write to the file. The broker looks like it perpetually is trying to get credentials that will never show up there. I think the apb side of things needs to be rethought.

If an apb isn't going to be outputting credentials, it shouldn't enter into the bind-init loop, and when the broker requests extracted credentials, it should be told "this apb doesn't output credentials for provision / bind, stop asking". That way, as an apb author, I can specifically say whether or not the broker should expect credentials out of a provision or a bind op.

rthallisey:
There are a few issues with this patch. First, it was written when provision didn't gather credentials. So we only needed to care about bind. Second, ```oc logs -f``` would trail the logs until the container completes.  This guaranteed that the service account could never expire until ExtractCredentials  returned.

Fixes:
 - Instead of creating bind-creds in an APB, let the playbook do it. If the playbook does it then the expectation is we are able to provide bind creds.
 - Don't return credentials until we either have them or the pod is no longer running.

rthallisey:
Update: I want to make a few improvements to this so hold off until then.

rthallisey:
Ok @eriknelson fixed the issues I noted above

jwmatthews:
@rthallisey have you seen success with:
1) Provision mediawiki
2) Provision rhscl-postgres
3) Bind mediawiki to postgres

rthallisey:
>  Provision mediawiki
    Provision rhscl-postgres
    Bind mediawiki to postgres
>
@jwmatthews Provision mediawiki works fine. It's not bindable and it doesn't generate creds. Provision rhscl-prostgres works. It's bindable and the creds are being added to etcd.
```
$ curl http://172.30.25.0:2379/v2/keys/extracted_credentials/ab360ffb-6f0a-4424-a4bc-d941e0db0c2d
{"action":"get","node":{"key":"/extracted_credentials/ab360ffb-6f0a-4424-a4bc-d941e0db0c2d","value":"{\"credentials\":{\"POSTGRES_DB\":\"test\",\"POSTGRES_HOST\":\"postgresql\",\"POSTGRES_PASSWORD\":\"test\",\"POSTGRES_PORT\":\"5432\",\"POSTGRES_USER\":\"test\"}}","modifiedIndex":27,"createdIndex":27}}

```
Binding mediawiki to postgres is not testable in asbcli. I'm going to test this probably through the service catalog or by fixing asbcli. 

But, I wanted to post this because my code doesn't touch this workflow since the bind credentials are consumed through etcd unless something changed.

jwmatthews:
@rthallisey please test through fusor/catasb using 'dev' branch.

That's the method we should be using to be sure everything is working for an end to end flow.

asbcli is good for developer testing, but we also need to be in habit of going through a full end to end flow with webui and service catalog.

rthallisey:
@jwmatthews sounds good. It hasn't been working for me for the past few days so I'll see if I can figure out the issue.

rthallisey:
I think this is good to go. @eriknelson can you help me test this?  I'm having trouble getting a working UI. I've tested this with the environment I have and the bind credentials make it into etcd.  After that, the service-catalog injects those from etcd into a running pod. So I'm confident this should work since I don't touch that part of the bind.

pkg/apb/ext_creds.go
```diff
@@ -10,82 +10,68 @@ import (
...
-	log.Debug("Calling monitorOutput on " + podname)
-	credOut, _ := monitorOutput(podname, namespace)
-	log.Debug("oc log output: %s", string(credOut))
+	credOut := make(chan []byte)
 
-	var creds *ExtractedCredentials
-	var err error
-	creds, err = buildExtractedCredentials(credOut)
+	log.Debug("Calling monitorOutput on " + podname)
+	go monitorOutput(namespace, podname, credOut, log)
```
jmrodri: +1 at this time and the way we do the work I don't see the need for a go routine to monitor the output.

pkg/apb/ext_creds.go
```diff
@@ -10,82 +10,68 @@ import (
...
-func buildExtractedCredentials(output []byte) (*ExtractedCredentials, error) {
-	stillWaiting := strings.Contains(string(output), "ContainerCreating") ||
-		strings.Contains(string(output), "NotFound")
+func monitorOutput(namespace string, podname string, mon chan []byte, log *logging.Logger) {
+	// TODO: Error handling here
+	// It would also be nice to gather the script output that exec runs
+	// instead of only getting the credentials
+
+	for r := 1; r <= CredentialRetries; r++ {
+		output, _ := RunCommand("oc", "exec", podname, GatherCredentialsCMD, "--namespace="+namespace)
```
jmrodri: 💯 this will totally bite us down the road. I can already foresee the exec erroring out for some odd reason and not having any idea what really happened.

rthallisey: Here's the log if I capture err with ```log.Error(err.Error())``` during a provision:
```
[2017-07-13T10:06:23.988-04:00] [ERROR] exit status 1
[2017-07-13T10:06:23.988-04:00] [WARNING] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] Retry attempt 1: Waiting for container to start
[2017-07-13T10:06:23.988-04:00] [WARNING] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] Retry attempt 1: exec into apb-5834b548-1611-4e5a-a4ee-a799f2fa3018 failed
[2017-07-13T10:06:25.614-04:00] [ERROR] exit status 1
[2017-07-13T10:06:25.614-04:00] [WARNING] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] Retry attempt 2: Waiting for container to start
[2017-07-13T10:06:25.614-04:00] [WARNING] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] Retry attempt 2: exec into apb-5834b548-1611-4e5a-a4ee-a799f2fa3018 failed
[2017-07-13T10:07:08.242-04:00] [ERROR] exit status 137
[2017-07-13T10:07:08.242-04:00] [WARNING] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] Retry attempt 3: exec into apb-5834b548-1611-4e5a-a4ee-a799f2fa3018 failed
[2017-07-13T10:07:11.669-04:00] [ERROR] exit status 1
[2017-07-13T10:07:11.669-04:00] [NOTICE] [apb-5834b548-1611-4e5a-a4ee-a799f2fa3018] APB completed
[2017-07-13T10:07:11.669-04:00] [INFO] Destroying APB sandbox...
```
Since we combine stderr and stdout [1] into one stream, both are being pushed to ```output```. ```err``` is the return value from the command similar to ```echo $?```. So it's probably not a good idea to capture it as anything but info. Also, I don't want to exit on error because the retry logic is in this function.  What I can do is move the retry logic out of this function and return on error.
[1] - https://github.com/openshift/ansible-service-broker/blob/9f8abd0258dbc312a687eba1f2909b02b14e6dad/pkg/apb/hack.go#L12

pkg/broker/broker.go
```diff
@@ -490,6 +475,15 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
 				return nil, err
 			}
 		}
+
+		sm := apb.NewServiceAccountManager(a.log)
```
rthallisey: done

pkg/apb/ext_creds.go
rthallisey: done

pkg/apb/ext_creds.go
```diff
@@ -10,82 +10,68 @@ import (
...
 
-func monitorOutput(podname string, namespace string) ([]byte, error) {
-	return RunCommand("oc", "logs", "-f", "--namespace="+namespace, podname)
+	return buildExtractedCredentials(bindOutput)
 }
 
-func buildExtractedCredentials(output []byte) (*ExtractedCredentials, error) {
-	stillWaiting := strings.Contains(string(output), "ContainerCreating") ||
-		strings.Contains(string(output), "NotFound")
+func monitorOutput(namespace string, podname string, mon chan []byte, log *logging.Logger) {
```
rthallisey: done

pkg/apb/ext_creds.go
rthallisey: I looked at moving out the retry logic and it wouldn't be as clean because some of the monitoring pieces would leak into extractCredentials.

rthallisey:
@eriknelson @jmrodri both broker-bind-output patches have been updated.

pkg/apb/types.go
```diff
@@ -100,6 +100,12 @@ const (
 	// NOTE: Applied to APB ServiceAccount via RoleBinding, not ClusterRoleBinding
 	// cluster-admin scoped to the project the apb is operating in
 	ApbRole = "cluster-admin"
+
+	// Bind credential gathering constants
+	// 5 seconds x 60 retries = 5 minute timout
+	WaitTime             = 5
```
shawn-hurley: Should add comments for exported constants

eriknelson:
Travis is failing due to a failing lint:

```
=================================
              Lint               
=================================
/home/travis/gopath/src/github.com/openshift/ansible-service-broker/pkg/apb/types.go:104:2: comment on exported const WaitTime should be of the form "WaitTime ..."
/home/travis/gopath/src/github.com/openshift/ansible-service-broker/pkg/apb/types.go:107:2: exported const CredentialRetries should have comment (or a comment on this block) or be unexported
The command "./scripts/travis.sh lint" exited with 1.
```
Going to merge this and immediately follow up with a fix for that. Thank you everyone!

rthallisey:
https://github.com/openshift/ansible-service-broker/issues/104

PR #127, opened 2017-05-26
Use the cluster client object to create the apb pod

rthallisey:
Using oc run relies on the oc client. Instead, let's use
the cluster client to talk directly with kubernetes to spawn
the apb pod.

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -8,11 +8,13 @@ import (
 
 	docker "github.com/fsouza/go-dockerclient"
 	logging "github.com/op/go-logging"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 
 	"github.com/pborman/uuid"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/homedir"
+	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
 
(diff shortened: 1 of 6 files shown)


eriknelson:
+1, this is inevitable and it's obvious the broker needs to move in this direction for better native integration. Thanks Ryan.

This is still good for a merge? I'd like to move towards using this API for creating ServiceAccounts and whatever other resources are required, this is a nice precedence.

rthallisey:
@eriknelson Yup this is all set to merge.

jmrodri:
@rthallisey please consult with @cfchase because he just added context support with namespace in it. See https://github.com/openshift/ansible-service-broker/pull/161

cfchase:
@jmrodri Made requested changes.  Since you're unavailable, will pass review on to @eriknelson and @rthallisey.  @rthallisey I grabbed and added the namespace from the context being passed in from the catalog.  I think you could use these changes for your PR https://github.com/openshift/ansible-service-broker/pull/127.

pkg/apb/client.go
```diff
@@ -174,20 +149,69 @@ func (c *Client) RunImage(
...
+		}
 	} else {
-		return RunCommand("oc", "run", fmt.Sprintf("aa-%s", uuid.New()),
-			"--env", fmt.Sprintf("OPENSHIFT_TARGET=%s", clusterConfig.Target),
-			"--env", fmt.Sprintf("OPENSHIFT_USER=%s", clusterConfig.User),
-			"--env", fmt.Sprintf("OPENSHIFT_PASS=%s", clusterConfig.Password),
-			fmt.Sprintf("--image-pull-policy=Always"),
-			fmt.Sprintf("--image=%s", spec.Image), "--restart=Never",
-			"--", action, "--extra-vars", string(params))
+		pod = &v1.Pod{
```
jwmatthews: I'm concerned about the differentiation of running in cluster or out, feels wrong to need to make this choice in the code.

pkg/apb/client.go
```diff
@@ -174,20 +149,69 @@ func (c *Client) RunImage(
...
+						}},
+						ImagePullPolicy: v1.PullAlways,
+					},
+				},
+				RestartPolicy: v1.RestartPolicyNever,
+			},
+		}
 	}
+
+	c.log.Notice(fmt.Sprintf("Creating pod %q in namespace default", pod.Name))
```
jwmatthews: Is this correct that the pod is coming up in the 'namespace default'?

rthallisey: Good catch. This should change based on the Namespace variable.

pkg/apb/client.go
rthallisey: 👍 Me too.

eriknelson:
Ended up building #174 on top of this (thank you @rthallisey !!). We should probably close this one since its functionality is actually being brought in.

rthallisey:
Closing in favor of #174

PR #129, opened 2017-05-30
switch rpm builds to vendor dependencies

jmontleon:


jmontleon:
I was able to recover the rpm builds using individual packages. Closing this... for now.

PR #132, opened 2017-05-31
Get all images

shawn-hurley:


pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+			go r.loadImageData(ctx, "docker://"+imageName.Namespace+"/"+imageName.Name, channel)
+		}
+	}
+
+	// getNextImages - will follow the next URL using go routines.
+	// The workflow is query the url, if Next is defined then kickoff another go routine to get those images.
+	// then kick of the loading of image data.
+	var getNextImages func(ctx context.Context, org, token, url string, ch chan<- *ImageData, cancelFunc context.CancelFunc)
+
+	getNextImages = func(ctx context.Context, org, token, url string, ch chan<- *ImageData, cancelFunc context.CancelFunc) {
```
rthallisey: how about we create this function outside loadBundleImageData? The function loadBundleImageData is a bit overloaded it would nice to pull out some logic wherever we can.

pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+		req, err = http.NewRequest("GET", url, nil)
+		if err != nil {
+			r.log.Errorf("unable to get next images for url: %v - %v", url, err)
+			cancelFunc()
+			close(ch)
+			return
+		}
+		req.Header.Set("Authorization", fmt.Sprintf("JWT %v", token))
```
rthallisey: This code does the same thing as lines 138-163. Maybe we can consolidate that piece into another function that each can use.

pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+		for _, imageName := range images {
+			r.log.Debugf("Trying to load %v/%v", imageName.Namespace, imageName.Name)
+			go r.loadImageData(ctx, "docker://"+imageName.Namespace+"/"+imageName.Name, channel)
+		}
+	}
+
+	// getNextImages - will follow the next URL using go routines.
+	// The workflow is query the url, if Next is defined then kickoff another go routine to get those images.
+	// then kick of the loading of image data.
+	var getNextImages func(ctx context.Context, org, token, url string, ch chan<- *ImageData, cancelFunc context.CancelFunc)
```
rthallisey: Next is essentially the next item in a list of pointers.  What if we had a function that made a request then returned next?  Then this main functions processes the result ->```If next != nil { getNextImages(...) }```

pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
 	}
 
 	channel := make(chan *ImageData)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
```
rthallisey: This is going to kill all requests when ```loadBundleImageData``` returns. I think we either want a timeout context for pure async or we can synchronously load images 100 at a time.

pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
 	}
 
+	// This is looking at the original call to dockerhub for the org.
+	if imageResp.Next != "" {
+		go getNextImages(ctx, org, string(tokenResp.Token), imageResp.Next, channel, cancelFunc)
+	}
+
+	loadImagesFromResult(ctx, imageResp.Results)
+	//If no results in the fist call then close the channel as nothing will get loaded.
+	if len(imageResp.Results) == 0 {
```
rthallisey: We shouldn't need this. If ```next == nil``` then there are no images left.

pkg/apb/dockerhub_registry.go
```diff
@@ -173,20 +169,78 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+		r.log.Info("canceled retrieval as no items in org")
+		cancelFunc()
```
rthallisey: Already have ```defer cancelFunc()``` above. Should really need one or the other.

pkg/apb/dockerhub_registry.go
```diff
@@ -145,48 +131,102 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+		}
 	}
 
-	channel := make(chan *ImageData)
+	// getNextImages - will follow the next URL using go routines.
+	// The workflow is query the url, if Next is defined then kickoff another go routine to get those images.
+	// then kick of the loading of image data.
+	var getNextImages func(ctx context.Context, org, token, url string, ch chan<- *ImageData, cancelFunc context.CancelFunc) (*ImageResponse, error)
+
+	getNextImages = func(ctx context.Context, org, token, url string, ch chan<- *ImageData, cancelFunc context.CancelFunc) (*ImageResponse, error) {
```
shawn-hurley: if you do that, you can't call it from within the same function.

pkg/apb/dockerhub_registry.go
```diff
@@ -145,48 +131,102 @@ func (r *DockerHubRegistry) loadBundleImageData(org string) ([]*ImageData, error
...
+			r.log.Errorf("unable to get next images for url: %v - %v", url, err)
+			cancelFunc()
+			close(ch)
+			return &iResp, err
+		}
+		//Keep getting the images
+		if iResp.Next != "" {
+			r.log.Debugf("getting next page of results - %v", iResp.Next)
+			//Fan out calls to get the next images.
+			go getNextImages(ctx, org, token, iResp.Next, channel, cancelFunc)
```
shawn-hurley: Which threads are you worried about not completing? I don't see anything in the threads that would block them from executing, and we are managing the preemptive closing of the channel. I think if it does fail and preemptively closes the channel do we really want to wait around on the main thread if we don't have to?

If we do add a wait group it will have to to be added to by both the getNextImage method and the loadImageData method. and then presumably we would wait for all to finish at line 251ish?

pkg/apb/dockerhub_registry.go
rthallisey: 👍

pkg/apb/dockerhub_registry.go
rthallisey: I can't come up with a scenario to where the threads hang around. The channel closing looks good.

We may not be able to even use a waitgroup because I think we would need to know the number of threads to wait for. And it could be that when we run it on line 251, some of them have already completed.  I think this patch looks good. If we have a problem later we could always come back and fix it.

PR #136, opened 2017-06-05
add container scripts to a package for proper install

jmontleon:


jmontleon:
We need to install everything from an rpm downstream. This will package the startup script(s) for running in a container so that I can install it properly.

jwmatthews:
ACK

PR #137, opened 2017-06-05
remove deleted file from Dockerfile

jmontleon:


rthallisey:
File was removed in https://github.com/openshift/ansible-service-broker/pull/121. 👍

PR #140, opened 2017-06-05
Fix build deploy

jmrodri:
This change allows the following workflow to work:

```
make clean # remove broker from build now
make build PROJECT=YOUR_DOCKER_HUB_ORG
docker push docker.io/YOUR_DOCKER_HUB_ORG/ansible-service-broker-apb:latest
make deploy # deploys to local oc cluster
```
This workflow uses the settings found in `etc/dev.config.yaml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,13 @@ prepare-build: install
 
 build: prepare-build
 	docker build ${BUILD_DIR} -t ${BROKER_APB_IMAGE}:${TAG}
+	@echo
+	@echo "Remember you need to push your image before calling make deploy"
+	@echo "    docker push ${BROKER_APB_IMAGE}:${TAG}"
 
 clean:
 	@rm -f ${GOPATH}/bin/broker
+	@rm -f build/broker
 
 vendor:
 	@glide install -v
@@ -35,13 +39,13 @@ vendor:
 test: vendor
 	go test ./pkg/...
 
-asb-image:
-	ansible-container build
-	ansible-container push --username $(DOCKERHUB_USER) --password $(DOCKERHUB_PASS) --push-to $(REGISTRY)/$(PROJECT) --tag $(TAG)
-	ansible-container shipit openshift --pull-from $(REGISTRY)/$(PROJECT) --tag $(TAG)
-	# fix bug in ansible-container
-	sed -i 's/-TCP//g' ./ansible/roles/ansible-service-broker-openshift/tasks/main.yml
-	docker build -t $(BROKER_APB_IMAGE) .
-	docker push $(BROKER_APB_IMAGE)
+#asb-image:
+#	ansible-container build
+#	ansible-container push --username $(DOCKERHUB_USER) --password $(DOCKERHUB_PASS) --push-to $(REGISTRY)/$(PROJECT) --tag $(TAG)
+#	ansible-container shipit openshift --pull-from $(REGISTRY)/$(PROJECT) --tag $(TAG)
+#	# fix bug in ansible-container
+#	sed -i 's/-TCP//g' ./ansible/roles/ansible-service-broker-openshift/tasks/main.yml
+#	docker build -t $(BROKER_APB_IMAGE) .
+#	docker push $(BROKER_APB_IMAGE)
 
 .PHONY: run run-mock-registry clean test build asb-image install prepare-build
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 REGISTRY         ?= docker.io
 PROJECT          ?= ansibleplaybookbundle
 TAG              ?= latest
-BROKER_APB_IMAGE = $(REGISTRY)/$(PROJECT)/ansible-service-broker-apb
+BROKER_IMAGE     = $(REGISTRY)/$(PROJECT)/ansible-service-broker
 BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/build"
 
 install: $(shell find cmd pkg)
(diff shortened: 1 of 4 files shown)


jmrodri:
One thing I don't like is that `make build` **ASSUMES** `ansibleplaybookbundle` as the PROJECT.

build/Dockerfile
```diff
@@ -3,6 +3,12 @@ MAINTAINER Ansible Service Broker Community
 
 RUN useradd -M -s /usr/sbin/nologin -u 1001 -g 0 ansibleapp
 
+RUN yum -y update \
```
rthallisey: how about we update after we install all the repos

build/Dockerfile
```diff
@@ -3,6 +3,12 @@ MAINTAINER Ansible Service Broker Community
...
+ && yum -y install epel-release centos-release-openshift-origin \
+ && yum -y install origin origin-clients net-tools bind-utils ansible jq \
```
rthallisey: we don't need jq anymore per https://github.com/openshift/ansible-service-broker/pull/121.  I'll add a patch to fix this everywhere else.

Makefile
```diff
@@ -25,23 +25,27 @@ prepare-build: install
...
 
-asb-image:
-	ansible-container build
-	ansible-container push --username $(DOCKERHUB_USER) --password $(DOCKERHUB_PASS) --push-to $(REGISTRY)/$(PROJECT) --tag $(TAG)
-	ansible-container shipit openshift --pull-from $(REGISTRY)/$(PROJECT) --tag $(TAG)
-	# fix bug in ansible-container
-	sed -i 's/-TCP//g' ./ansible/roles/ansible-service-broker-openshift/tasks/main.yml
-	docker build -t $(BROKER_APB_IMAGE) .
-	docker push $(BROKER_APB_IMAGE)
+#asb-image:
```
rthallisey: If we're not going to use this any more maybe just get rid of it.  We could look in the git history to add it back in if we need to.

templates/broker-deployment_template.yaml
```diff
@@ -18,7 +18,7 @@ spec:
     spec:
       restartPolicy: Always
       containers:
-        - image: "ansibleplaybookbundle/ansible-service-broker-asb:latest"
+        - image: "{{dockerhub_org}}/ansible-service-broker-asb:latest"
```
rthallisey: 👍

rthallisey:
> One thing I don't like is that make build ASSUMES ansibleplaybookbundle as the PROJECT.
>
I know what you mean, but IMO it works as a default in that it's useful for CI & releasing images. Also, it errors when a user tries to use it without permission.  To summarize my opinion, it's not harmful as ```ansibleplaybookbundle```, just a little inconvenient if you don't notice it.

jmrodri:
@rthallisey sadly I think almost everyone has push capabilities to ansibleplaybookbundle. Because we don't do an automatic push it's ok, but I've been bit by it defaulting to ansibleplaybookbundle. Leaving it alone for now.

Makefile
jmrodri: Removed

build/Dockerfile
jmrodri: I remember adding it first because the other packages didn't install. It was a while ago.

build/Dockerfile
jmrodri: So do you need me to remove it from this patch? or will it be removed after this one goes in?

eriknelson:
I don't have an alternative to pitch, but I don't think that org is a sane default. We're viewing it as our "production" location right now, and it's too easy to build and push there, since most of us do have access. We should have explicit commands to target it (`make release`), and set up CI to use those commands so people can't accidentally target it using implicit defaults.

One alternative is to require a variable to be set before running `make build`? Not perfect, open to alternatives.

jmrodri:
TEST
---------
Make sure your `etc/dev.config.yaml` is configured properly:

```
---                                       
registry:
  name: dockerhub
  url: http://localhost:1337
  user: YOUR_DOCKER_HUB_USER
  pass: YOUR_DOCKER_HUB_PASSWORD
  org: YOUR_DOCKER_HUB_ORG
dao:
  etcd_host: localhost
  etcd_port: 2379
log:                                      
  logfile: /tmp/ansible-service-broker.log
  stdout: true
  level: debug
  color: true                             
openshift:
  target: 172.17.0.1:8443
  user: admin
  pass: admin
```
Make sure you have a cluster
```
oc cluster up
# create an admin user
```
Now you can rebuild and deploy the broker:
```
make clean # remove broker from build now
make build PROJECT=YOUR_DOCKER_HUB_ORG
docker push docker.io/YOUR_DOCKER_HUB_ORG/ansible-service-broker-apb:latest
make deploy # deploys to local oc cluster
```
See if the pods came up
```
oc project ansible-service-broker
oc logs asb<TAB>
```

rthallisey:
```make release``` with a default to ansibleplaybookbundles wfm. Then we require PROJECT be set for ```make build```.  👍

**Comment by eriknelson** [comment]
The broker should be able to run as anyuid following best container practices as non-root (apb base is a good example of this). It's not within the scope of this PR though, so I logged #142 to track this separately.

File: build/Dockerfile [hunk 5]
```diff
@@ -1,7 +1,15 @@
...
 MAINTAINER Ansible Service Broker Community
 
-RUN useradd -M -s /usr/sbin/nologin -u 1001 -g 0 ansibleapp
+RUN mkdir /opt/ansibleapp
+RUN useradd -M -b /opt/ansibleapp -s /usr/sbin/nologin -u 1001 -g 0 ansibleapp
+RUN chown -R 1001:0 /opt/ansibleapp
+
+RUN yum -y update \
+ && yum -y install epel-release centos-release-openshift-origin \
+ && yum -y install origin origin-clients net-tools bind-utils ansible \
```
rthallisey: Doesn't oc come from origin-clients? What do you mean by oc?

jmrodri: I believe the deploy.sh was using the `ansibleplaybookbundle/ansible-service-broker-asb` so it never really ran the local one IMO.

eriknelson: I was surprised to see these as additions since they're core dependencies, so I was wondering how this container could have ever worked without them. @jmrodri explained `make deploy` was actually pulling the old `ansibleplaybookbundle/ansible-service-broker-asb` that had been produced by the ansible-container builds instead of new builds produced by this Dockerfile before this change. Makes sense why they needed to be added in this change now.

PR #145, opened 2017-06-06
better facilitate automate copr and manual brew builds

jmontleon:


jmontleon:
I created a copr-macros package within our copr buildroots that's automatically installed for every one of our Copr builds. It drops a macro file that defines copr as 1. With these changes this results in a timestamp being appended to facilitate automatically incrementing the release number. Anywhere else where this macro is not defined the timestamp will be nil, so brew won't be upset because of the release mismatch resulting (because there the SRPM and RPM are built in separate processes one at a time).

PR #148, opened 2017-06-07
rpm improvements and Dockerfile to install from rpm

jmontleon:


jwmatthews:
@jmontleon can we document usage in the repo so others understand?

thinking we call out:

Release:  Docker image is built from RPMs in Copr and include example of commands

Local:  Docker image built from local executable, uses Docker-src and include an example of commands

jmrodri:
This change will break the current Makefile setup. `make build` will try to build from source but use the `Dockerfile` which I believe will blow up.

jmontleon:
@jmrodri, why will it blow up? It will build from source and then created an image installed from rpm which is almost certainly not what's wanted, but don't think it will blow up. Do you want something like a `make release` and `make dev`? I can test something like this today:

````
dev: prepare-build
    docker build ${BUILD_DIR} -f ${BUILD_DIR}/Dockerfile-src -t ${BROKER_IMAGE}:${TAG}
    @echo
    @echo "

release:
    docker build ${BUILD_DIR} -t ${BROKER_IMAGE}:${TAG}
```

rthallisey:
@jmontleon Yes. Can we do ```make release``` and ```make build```? ```make build``` will use Dockerfile-src and ```make release``` will use the lastest rpm?

IMO the rpm building should be a downstream thing and not in this repo at all. ```make release``` should build a container with tag ```$release``` using the source code from repo tag ```$release```. Then, publish that container to the ansibleplaybookbundles dockerhub org. That org will hold stable containers with tags relative to their releases.  Finally, the binary package should be apart of the shipped broker container, which is based off the tagged source. That container will be published in the RHCC. @jwmatthews @jmrodri @eriknelson @jmontleon thoughts?

jmontleon:
Absolutely 100% disagree. We should be packaging our stuff upstream and downstream. Do you really want to drop binaries and files, run pip install in the case of asb-base and tell people that's how we suggest distributing software?

jwmatthews:
In regard to RPM packaging I would prefer it exists upstream as well.

Would like to see the published images to dockerhub be from tagged RPMs so we can trace back through exactly what source was used.   

So...think this means we want the 2 things we've discussed:

release:  uses built rpms from copr
devel/test:  something else that is simpler and is just a local build and putting that executable in the image for local testing

rthallisey:
> Do you really want to drop binaries and files, run pip install in the case of apb-base and tell people that's how we suggest distributing software? 
>
I'm not suggesting we do that.  I'm saying that we separate downstream from upstream. Our primary means of consumption is not a binary, it's a container. So our upstream build process should be focused on container building, while the downstream should be about building the binary and inserting it into the container.

> Would like to see the published images to dockerhub be from tagged RPMs so we can trace back through exactly what source was used.
>
We can still track this because each container published will be the same ```$release``` as the code.`

jmontleon:
@jmrodri this should make it possible to go either route:

```
`$ make release-image
docker build "/home/jmontleo/Documents/openshift-ansible/src/github.com/openshift/ansible-service-broker/build" -t docker.io/ansibleplaybookbundle/ansible-service-broker:latest
Sending build context to Docker daemon 34.77 MB
Step 1/6 : FROM centos:7
 ---> 3bee3060bfc8
Step 2/6 : MAINTAINER Ansible Service Broker Community
 ---> Using cache
 ---> 020d3ab5847d
Step 3/6 : RUN curl https://copr.fedorainfracloud.org/coprs/g/ansible-service-broker/ansible-service-broker/repo/epel-7/group_ansible-service-broker-ansible-service-broker-epel-7.repo -o /etc/yum.repos.d/asb.repo
 ---> Using cache
 ---> 6733d98b4779
Step 4/6 : RUN yum -y update  && yum -y install epel-release centos-release-openshift-origin  && yum -y install origin-clients ansible-service-broker ansible-service-broker-container-scripts  && yum clean all
 ---> Using cache
 ---> 121b4b68196d
Step 5/6 : USER ansibleservicebroker
 ---> Using cache
 ---> 38b8deed04de
Step 6/6 : ENTRYPOINT entrypoint.sh
 ---> Using cache
 ---> 095797e0f79e
Successfully built 095797e0f79e

Remember you need to push your image before calling make deploy
    docker push docker.io/ansibleplaybookbundle/ansible-service-broker:latest
```
$ make dev-image
go install -ldflags="-s -w" ./cmd/broker
cp "/home/jmontleo/Documents/openshift-ansible"/bin/broker build/broker
docker build -f "/home/jmontleo/Documents/openshift-ansible/src/github.com/openshift/ansible-service-broker/build"/Dockerfile-src "/home/jmontleo/Documents/openshift-ansible/src/github.com/openshift/ansible-service-broker/build" -t docker.io/ansibleplaybookbundle/ansible-service-broker:latest
Sending build context to Docker daemon 34.77 MB
Step 1/13 : FROM centos:7
 ---> 3bee3060bfc8
Step 2/13 : MAINTAINER Ansible Service Broker Community
 ---> Using cache
 ---> 020d3ab5847d
Step 3/13 : RUN mkdir /opt/ansibleservicebroker
 ---> Using cache
 ---> dfde31bd1aa7
Step 4/13 : RUN useradd -M -b /opt/ansibleservicebroker -s /usr/sbin/nologin -u 1001 -g 0 ansibleservicebroker
 ---> Using cache
 ---> 020c8ef51db0
Step 5/13 : RUN chown -R 1001:0 /opt/ansibleservicebroker
 ---> Using cache
 ---> 4b96e037b5da
Step 6/13 : RUN yum -y update  && yum -y install epel-release centos-release-openshift-origin  && yum -y install origin-clients net-tools bind-utils  && yum clean all
 ---> Using cache
 ---> e9dedb2c58d2
Step 7/13 : RUN mkdir /var/log/ansible-service-broker     && touch /var/log/ansible-service-broker/asb.log     && mkdir /etc/ansible-service-broker     && mkdir -p /usr/local/ansible-service-broker/bin
 ---> Using cache
 ---> c7d85b88a588
Step 8/13 : RUN chown -R ansibleservicebroker: /var/log/ansible-service-broker     && chown -R ansibleservicebroker: /etc/ansible-service-broker
 ---> Using cache
 ---> 141c9955e3da
Step 9/13 : COPY config.yaml /etc/ansible-service-broker/config.yaml
 ---> Using cache
 ---> 2ec1f61c8119
Step 10/13 : COPY entrypoint.sh ansible-service-broker /usr/bin/
 ---> Using cache
 ---> 2a50e6a648e4
Step 11/13 : COPY broker /usr/bin/asbd
 ---> Using cache
 ---> 1cd3fc4c81aa
Step 12/13 : USER ansibleservicebroker
 ---> Using cache
 ---> 501c0bfe45e7
Step 13/13 : ENTRYPOINT entrypoint.sh
 ---> Using cache
 ---> ecfff5cffa15
Successfully built ecfff5cffa15

Remember you need to push your image before calling make deploy
    docker push docker.io/ansibleplaybookbundle/ansible-service-broker:latest
```

File: Makefile [hunk 1]
```diff
@@ -23,7 +23,13 @@ run-mock-registry: ${GOPATH}/bin/mock-registry vendor
...
 	cp "${GOPATH}"/bin/broker build/broker
 
-build: prepare-build
+dev-image: prepare-build
+	docker build -f ${BUILD_DIR}/Dockerfile-src ${BUILD_DIR} -t ${BROKER_IMAGE}:${TAG}
+	@echo
+	@echo "Remember you need to push your image before calling make deploy"
+	@echo "    docker push ${BROKER_IMAGE}:${TAG}"
+
+release-image:
```
rthallisey: When we build the release image is the expectation that we're using it locally? My understanding is that it would be published.  If that's the case we could just do a ```docker build ...``` and ```docker push ...```.

README.md
```diff
@@ -78,7 +78,8 @@ or the configuration file can be specified by cli args as well.
 * `make run-mock-registry`: Mock registry. Entirely separate binary.
 * `make test`: Runs the test suite.
 * `make vendor`: Updates the dependencies
-* `make build`: Builds a docker container of the current source
+* `make dev-image`: Builds a docker container of the current source
+* `make release-image` Builds a docker container using the latest rpm from [Copr](https://copr.fedorainfracloud.org/coprs/g/ansible-service-broker/ansible-service-broker/)
```
rthallisey: Might be just me, but I'm more a fan of ```make build``` and ```make release```.

jmontleon: When I saw 'build' I initially thought 'go build' not building an image. Build is very ambiguous.

Makefile
jmontleon: Seems like it might be nice to have the opportunity to sanity test it before pushing.

README.md
rthallisey: I can totally see that. My perspective was 'docker build'.

Makefile
rthallisey: Perhaps it should be ```make binary-image```? or ```make build-binary```? I look at release as a tagged release of code. So it's something already tested and stable that will be published with the appropriate tag. I don't view a release as something that's currently under heavy development. I'm not saying that's what you're suggestion, but it's further clarification of my initial comment.

jmontleon: Talked to @jmrodi. Think we're going to do
make release-image
make push
make release: release-image

We may change `make release` to be `make release: release-image push` later but for now we'll leave them as discrete tasks.

Sounds like `make build` might get used for a go build and `make install` changed to install the broker in /usr/local/bin or something so it might be a good time to free up `make build` in case we make that change.

README.md
jmontleon: Talked to @jmrodi. Think we're going to do
make release-image
make push
make release: release-image

We may change `make release` to be `make release: release-image push` later but for now we'll leave them as discrete tasks.

Sounds like `make build` might get used for a go build and `make install` changed to install the broker in /usr/local/bin or something so it might be a good time to free up `make build` in case we make that change.

jmrodri:
#### summary
I'd prefer the following targets (a mishmash of the all of the stuff in background below):

`make build` should run `go install`, or just make build depend on `install` target.
`make deploy` should run deploy.sh
`make build-image` should build the dev image. I would also be ok with just `make dev-image`.
`make release-image` should build the rpm based container image
`make push` would push the image up to whatever registry you want.
`make release` should simply call `release-image` for now. But if we choose to include the push or other steps later, we can simply add it to `release`. The reason for having it now is so I don't have to change my workflow when we add more targets.

Is this acceptable?

#### background
I chatted with Erik on Monday to come up to speed with the current way of building images. That's when we realized that I fixed all the problems he was hitting in my recovery branch. Those got pushed by a separate PR already.  But that discussion led to the following:

meeting notes

The gist was the following:

`make build` should run `go install`
`make deploy` should run deploy.sh
`make build-image` should build the dev image
`make push` would push the image up to whatever registry you want.

In an attempt to start the process for releasing, I sent this in a previous email:
```
* make release
  * this will build the docker image from the tag
  * pushes to dockerhub in ansibleplaybookbundle org
* make build
  * builds the docker image from the latest commit
  * does not push anywhere except local repository
  * change: should default to developers org not ansibleplaybookbundle
```
The above assumed we were doing image builds straight from code. But in light of the fact that there are basically 3 releases:

* create container from local code for dev purposes
* release a container built from a tag from rpms (upstream)
* release a container built from a tag from rpms (downstream)

README.md
jmrodri: @rthallisey AH, that's where you get `build` from the perspective of `docker build`. I always thought of it as compiling a binary which is why it felt weird to run `make build` to build a container.

Makefile
jmrodri: @rthallisey I suggested in another comment that we use a `make release` that calls `release-image`. For now it will *NOT* do a push. I want to make the push explicit. If it becomes the norm that make release is always followed by a push, then we simply update the target to include push.

eriknelson:
@jmrodri +1 that all makes sense to me, where I'd like to see things go.

jmontleon:
FWIW I am trying to separate build/install etc in a separate PR so I think it makes more sense. Will be open to suggestions once it gets there.

README.md
rthallisey: @jmrodri Ya the reason I think of docker build is because the main way to run this stuff is in the container. Where ```make build``` translates to: get my code ready for me to run it. Those steps are compile the code and build and image.

rthallisey:
@jmrodri looks good 👍

Makefile
```diff
@@ -20,15 +20,27 @@ deploy:
 run-mock-registry: ${GOPATH}/bin/mock-registry vendor
 	@${GOPATH}/src/github.com/openshift/ansible-service-broker/cmd/mock-registry/run.sh
 
-prepare-build: install
+prepare-build-image: install
```
rthallisey: Can you update and add all the rules to .PHONY at the bottom.

README.md
```diff
@@ -78,7 +78,9 @@ or the configuration file can be specified by cli args as well.
 * `make run-mock-registry`: Mock registry. Entirely separate binary.
 * `make test`: Runs the test suite.
 * `make vendor`: Updates the dependencies
-* `make build`: Builds a docker container of the current source
+* `make dev-image`: Builds a docker container of the current source
```
jwmatthews: Should this be 'make build-image'   or should we change the target of the Makefile 'build-image' to be 'dev-image'?

jmontleon: make build-image. Fixed.make build-image. Fixed. Thanks for catching it.

jwmatthews:
ACK

PR #150, opened 2017-06-07
create persistent volume claim for etcd data

fabianvf:
Verified that data persists after destroying the pod and allowing the deployment to recreate it.

### DIFF
--- a/ansible-service-broker-all.yaml
+++ b/ansible-service-broker-all.yaml
@@ -62,6 +62,19 @@ spec:
   port:
     targetPort: port-1338
 
+
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: etcd
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
@@ -98,7 +111,13 @@ spec:
           env:
           - name: ETCDCTL_API
             value: "3"
-
+          volumeMounts:
+            - mountPath: /data
+              name: etcd
+      volumes:
+        - name: etcd
+          persistentVolumeClaim:
+            claimName: etcd
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
--- a/templates/etcd-deployment.yaml
+++ b/templates/etcd-deployment.yaml
@@ -1,3 +1,16 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: etcd
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
+---
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
@@ -33,3 +46,11 @@ spec:
           env:
           - name: ETCDCTL_API
             value: "3"
+          volumeMounts:
+            - mountPath: /data
+              name: etcd
+      volumes:
+        - name: etcd
+          persistentVolumeClaim:
+            claimName: etcd
+


fabianvf:
[(3) Ensure our etcd data will be persisted between restarts, node relocation, etc.[asb]](https://trello.com/c/M7ZsnttE/253-3-ensure-our-etcd-data-will-be-persisted-between-restarts-node-relocation-etcasb)

rthallisey:
This is tricky because we're dependant on the user already having a PV set up named etcd with at least 1GB of storage. Maybe etcd should be removed from this template since it requires it's own set up process? If we want an AIO template for folks to curl and run, then it should be a host backed PV that we include in here since we can likely guarantee that will work.

fabianvf:
I think the PV can be named whatever, right? The name is attached to the PVC. Still, it does require the user to have PVs configured. Agree that host backed PV is probably a good idea, do you think we will hit issues with the user id mismatch?

jwmatthews:
I think it's reasonable to assume for tech preview that a PV be available to satisfy the claim on the cluster or we do not deploy, I'm not aware of any requirements on naming or anything other than a 1 GB+ PV be available to satisfy this.

Is the concern with matching user id related to a restart and the user id of the restarted container changing from prior run?

jwmatthews:
One more note on the intent of this change for etcd.

The main driver is to configure etcd for our deployment in tech-preview to be persistent.

We will deploy in one of 2 ways:
 - An all in one openshift template with an "oc process | oc create -f -f"
 - Ansible from openshift installer

For that tech-preview deployment we must have etcd persisted, it is reasonable to have a hard requirement that a PV be available to satisfy this.  That PV should be satisfied as the cluster admin sees fit, i.e. we defer that setup to the admin and just document our need.

For development though...fine if we have an alternate setup that does not require a PV if that makes life easier.  In our testing with 'oc cluster up' we already have a series of PVs created automatically, so testing is easy with the PVC, there are ~100 pre-created 1gb+ pvs available.

I imagine for catasb/oc cluster up testing we will proceed with the full openshift template that deploys a persisted etcd.  (Note the full openshift template is still being developed).

jwmatthews:
@fabianvf I'll test these changes in catasb and update PR later today.

rthallisey:
Looks like when you create a PVC without a PV, the cluster creates a PV for you and binds that volume to your claim. I didn't know that. https://kubernetes.io/docs/concepts/storage/persistent-volumes/#dynamic
```
[rhallisey@localhost ansible-service-broker]$ oc get pvc
NAME        STATUS    VOLUME    CAPACITY   ACCESSMODES   AGE
test-1   Bound     pv2       10Gi       RWO,RWX       1m
[rhallisey@localhost ansible-service-broker]$ oc get pv
NAME      CAPACITY   ACCESSMODES   RECLAIMPOLICY   STATUS      CLAIM                           REASON    AGE
pv1       10Gi       RWO,RWX       Retain          Bound       mediawiki123/mediawiki123-pvc             84d
pv2       10Gi       RWO,RWX       Retain          Bound       default/test-1                         84d
pv4       10Gi       RWO,RWX       Retain          Available                                             84d
```

fabianvf:
Interesting, did you have to change a setting or anything for that to happen? When I created etcd deployment without a PV it just exploded.

rthallisey:
@fabianvf nevermind. I think something weird was happening there with the existing volumes. I cleared out all the pvs and pvcs and tried again and it didn't work.

jwmatthews:
In regard to dynamic PV creation, this can happen...depends on how you configure the cluster.  For example with gluster you can use heketi to handle the dynamic creation of PVs on gluster, for ec2/gce there are other dynamic provisioning plugins.

For "regular" origin clusters we bring up in testing the PV creation is generally _not_ dynamic, so ahead of time the admin creates a number of PVs.  In oc cluster up it defaults to creating 100 PVs, in our vagrant envs in past we'd create perhaps 4-10 PVs...

fabianvf:
Any barrier to merging then?

rthallisey:
@jwmatthews Did you get a chance to test this?

dymurray:
I just tested and saw success with a PV attached to the etcd pod. No problems merging here.

dymurray:
ACK

PR #151, opened 2017-06-08
Cleanup some logging and do not log passwords anymore

rthallisey:
The broker logs the user, cluster address, and password to
stdout.  Don't log the password.

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -141,12 +141,10 @@ func (c *Client) RunImage(
 	if !clusterConfig.InCluster {
 		c.log.Debug("target: [ %s ]", clusterConfig.Target)
 		c.log.Debug("user: [ %s ]", clusterConfig.User)
-		c.log.Debug("password:[ %s ]", clusterConfig.Password)
 	}
 	c.log.Debug("name:[ %s ]", spec.Name)
 	c.log.Debug("image:[ %s ]", spec.Image)
 	c.log.Debug("action:[ %s ]", action)
-	c.log.Debug("extra-vars:[ %s ]", extraVars)
 
 	// It's a critical error if a Namespace is not provided to the
 	// broker because its required to know where to execute the pods and
@@ -269,9 +267,6 @@ func OcLogin(log *logging.Logger, args ...string) error {
 		log.Debug(string(output))
 		return err
 	}
-
-	log.Debug("No error reported after running oc login. Cmd output:")
-	log.Debug(string(output))
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


rthallisey:
[Security cleanup of asb code](https://trello.com/c/FevyaWvn/280-security-cleanup-of-asb-code)

pkg/apb/client.go
```diff
@@ -200,8 +194,5 @@ func (c *Client) refreshLoginToken(clusterConfig ClusterConfig) error {
 	if err != nil {
 		return err
 	}
-
-	c.log.Debug("No error reported after running oc login. Cmd output:")
-	c.log.Debug(string(output))
```
jmrodri: But this is really really useful, especially for debugging. I guess we could add it when we're debugging.

pkg/apb/client.go
```diff
@@ -189,9 +184,8 @@ func (c *Client) refreshLoginToken(clusterConfig ClusterConfig) error {
 	c.log.Debug("Refreshing login token...")
 	c.log.Debug("target: [ %s ]", clusterConfig.Target)
 	c.log.Debug("user: [ %s ]", clusterConfig.User)
-	c.log.Debug("password:[ %s ]", clusterConfig.Password)
```
jmrodri: :+1:

pkg/apb/client.go
```diff
@@ -158,14 +158,9 @@ func (c *Client) RunImage(
...
 
 	c.log.Debug("Running OC run...")
-	c.log.Debug("clusterConfig:")
-	c.log.Debug("target: [ %s ]", clusterConfig.Target)
-	c.log.Debug("user: [ %s ]", clusterConfig.User)
-	c.log.Debug("password:[ %s ]", clusterConfig.Password)
 	c.log.Debug("name:[ %s ]", spec.Name)
 	c.log.Debug("image:[ %s ]", spec.Image)
 	c.log.Debug("action:[ %s ]", action)
-	c.log.Debug("params:[ %s ]", string(params))
```
jmrodri: this is also very useful to see what's getting passed in.

rthallisey: Params can contain the passwords for an app.  Maybe we can output ****** instead.

pkg/apb/client.go
rthallisey: If the oc login errors it will return the error.  Otherwise the output is the just a block of text from an oc login.  Either way we end up with the error message or we print success on line 158.

rthallisey:
https://github.com/openshift/ansible-service-broker/issues/181

eriknelson:
ACK, I think it's reasonable to merge this for TP1 given #181.

PR #153, opened 2017-06-08
fixing run on mac OS X by upping version of containers.

shawn-hurley:
Linux users should be able to just run glide update

mac os x users need to remove /vendor and re-run make  vendor (or glide install -v)


jwmatthews:
This fixed my issue on mac, was able to compile.

jmontleon:
What is github.com/tchap/go-patricia for? This right off the bat will break the rpm builds. 

I just built a new github.com/containers/storage and github.com/opencontainers/selinux to fix the updated runc fedora pushed. I am not too worried about these but it would be good to audit the hashes updated here with the existing packages (https://copr.fedorainfracloud.org/coprs/g/ansible-service-broker/ansible-service-broker/packages/) and figure out which is newer in each case.

Then either reduce/remove library versions to the existing rpm versions (and hopefully still have success on Mac) or update the rpms ahead of time.

jmrodri:
-1 I 'd like to know why gorilla/mux and gorilla/handlers are getting bumped because of a new storage. For my parameter schema work, I had to add jsschema. I found the cleanest way to get it was to run `glide get -v github.com/USER/PROJECT`. This would update the glide.yaml, and the lock file and seemed really good at not updating a bunch of other stuff.

I'm really starting to hate deps. :)

shawn-hurley:
I am working on a fix, where I am moving most of the stuff back down to what they were before and only updating containers

jmontleon:
I have a feeling all that needs bumping is containers/storage (hash on line 28)

And 226-230.

The selinux stuff was split out of containers/storage. I already have that built upstream and the repo requested downstream, so no issue there. We'd just have to update containers/storage again. 

I don't see any references to etcd, gorilla stuff, cleanhttp, etc. Still looking over containers/storage though

jmontleon:
We'll need go-patricia too. storage/layers.go is pulling in the new truncindex sub-package which in turn requires this. @shawn-hurley feel up to updating the containers/storage package upstream in copr and adding golang-github-tchap-go-patricia?

You'll also want to add golang(github.com/thcap/go-patricia) to the BuildRequires in extras/ansible-service-broker.spec with all the rest and Requires for the devel package (same file further down).

jmontleon:
Updated glide files look better. Just fix the RPM spec (add golang(github.com/thcap/go-patricia) to the BuildRequires in extras/ansible-service-broker.spec with all the rest and Requires for the devel package in the same file further down) and once go-patricia is built in copr I think we're good.

jmontleon:
Updated deps are available:
https://copr.fedorainfracloud.org/coprs/g/ansible-service-broker/ansible-service-broker/builds/

+1 

Thanks @shawn-hurley

rthallisey:
> Linux users should be able to just run glide update
>
> mac os x users need to remove /vendor and re-run make vendor (or glide install -v)
>
@shawn-hurley Is this something that should be documented?

djwhatle:
ACK, successfully built on OS X Sierra 10.12.5

eriknelson:
ACK, verified this builds on linux as well. Thanks @shawn-hurley!

PR #154, opened 2017-06-08
log in with serviceaccount certs and token

fabianvf:
Remove openshift_{target|user|password}.

Still trying to figure out how to test

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -119,6 +119,8 @@ func (c *Client) RunImage(
 		return nil, err
 	}
 
+	inCluster := clusterConfig.Target == ""
+
 	////////////////////////////////////////////////////////////////////////////////
 	// This needs a lot of cleanup. Broker was originally written to run
 	// inside a machine that also had a running dockerd available on /var/run/docker.sock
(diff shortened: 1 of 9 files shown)


fabianvf:
[(2) Create a ServiceAccount for running the Ansible Service Broker, update the yaml resource files to leverage the Service Account[asb]](https://trello.com/c/NYTfdLPe/254-2-create-a-serviceaccount-for-running-the-ansible-service-broker-update-the-yaml-resource-files-to-leverage-the-service-accounta)

pkg/apb/client.go
```diff
@@ -149,28 +142,21 @@ func (c *Client) RunImage(
 	//"-e", fmt.Sprintf("OPENSHIFT_PASS=%s", clusterConfig.Password),
 	//spec.Name, action, "--extra-vars", string(params))
 
-	err = c.refreshLoginToken(clusterConfig)
+	err = c.Login()
```
fabianvf: Unless I'm wrong, this shouldn't expire, so I should probably move it to initialization or something.

pkg/apb/client.go
```diff
@@ -185,16 +171,13 @@ func (c *Client) PullImage(imageName string) error {
...
+func (c *Client) Login() error {
+	c.log.Debug("Logging in to openshift...")
 
 	output, err := runCommand(
-		"oc", "login", "--insecure-skip-tls-verify", clusterConfig.Target,
-		"-u", clusterConfig.User,
-		"-p", clusterConfig.Password,
+    "oc", "login", "https://kubernetes.default",
+    "--certificate-authority", "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+    "--token", "$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)",
```
rthallisey: This assumes you're running in the cluster. Are we moving away from allowing someone to run the broker locally?

fabianvf: My current plan was to get to TP with that assumption and then go back and add in the logic that will let us run well in and out of a cluster. Talked with @eriknelson and he seemed to agree. AFAICT you can't really run our official APBs without the catalog at this point, so I didn't think it was a huge compromise.

eriknelson: I actually hadn't thought about this, I thought you were asking about openshift vs k8s. @rthallisey brings up a good point. Throws a wrench into broker development workflows. I wonder if there's a way to fake this out while running locally?

jmrodri:
NACK. I thought we were going to still support the configuration file so that I could run this locally and point to a cluster. That is my developer workflow.

pkg/apb/client.go
jmrodri: +1 to what @eriknelson said about thinking you meant openshift vs k8s.
+1 to what @rthallisey also said about wanting to run locally, that's the way I run my broker 80% of the time until I get the feature ready. Then do final integration test in a containerized env.

jmrodri:
Also, need to run :+1: 
`gofmt -d ./cmd`
`gofmt -d ./pkg`

Fix the errors that get printed out.

pkg/apb/client.go
fabianvf: Doubt it, I can work on making it use the normal credentials if they are provided

pkg/apb/client.go
jwmatthews: Agree, ServiceAccounts should not expire

pkg/apb/client.go
jwmatthews: If we read the info in from a config file it will allow us to tweak for running in cluster or outside.

rthallisey: You can run them with a local broker. Everything is an http request, but the format could change in which case asbcli needs to change.

fabianvf: > You can run them with a local broker. Everything is an http request, but the format could change in which case asbcli needs to change.

You can technically run them, but they'll fail without getting the namespace passed through, and mediawiki currently won't work at all without a successful bind (it no longer accepts the parameters required to manually attach it to a postgresql).

rthallisey: @fabianvf You can check if we're running locally or in cluster here:
https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/client.go#L78
Since every pod in the cluster is given a configfile, we'll check if that exists and use it. If it doesn't, then we use the local cluster config. That should give you the check you need.

After the TP we should really split out these code  paths to make this cleaner.

fabianvf: Moved the serviceaccount login logic to the broker init, since it only needs to happen once, and I just check if the openshift target is specified, and if it isn't try to log in with the service account.

pkg/broker/broker.go
```diff
@@ -53,9 +54,46 @@ func NewAnsibleBroker(
...
+		err := broker.Login()
+		if err != nil {
+			return broker, err
+		}
+	}
+
 	return broker, nil
 }
 
+func (a AnsibleBroker) Login() error {
```
rthallisey: Is there anyway to combine this with refreshLoginToken? These functions do the exact same thing, but different input into oc run.

fabianvf: there might be, though this function only needs to be called once while the refreshLoginToken needs to be called every run. Maybe I could pull out the shared functionality into its own method and have both call it.

fabianvf: there is also some more difference than just the oc run inputs, would have to surround the debug logs with a conditional and still need to read in the token.

fabianvf: @rthallisey took a stab at it, kind of feel like the OcLogin function should live somewhere else but not sure where that would be...

rthallisey: We can re evaluate it's location post TP when we do some code cleanup.  How about write a TODO comment just so we're reminded of it.

fabianvf: Done

PR #155, opened 2017-06-08
Added an OpenShift template for deploying broker & etcd

jwmatthews:
This adds an OpenShift template which shows how to parameterize options.

The run_template.sh is an example of how to deploy the template with parameters.

Currently this is for information purposes.

Next step will be to update catasb to begin using this file opposed to the "all in one" yaml we have in the project root.

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -1,0 +1,177 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: ansible-service-broker
+objects:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: asb
+    labels:
+      app: ansible-service-broker
+      service: asb
+  spec:
+    ports:
+      - name: port-1338
+        port: 1338
+        targetPort: 1338
+        protocol: TCP
+    selector:
+      app: ansible-service-broker
+      service: asb
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: etcd
+    labels:
+      app: ansible-service-broker
+      service: etcd
+  spec:
+    ports:
+      - name: etcd-advertise
+        port: 2379
+    selector:
+      app: ansible-service-broker
+      service: etcd
+- apiVersion: v1
+  kind: Route
+  metadata:
+    name: asb-1338
+    labels:
+      app: ansible-service-broker
+      service: asb
+  spec:
+    to:
+      kind: Service
+      name: asb
+    port:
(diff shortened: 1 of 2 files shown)


templates/run_template.sh
```diff
@@ -0,0 +1,24 @@
+PROJECT="ansible-service-broker"
+BROKER_IMAGE="ansibleplaybookbundle/ansible-service-broker:latest"
+OPENSHIFT_TARGET="https://kubernetes.default"
+OPENSHIFT_USER="admin"
+OPENSHIFT_PASS="admin"
+DOCKERHUB_USER="CHANGEME"
+DOCKERHUB_PASS="CHANGEME"
+DOCKERHUB_ORG="ansibleplaybookbundle"
```
jmrodri: Why is this better than using a config to get the above information?

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -0,0 +1,177 @@
+apiVersion: v1
```
jmrodri: `Can we remove the `ansible-service-broker-all.yaml` ?

templates/run_template.sh
jwmatthews: The benefit is now that the variables are specified as parameter you can use "oc process -p DOCKERHUB_ORG=foo | oc create -f -" to deploy, instead of a combination of sed scripts to modify the yaml.

Example of using a template is here: 
https://github.com/jwmatthews/ansible-service-broker/blob/template/templates/run_template.sh#L24

templates/deploy-ansible-service-broker.template.yaml
jwmatthews: Yes, I'd like to test this template a bit more then modify the PR to move the ansible-service-broker-all.yaml and use this in it's place.

templates/run_template.sh
jwmatthews: Also...as to the Config...the ideal flow will be we remove these parameters and get them from a ConfigMap, the point of this template work was to get us setup to use a template with the code as it is now.  

I assume after move to ConfigMap we will tweak this template to remove some or all of the parameters.

jwmatthews:
Our plan is to merge this in now then add in a few other changes for ConfigMaps that Dylan has put together, as well as incorporate the persistent etcd work Fabian put in.

For next day or so this will be a parallel path with ansible-service-broker-all.yaml

Once we are happy with the template we will move catasb to using it and can remove ansible-service-broker-all.yaml

dymurray:
ACK

PR #157, opened 2017-06-09
Fix makefile

jmontleon:
This needs to be rebased on master once https://github.com/openshift/ansible-service-broker/pull/148 is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,25 +3,44 @@ PROJECT          ?= ansibleplaybookbundle
 TAG              ?= latest
 BROKER_IMAGE     = $(REGISTRY)/$(PROJECT)/ansible-service-broker
 BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/build"
+INSTALL_DIR      ?= /usr/local
 
-install: $(shell find cmd pkg)
-	go install -ldflags="-s -w" ./cmd/broker
+vendor:
+	@glide install -v
 
-${GOPATH}/bin/mock-registry: $(shell find cmd/mock-registry)
-	go install ./cmd/mock-registry
+build: vendor
+	go build -ldflags="-s -w" ./cmd/broker
 
-# Will default run to dev profile
-run: install vendor
-	@${GOPATH}/src/github.com/openshift/ansible-service-broker/scripts/runbroker.sh dev
+install: build
+	cp broker ${INSTALL_DIR}/bin/ansible-service-broker
+	mkdir -p ${INSTALL_DIR}/etc/ansible-service-broker
+	cp etc/ex.dev.config.yaml ${INSTALL_DIR}/etc/ansible-service-broker/config.yaml
 
-deploy:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -3,35 +3,69 @@ PROJECT          ?= ansibleplaybookbundle
...
 	docker build ${BUILD_DIR} -t ${BROKER_IMAGE}:${TAG}
 	@echo
 	@echo "Remember you need to push your image before calling make deploy"
-	@echo "    docker push ${BROKER_IMAGE}:${TAG}"
+	@echo "    make push"
+
+push:
+	docker push ${BROKER_IMAGE}:${TAG}
+
+release: release-image
```
rthallisey: How about putting the contents of release-image into release since ```make release``` does nothing other than run release-image?

Makefile
```diff
@@ -3,35 +3,69 @@ PROJECT          ?= ansibleplaybookbundle
...
+install: build
+	cp broker ${INSTALL_DIR}/bin/ansible-service-broker
+	mkdir -p ${INSTALL_DIR}/etc/ansible-service-broker
+	cp etc/ex.dev.config.yaml ${INSTALL_DIR}/etc/ansible-service-broker/config.yaml
 
-# Will default run to dev profile
-run: install vendor
-	@${GOPATH}/src/github.com/openshift/ansible-service-broker/scripts/runbroker.sh dev
+run: install 
+	${INSTALL_DIR}/bin/ansible-service-broker --config ${INSTALL_DIR}/etc/ansible-service-broker/config.yaml
```
rthallisey: 👍

Makefile
jmontleon: @jmrodri I think the logic was that if we do that and later wanted release to do release & push we wouldn't need to change the workflow you would still just run release in the future and not have to change your workflow. 

Granted we could just put the docker push command under release at that point, but this would also allows them to be run as separate steps if you want to check the image in between steps if you desired.

I think I was tarcking @jmrodi's advice but if I did it wrong/too convoluted happy to fix.

rthallisey: 👍

jmrodri:
This breaks travis. Look at `scripts/travis.sh` at the build step. Change `make install` in that script to be `make build`.

rthallisey:
@jmontleon Is this still WIP? Or is it good to go?

jmontleon:
I need to rebase it after #148 is merged. It says it's ready to go, but has not been merged yet.

README.md
```diff
@@ -65,23 +65,39 @@ make vendor
...
-to `etc/dev.config.yaml`, and edit as desired. `scripts/runbroker.sh` should
-handle providing the location to this file. Of course, this can be customized
-or the configuration file can be specified by cli args as well.
+to `etc/config.yaml`, and edit as desired. 
+
 
-## Targets
 
+## Targets
+### Broker Targets
```
rthallisey: This might be too much, but do you think we could sort some of these into a developer targets section?

rthallisey:
👍

PR #158, opened 2017-06-09
Update RHCC registry with new query and logic to handle Spec creation

dymurray:


pkg/apb/rhcc_registry.go
```diff
@@ -63,19 +63,73 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
+	}
+
+	/*	if hist.History == nil {
+		r.log.Error("V1 Schema Manifest does not exist in registry")
+		return nil, nil
+	}*/
+
+	err = json.Unmarshal([]byte(hist.History[0]["v1Compatibility"]), &conf)
+	if err != nil {
+		r.log.Error("Error unmarshalling intermediary JSON response")
```
shawn-hurley: Can we print the error out here? I think that helps while debugging if something goes wrong

pkg/apb/rhcc_registry.go
```diff
@@ -63,19 +63,73 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
+		return nil, err
+	}
+
+	r.log.Debug(r.config.Url)
+	encodedSpec := conf.Config.Label.Spec
+	decodedSpecYaml, _err := b64.StdEncoding.DecodeString(encodedSpec)
```
shawn-hurley: I would not create a new error variable here, I think that it should be err, imo

pkg/apb/rhcc_registry.go
```diff
@@ -63,19 +63,73 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
+		Version string `json:"com.redhat.apb.version"`
+	}
+
+	type config struct {
+		Label label `json:"Labels"`
+	}
+
+	hist := struct {
+		History []map[string]string `json:"history"`
+	}{}
```
jmrodri: Why are these all private structs defined within the method? I would've made them normal structs at the top of the file.

pkg/apb/rhcc_registry.go
```diff
@@ -63,19 +63,73 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
+		return nil, err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	type label struct {
```
rthallisey: Part of this is in ```pkg/apb/types/go``` https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/types.go#L15.

The way you're reading in the json IMO is better.  Right now we use the string 'com.redhat.apb.spec' as a map key when we should really be decoding into a struct. We could either leave this as additional tech debt or try and fix it now. I might take a crack at doing this tomorrow.

jmrodri: Ah I see, I'd be ok with it being in the struct instead of being a map key. That makes sense to me. :+1:

rthallisey:
Update here: After some testing it looks like RHCC images can be inspected. This change may just be a config file change to point at the correct registry.

dymurray:
https://github.com/openshift/ansible-service-broker/issues/180

Above issue is used to track Ryan's comment that there is overlapping logic here we can make a lot cleaner for using different registries.

pkg/apb/rhcc_registry.go
dymurray: I made these private structs because they are intermediary objects I need to parse the JSON down to the spec which gives us the labels. I figured this data object wouldn't be used outside of this function.

rthallisey: I'm fine with leaving everything in this one file. I'd rather we not touch anything outside since this will change post TP.

dymurray: +1 I agree.

eriknelson:
ACK

PR #161, opened 2017-06-09
Add namespace parameter from service context.

cfchase:


jmrodri:
looks ok to me.

pkg/apb/client.go
```diff
@@ -172,24 +173,39 @@ func (c *Client) RunImage(
...
 			fmt.Sprintf("--image-pull-policy=Always"),
 			fmt.Sprintf("--image=%s", spec.Image), "--restart=Never",
-			"--", action, "--extra-vars", string(params))
+			"--", action, "--extra-vars", extraVars)
 	}
 }
 
+// TODO: Instead of putting namespace directly as a parameter, we should create a dictionary
+// of apb_metadata and put context and other variables in it so we don't pollute the user
+// parameter space.
```
cfchase: Will address TODO in a separate PR with apb-examples since they need changes

cfchase:
[(3) Need a mechanism to pass in namespace to apbs [apb] [asb]](https://trello.com/c/I8VOLN6K/230-3-need-a-mechanism-to-pass-in-namespace-to-apbs-apb-asb)

jmrodri:
Here's a test file `pkg/apb/client_test.go`. To run just the apb tests run: `go test -v ./pkg/apb`. Once you get this passing, then you can run `make test`. 

Currently the method panics if you pass in a nil parameters.

```
package apb

import (
	"testing"

	ft "github.com/openshift/ansible-service-broker/pkg/fusortest"
)

func TestNilPanics(t *testing.T) {

	context := &Context{Platform: "kubernetes", Namespace: "testing-project"}
	t.Log("calling createExtraVars")
	value, err := createExtraVars(context, nil)
	if err != nil {
		t.Log("we have an error")
		t.Fatal(err)
	}

	t.Log(value)
	ft.AssertEqual(t, value, "", "invalid value")
}

func TestCreateExtVars(t *testing.T) {

	context := &Context{Platform: "kubernetes", Namespace: "testing-project"}
	parameters := &Parameters{"key": "param"}
	value, err := createExtraVars(context, parameters)
	if err != nil {
		t.Log("we have an error")
		t.Fatal(err)
	}

	expected := "{\"key\":\"param\",\"namespace\":\"testing-project\"}"
	ft.AssertEqual(t, value, expected, "extravars do not match")
}
```

jmrodri:
@cfchase please consult with @rthallisey because he has a PR that does something with namespace as well. See https://github.com/openshift/ansible-service-broker/pull/127

jmrodri:
Change looks spec compliant. :+1:

jmrodri:
I would update `test/provision*.sh` files to pass in the context:
```
curl http://username:password@broker-url/v2/service_instances/:instance_id?accepts_incomplete=true -d '{
  "context": {
    "platform": "cloudfoundry",
    "some_field": "some-contextual-data"
  },
  "service_id": "service-guid-here",
  "plan_id": "plan-guid-here",
  "organization_guid": "org-guid-here",
  "space_guid": "space-guid-here",
  "parameters": {
    "parameter1": 1,
    "parameter2": "foo"
  }
}' -X PUT -H "X-Broker-API-Version: 2.11" -H "Content-Type: application/json"
```

cfchase:
@jmrodri made requested changes.  @rthallisey @eriknelson Will pass code review on to you two since I believe your changes are dependent on this PR.

pkg/broker/types.go
```diff
@@ -60,6 +60,7 @@ type ProvisionRequest struct {
 	PlanID            uuid.UUID      `json:"plan_id"`
 	ServiceID         uuid.UUID      `json:"service_id"`
 	SpaceID           uuid.UUID      `json:"space_guid"`
+	Context           apb.Context    `json:"context,omitempty"`
```
rthallisey: This might cause issues if we accept an empty Context.

pkg/apb/client.go
```diff
@@ -227,3 +228,21 @@ func OcLogin(log *logging.Logger, args ...string) error {
 	log.Debug(string(output))
 	return nil
 }
+
+// TODO: Instead of putting namespace directly as a parameter, we should create a dictionary
+// of apb_metadata and put context and other variables in it so we don't pollute the user
+// parameter space.
+func createExtraVars(context *Context, parameters *Parameters) (string, error) {
+	var paramsCopy Parameters
+	if parameters != nil && *parameters != nil {
```
rthallisey: Why are we checking both ```parameters != nil && *parameters != nil```?  parameters is ```*Parameters```.

pkg/apb/client.go
```diff
@@ -227,3 +228,21 @@ func OcLogin(log *logging.Logger, args ...string) error {
...
+		paramsCopy = *parameters
+	} else {
+		paramsCopy = make(Parameters)
+	}
+
+	if context != nil {
```
rthallisey: What if ```context == nil```? Below we allow this to be empty, but up here we don't set a default value for namespace.

pkg/apb/provision.go
```diff
@@ -13,7 +13,7 @@ import (
 // Maybe apb defines its own interface and accepts that optionally
```
rthallisey: Do we also need the context for bind and unbind? Right now bind does an ```oc logs <podname>```. If that pod is not in the default namespace, bind won't work.

pkg/apb/types.go
```diff
@@ -45,6 +45,11 @@ type Spec struct {
 	Parameters []*ParameterDescriptor `json:"parameters"`
 }
 
+type Context struct {
+	Platform  string `json:"platform"`
```
rthallisey: Just curious, what are the options for platform?

pkg/apb/client.go
cfchase: I don't think we should set a default for namespace.  If an APB expects it and it doesn't get passed in, it should fail.  If it doesn't expect it, it shouldn't get a default one.

pkg/apb/types.go
cfchase: All I've seen so far is "kubernetes" and I haven't found the spec to see what else is possible.
```json
{
  "organization_guid": "996e99e2-4c7f-11e7-acf8-54ee755dba17",
  "plan_id": "4c10ff42-be89-420a-9bab-27a9bef9aed8",
  "service_id": "55c53a5d-65a6-4c27-88fc-e027410b1882",
  "space_guid": "996e99e2-4c7f-11e7-acf8-54ee755dba17",
  "accepts_incomplete": true,
  "context": {
    "platform": "kubernetes",
    "namespace": "hello-world"
  }
}
```

pkg/apb/client.go
cfchase: Unfortunately, I ran into the case of "bind" in testing where the parameters wasn't nil, but was pointing to an uninitialized map.  in this case `parameters != nil` but `*parameters == nil`.

Jesus pointed out that if nil could be passed in as the `parameters` argument as well as he showed in his unit test, so I should also check for that.

pkg/apb/provision.go
cfchase: There is no context object supplied during the bind request.  It uses the namespace saved on the service instance during the provision.  This made sense to me since we wouldn't want something different from the actual namespace where the service lives.

https://github.com/cfchase/ansible-service-broker/blob/103d27d9d67e6d44a7aea0d7a9cb94b038cabb2d/pkg/apb/bind.go#L34

https://github.com/cfchase/ansible-service-broker/blob/103d27d9d67e6d44a7aea0d7a9cb94b038cabb2d/pkg/broker/broker.go#L223

pkg/apb/client.go
shawn-hurley: He has to check both, because a pointer to a map, is not very useful. that really should be ```parameters Parameters``` and not a pointer as maps are [already what you would be looking for](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it) with a pointer.

pkg/broker/types.go
cfchase: I'm sure what to do with it at this point before the Context spec is a little more firm.  Some of our APBs default the namespace to a newly created project if absent so when writing this I wanted that to work.  

I think we should just be returning a 400 in the handler if it's not included. Perhaps we should wait until the spec is finalized though and add the check in the handler if it's a "required" field.

pkg/apb/client.go
rthallisey: I agree we shouldn't set a default namespace. But, I think we want to get rid of omitempty above so that we error before we get to here.

cfchase: Good call.  Done.

jmrodri: Yep spec expects passing context

PR #166, opened 2017-06-12
configuration option to Broker to control launch_apb_on_bind

johnkim76:


johnkim76:
[Update Broker to incorporate workaround for lack of async bind.](https://trello.com/c/BQerOj86/276-update-broker-to-incorporate-workaround-for-lack-of-async-bind)

pkg/app/app.go
```diff
@@ -121,7 +121,7 @@ func CreateApp() App {
 
 	app.log.Debug("Creating AnsibleBroker")
 	if app.broker, err = broker.NewAnsibleBroker(
-		app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine,
+		app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.LaunchApbOnBind,
```
jmrodri: -1 instead of passing in a boolean, we should be looking to pass in a Broker config. That will give us flexibility to add other settings.

pkg/app/config.go
```diff
@@ -11,12 +11,13 @@ import (
...
-	Openshift  apb.ClusterConfig
-	ConfigFile string
-	DevBroker  bool
+	Registry        apb.RegistryConfig
+	Dao             dao.Config
+	Log             LogConfig
+	Openshift       apb.ClusterConfig
+	ConfigFile      string
+	DevBroker       bool
+	LaunchApbOnBind bool
```
jmrodri: I would move DevBroker & LaunchApbOnBind to a BrokerConfig struct. Then make Config struct have a
BrokerConfig.

```
type Config struct {
...
BrokerConfig BrokerConfig
}
```

Then make BrokerConfig:
```
type BrokerConfig struct {
DevBroker bool
LaunchApbOnBind bool
}
```

pkg/broker/broker.go
```diff
@@ -37,11 +37,12 @@ type Broker interface {
...
-	log           *logging.Logger
-	clusterConfig apb.ClusterConfig
-	registry      apb.Registry
-	engine        *WorkEngine
+	dao            *dao.Dao
+	log            *logging.Logger
+	clusterConfig  apb.ClusterConfig
+	registry       apb.Registry
+	engine         *WorkEngine
+	lauchApbOnBind bool
```
jmrodri: Instead of passing in lauchApbOnBind (which has a typo by the way), it should pass in `brokerConfig BrokerConfig`.

pkg/broker/broker.go
```diff
@@ -428,9 +431,12 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
 		a.log.Debug("%+v", provExtCreds)
 	}
 
-	bindExtCreds, err := apb.Bind(instance, &params, a.clusterConfig, a.log)
-	if err != nil {
-		return nil, err
+	bindExtCreds := &apb.ExtractedCredentials{Credentials: make(map[string]interface{})}
+	if a.lauchApbOnBind {
```
jmrodri: `if a.brokerConfig.LaunchApbOnBind {` instead.

pkg/app/app.go
johnkim76: will update.

pkg/broker/broker.go
johnkim76: will update

pkg/broker/broker.go
johnkim76: will update

pkg/broker/broker.go
johnkim76: thanks for catching the typo... will update

pkg/app/config.go
johnkim76: will update

etc/ex.dev.config.yaml
```diff
@@ -17,4 +17,7 @@ openshift:
   target: localhost:8443
   user: OPENSHIFT_USER
   pass: OPENSHIFT_PASS
-devbroker: true
+broker:
+  devbroker: true
+  launchapbonbind: false
```
rthallisey: There are more config files than the dev one.  Personally, I think we should only have one config file for the broker. Let's go with just this change and I'll follow up to remove the other config files and add a doc explaining the use cases. Created https://github.com/openshift/ansible-service-broker/issues/170.

PR #167, opened 2017-06-13
add context support

jmrodri:


scripts/asbcli/asbcli
```diff
@@ -460,14 +468,16 @@ class App(object):
 # 24   -v \
 # 25   http://cap.example.com:1338/v2/service_instances/$instanceUUID
 ############################################################
-    def _req_provision(self, service_id, answers):
+    def _req_provision(self, service_id, answers, namespace):
```
rthallisey: 👍 

I think we also need this for bind, unbind, and deprovision.  Any place that launches an apb container will require a namespace.

eriknelson: Absolutely. The broker expects a namespace on every op as of TP1. I'd really like to keep asbcli up-to-date if its still helpful for people, so this is good stuff. I started to patch this myself, but I'm hoping to catch up with @jmrodri to talk about what we need to do to keep asbcli alive (much to my chagrin :P)

jmrodri: bind, unbind, and deprovision get their namespaces from the service instance that the broker is caching: `instance.Context`.  There is no support for passing in namespace for bind, unbind,  & deprovision. Also the openservicebroker api does not define a context for those methods.

eriknelson: @jmrodri You're totally right. Broker does get it from the stored instance.

eriknelson:
I think this is ready to go, ACK.

jmrodri:
@rthallisey can you re-review, update the review above.

PR #168, opened 2017-06-13
Added persistent volume claim for etcd and config map updates

dymurray:


templates/run_template.sh
```diff
@@ -6,9 +6,12 @@ OPENSHIFT_PASS="admin"
 DOCKERHUB_USER="CHANGEME"
 DOCKERHUB_PASS="CHANGEME"
 DOCKERHUB_ORG="ansibleplaybookbundle"
+REGISTRY_TYPE="dockerhub"
+REGISTRY_URL="docker.io"
+DEV_BROKER="true"
```
rthallisey: I think this depends on https://github.com/openshift/ansible-service-broker/pull/166

dymurray: Ah, thank you I will fix my config syntax too.

jwmatthews:
@dymurray I tested deploying this from catasb, basic deployment checked out.

rthallisey:
Ok just merged #166. @jwmatthews do you want to test again?

jwmatthews:
I've tested what I can, deployment of broker/etcd is coming up.  Having issues with provision of APB through broker, but I think that is to be expected at moment.

I'm good with merging this and continue to debug.

PR #171, opened 2017-06-13
Move to using only a single config file

### LINKED ISSUE #170: Reduce the number of broker config files to one
We have 4 broker config files in the etc directory. IMO we should only carry one, but document each use case.

rthallisey:
We are carry four different broker config files. We should move to using one config file and edit it to use different broker configurations.

Fixes #170


rthallisey:
Missing: Using a config file in the config map template

eriknelson:
LGTM, I originally split out separate env files for potentially having different run profiles but the need for that never really surfaced and it complicated things.

build/broker-config.yaml
```diff
@@ -9,11 +9,14 @@ dao:
   etcd_host: etcd
   etcd_port: 2379
 log:
-  logfile: /var/log/ansible-service-broker/asb.log
+  logfile: /tmp/ansible-service-broker.log
```
eriknelson: Why move the log file out of /var/log?

rthallisey: I copied over etc/dev.config.yaml. This should probably /var/log everywhere.

eriknelson: Ah, sounds good.

eriknelson:
ACK. spoke with @rthallisey and @jwmatthews. Everyone is happy with this change.

PR #174, opened 2017-06-14
Sandboxed APB Implementation

eriknelson:
Sandboxes apb with dynamically created ServiceAccounts and RoleBindings.

I'm not very happy with the way DestroyApbSandbox works here. We need to revisit it since authors working in the broker.go should should *not* have to worry about leaking the sandboxes; they should always be cleaned up on the conclusion of the pods. I'd personally like to see this get implemented as an event handler that executes on some kind of a pod terminate trigger, but I'm unable to get that in by tomorrow. I plan on following this up with an issue to revisit post TP1.

### DIFF
--- a/etc/ex.mock.config.yaml
+++ b/etc/ex.mock.config.yaml
@@ -17,3 +17,6 @@ openshift:
   target: {{OPENSHIFT_TARGET}}
   user: {{OPENSHIFT_USER}}
   pass: {{OPENSHIFT_PASS}}
+broker:
+  devbroker: true
+  launchapbonbind: false
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -8,11 +8,13 @@ import (
 
 	docker "github.com/fsouza/go-dockerclient"
 	logging "github.com/op/go-logging"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 
 	"github.com/pborman/uuid"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/homedir"
+	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
 
@@ -113,44 +115,15 @@ func (c *Client) RunImage(
 	spec *Spec,
 	context *Context,
 	p *Parameters,
-) ([]byte, error) {
+) (string, error) {
 	// HACK: We're expecting to run containers via go APIs rather than cli cmds
 	// TODO: Expecting parameters to be passed here in the future as well
 
 	extraVars, err := createExtraVars(context, p)
+
 	if err != nil {
-		return nil, err
+		return "", err
 	}
-
-	////////////////////////////////////////////////////////////////////////////////
-	// This needs a lot of cleanup. Broker was originally written to run
-	// inside a machine that also had a running dockerd available on /var/run/docker.sock
-	// If the broker is running inside of a container on ocp or k8s, this docker runtime
-	// is not available to the broker's environment in the same way. This requires
-	// the broker to run the apb metacontainer remotely...somehow.
-	// * options for docker are ugly:
-	// -> Nested docker? (bad)
-	// -> Remote docker (not a ton of options?) TCP socket?
-	// * We know we've got an available ocp cluster _somewhere_. Use oc run instead of
-	// docker? Requires an auth'd oc client available to the broker, but that can
-	// be baked into the broker's container runtime. This is done as a temporary soln.
-	//
-	// TODO: Need to figure out the right way to accomplish running metacontainers
-	// in remote runtimes longterm.
-	////////////////////////////////////////////////////////////////////////////////
-	//oc run ansible-service-broker-apb --env "OPENSHIFT_TARGET=10.1.2.2:8443"
-	//--env "OPENSHIFT_USER=admin" --env "OPENSHIFT_PASS=derp"
-	//--image=apb/ansible-service-broker-ansibleapp --restart=Never --
-	//provision -e "dockerhub_user=eriknelson" -e "dockerhub_pass=derp"
-	//-e "openshift_target=10.1.2.2:8443" -e "openshift_user=admin" -e "openshift_pass=derp"
-
-	// NOTE: Older approach when docker is easily available to the broker to run
-	// metacontainers, i.e., just running on
-	//return RunCommand("docker", "run",
-	//"-e", fmt.Sprintf("OPENSHIFT_TARGET=%s", clusterConfig.Target),
-	//"-e", fmt.Sprintf("OPENSHIFT_USER=%s", clusterConfig.User),
-	//"-e", fmt.Sprintf("OPENSHIFT_PASS=%s", clusterConfig.Password),
-	//spec.Name, action, "--extra-vars", string(params))
 
 	if !clusterConfig.InCluster {
 		err = c.refreshLoginToken(clusterConfig)
@@ -158,12 +131,11 @@ func (c *Client) RunImage(
 		if err != nil {
 			c.log.Error("Error occurred while refreshing login token! Aborting apb run.")
 			c.log.Error(err.Error())
-			return nil, err
+			return "", err
 		}
 		c.log.Notice("Login token successfully refreshed.")
 	}
 
-	c.log.Debug("Running OC run...")
 	c.log.Debug("clusterConfig:")
 	if !clusterConfig.InCluster {
 		c.log.Debug("target: [ %s ]", clusterConfig.Target)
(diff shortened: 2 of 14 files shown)


jwmatthews:
Tested with rhscl-postgres-apb
Initial testing, got further and saw APB launched, but APB was launched in 'default' namespace for me.

Broker:
[INFO] Checking if project demo exists...
....
....

[NOTICE] Creating pod "apb-c2a77414-7573-4fa3-8d3c-e96f098e441b" in the default namespace



I am testing with etcd from quay, persisted and using the newer template, so could be another variable is altering the behavior I'm seeing.  


Ultimately APB failed to complete:

$ oc logs -f apb-c2a77414-7573-4fa3-8d3c-e96f098e441b
Got OPENSHIFT credentials.
Login successful.
You have access to the following projects and can switch between them with 'oc project <projectname>':
    ansible-service-broker
  * default
    demo
    kube-public
    kube-system
    myproject
    openshift
    openshift-infra
    service-catalog
Using project "default".
Welcome! See 'oc help' to get started.
PLAY [Deploy rhscl-postgresql-apb to openshift] ********************************
TASK [ansible.kubernetes-modules : Intall latest openshift client] *************
skipping: [localhost]
TASK [rhscl-postgresql-apb-openshift : create namespace] ***********************
ok: [localhost]
TASK [rhscl-postgresql-apb-openshift : set volume state to present] ************
changed: [localhost]
TASK [rhscl-postgresql-apb-openshift : set service state to present] ***********
changed: [localhost]
TASK [rhscl-postgresql-apb-openshift : scale deployment down] ******************
skipping: [localhost]
TASK [rhscl-postgresql-apb-openshift : set deployment config state to present] ***
changed: [localhost]
TASK [rhscl-postgresql-apb-openshift : delete replication controller] **********
skipping: [localhost]
TASK [rhscl-postgresql-apb-openshift : Wait for postgres to come up] ***********
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for 172.30.204.198:5432"}

jwmatthews:
Below is a sample of the config I tried from the template, tried changing devbroker, didn't see it impact behavior.  Still seeing APB launched in 'default'

$ oc exec asb-4137035260-xxj0z -it /bin/bash
bash-4.2$ cat /etc/ansible-service-broker/config.yaml 
registry:
  name: dockerhub
  url: docker.io
  user: REMOVED
  pass: REMOVED
  org: ansibleplaybookbundle
dao:
  etcd_host: etcd
  etcd_port: 2379
log:
  logfile: /var/log/ansible-service-broker/asb.log
  stdout: true
  level: debug
  color: true
openshift:
  target: kubernetes.default
  user: admin
  pass: admin
broker:
  devbroker: False
  launchapbonbind: false

eriknelson:
The presence of openshift.target in the config will trigger [this branch](https://github.com/openshift/ansible-service-broker/blob/12846db61cb72081d476d57add0b023298b6919a/pkg/apb/client.go#L195) and cause the APB pod to run in the default namespace due to cluserConfig.InCluster == true. That config value is set here:

https://github.com/openshift/ansible-service-broker/blob/12846db61cb72081d476d57add0b023298b6919a/pkg/app/config.go#L50

We need to get rid of that second branch but we'd like to address it post TP1.

jwmatthews:
Success, issue was I going down the 'local' path of the Broker, so it was using the 'default' namespace.
We should discuss this local path, feels problematic.
The change I needed was to set the config file to not have any openshift parameters in it.

Sample info here:
https://gist.github.com/jwmatthews/519bf2b77af28e37f73c17b2fa25ef0b

jwmatthews:
Recap of my testing:

- Using a modified openshift template with a configmap to write the broker config file
- Using a patch in catasb to render the openshift template
- Using etcd from quay and persistend storage

(the etcd we built from ansibleapp with persistent storage was causing issues with broker, looked like a restart after 90 seconds from etcd start was causing some data to be lost and crashing broker, havent seen issue since switching to etcd from quay)

pkg/broker/broker.go
```diff
@@ -439,11 +450,21 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
...
+
+		sm := apb.NewServiceAccountManager(a.log)
+		a.log.Info("Destroying APB sandbox...")
+		sm.DestroyApbSandbox(podName, instance.Context.Namespace)
+
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		a.log.Warning("Broker configured to *NOT* launch and run APB bind")
```
rthallisey: If we're defaulting LaunchApbOnBind to false, should we warn about it at all? Because isn't the assumption we don't want folks to turn it on risking potential timeouts?

etc/ex.mock.config.yaml
```diff
@@ -17,3 +17,6 @@ openshift:
   target: {{OPENSHIFT_TARGET}}
   user: {{OPENSHIFT_USER}}
   pass: {{OPENSHIFT_PASS}}
+broker:
```
rthallisey: Patch #171 will deal with cleaning up these config files. Don't worry about fixing these.

fabianvf:
[(5) Update ASB to create a dynamic ServiceAccount per APB action invoked.[asb]](https://trello.com/c/wBdC732J/255-5-update-asb-to-create-a-dynamic-serviceaccount-per-apb-action-invokedasb)

pkg/apb/client.go
```diff
@@ -175,20 +147,89 @@ func (c *Client) RunImage(
 	c.log.Debug("action:[ %s ]", action)
 	c.log.Debug("extra-vars:[ %s ]", extraVars)
 
+	// TODO(rhallisey): Cleanup the local vs incluster code paths so there
+	// is almost no variation between the two.
+	//
+	// If we're using a locally running broker we want to use the default
+	// namespace.
+	ns := context.Namespace
```
eriknelson: That's an asbcli change? And 155 just aliases the namespace, definitely needed. Did you mean just remove the comment?

pkg/apb/client.go
```diff
@@ -175,20 +147,89 @@ func (c *Client) RunImage(
...
 	} else {
-		return RunCommand("oc", "run", fmt.Sprintf("aa-%s", uuid.New()),
-			"--env", fmt.Sprintf("OPENSHIFT_TARGET=%s", clusterConfig.Target),
-			"--env", fmt.Sprintf("OPENSHIFT_USER=%s", clusterConfig.User),
-			"--env", fmt.Sprintf("OPENSHIFT_PASS=%s", clusterConfig.Password),
-			fmt.Sprintf("--image-pull-policy=Always"),
-			fmt.Sprintf("--image=%s", spec.Image), "--restart=Never",
-			"--", action, "--extra-vars", extraVars)
+		// Set the namespace to default for the local broker.
+		ns = "default"
```
eriknelson: Not sure what you mean.

pkg/apb/svc_acct.go
```diff
@@ -0,0 +1,197 @@
...
+// Sets up ServiceAccount based apb sandbox
+// Returns service account name to be used as a handle for destroying
+// the sandbox at the conclusion of running the apb
+func (s *ServiceAccountManager) CreateApbSandbox(
+	namespace string,
+	apbId string,
+) (string, error) {
+	svcAccountName := apbId
+	roleBindingName := apbId
+	apbRole := "cluster-admin"
```
eriknelson: I deliberately put it where it will be used because this is going to change when apbs themselves request the roles that they need for running.

pkg/apb/svc_acct.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (s *ServiceAccountManager) createFile(apbId string) string {
+	rFilePath := filepath.Join(resourceDir(), apbId+".yaml")
+	s.log.Debug("Creating resource file %s", rFilePath)
+	var _, err = os.Stat(rFilePath)
+
+	if os.IsNotExist(err) {
+		var file, err = os.Create(rFilePath)
+		if err != nil {
+			s.log.Error("Something went wrong touching resource file!")
```
eriknelson: Good catch.

pkg/apb/svc_acct.go
```diff
@@ -0,0 +1,197 @@
...
+		defer file.Close()
+	}
+	return rFilePath
+}
+
+func (s *ServiceAccountManager) DestroyApbSandbox(
+	handle string,
+	namespace string,
+) error {
+	if handle == "" {
```
eriknelson: Yes, any of the apb methods that run but have error'd out before creating the sandbox return "" for the podName. I'm not happy with it, but this is something I want to come back to after TP1 to do sandbox destruction the right way (some kind of pod termination event handler).

pkg/apb/svc_acct.go
```diff
@@ -0,0 +1,197 @@
...
+		"oc", "delete", "rolebinding", handle, "--namespace="+namespace,
+	)
+	if err != nil {
+		s.log.Error("Something went wrong trying to destroy the rolebinding!")
+		s.log.Error(err.Error())
+		s.log.Error("oc delete output:")
+		s.log.Error(string(output))
+		return err
+	}
+	s.log.Debug("Successfully deleted rolebinding %s, namespace ", handle, namespace)
```
eriknelson: Good catch, I will delete the resource file.

pkg/broker/broker.go
eriknelson: I wanted to call attention to it so that people are not surprised if they expect their APB binds to run and the broker is configured to *not* run them.

pkg/apb/client.go
rthallisey: The asbcli change will set context.Namspace so that local brokers don't need this hard coded.

etc/ex.mock.config.yaml
eriknelson: I'd prefer to add these in until #171 is discussed and merged at a later date. These aren't used now, but if for whatever reason somebody looks at it, they should reflect the current schema.

pkg/apb/svc_acct.go
rthallisey: OK I'll add it to the list of post TP changes.

rthallisey: @eriknelson Actually can you add it? http://etherpad.corp.redhat.com/post-tech-preview-changes

pkg/apb/client.go
eriknelson: So I should just remove 195 if context.Namespace is correct?

pkg/apb/svc_acct.go
rthallisey: Then should it be passed into the function?

etc/ex.mock.config.yaml
jwmatthews: Fine with adding the config sections in now if you want.  

Going forward the runtime config will be generated from the Template in a ConfigMap, so these config files only serve as reference points.

If we want multiple config examples for various situations or a single with comments, fine with either.

pkg/apb/svc_acct.go
eriknelson: @rthallisey Oh sure, didn't realize we had this out there.

pkg/broker/broker.go
jwmatthews: +1 to the warning

pkg/apb/svc_acct.go
eriknelson: That puts a burden on every caller that I would prefer to avoid. I could make it a local const in the file.

rthallisey: I'm not so worried about it being a const, but it's exposure.  Normally,  I would think a variable like this would end up somewhere in types because it's the default Role value for the service account.

eriknelson: I'll move to types

pkg/apb/client.go
jwmatthews: I assume we'll want to keep line 155 since we reuse the 'ns' alias in a few spots.

I'd like add an explicit check for context.Namespace being sent and non-empty.
If we don't get a namespace in context.Namespace let's log a statement explicitly noting this and error out.

pkg/apb/client.go
jwmatthews: Agree, removing 195 to get rid of the hard coded 'default' namespace would be good.

This tripped me up last night and caused some pain in understanding why my APBs were going to 'default'.

In general I'd like to see us remove the branch for 'local' runs, worried about it complicating debugging.  Doesn't have to be done in this PR

pkg/apb/client.go
eriknelson: @rthallisey @jwmatthews SGTM, will keep 155 since it's just an alias, check for missing namespace (log + error out if missing) and pull 195 ns := "default".

pkg/apb/client.go
fabianvf: branch for local runs is useful for testing with the broker when its not running inside a cluster, like for local development/testing. Ideally though, the branch would be minimal, it should only affect the the actual `oc login` call and definitely not impact namespace/etc.

pkg/apb/provision.go
```diff
@@ -34,28 +34,56 @@ func Provision(
...
-		return nil, err
+		return podName, nil, err
 	}
 
-	return extractCredentials(output, log)
+	creds, err := extractCredentials(podName, context.Namespace, log)
+	return podName, creds, err
+}
+
+func projectExists(project string) bool {
```
eriknelson: I don't think we should ignore the RunCommand error, and a RunCommand with exit code support is potentially useful to have.

rthallisey: It's not that I want to ignore the RunCommand error, but I think we can accomplish this same thing with one less oc command.  Up to this point we are under the assumption that we need to create a project.  This check determines whether we should create it or not. I'm saying, instead of checking this way, create the project and  check if the error returned = "this project already exists". That will be the same functionality as this.

pkg/apb/svc_acct.go
```diff
@@ -134,20 +136,27 @@ func (s *ServiceAccountManager) createResourceDir() {
...
-	rFilePath := filepath.Join(resourceDir(), apbId+".yaml")
+func (s *ServiceAccountManager) createFile(handle string) (string, error) {
+	rFilePath := filePathFromHandle(handle)
 	s.log.Debug("Creating resource file %s", rFilePath)
-	var _, err = os.Stat(rFilePath)
 
-	if os.IsNotExist(err) {
-		var file, err = os.Create(rFilePath)
+	_, err := os.Stat(rFilePath)
+	if err != nil && os.IsNotExist(err) {
```
rthallisey: This could be a single line.  ```if _, err := os.Stat(rFilePath); os.IsNotExist(err)```

pkg/apb/provision.go
eriknelson: @rthallisey are you saying we should parse the output of the oc command? Scraping stdout for error types is brittle.

pkg/apb/svc_acct.go
eriknelson: I will update, but this pattern is entirely inconsistent across the codebase. We should decide if this style is grounds to reject PRs going forwards or not, and stick to that.

pkg/apb/provision.go
rthallisey: I doubt it would change much from "doesn't exist", but agree stdout isn't ideal. Let's leave it as is.

PR #179, opened 2017-06-14
Change rpms to build using vendor dependencies

jmontleon:


jmrodri:
-10000 I've said I didn't want vendor in the main repo. 200MB of noise.

jmrodri:
My responses to the email thread:
> Can we create a different repository? 

> We only need the committed vendor for building rpms. Not for development. 
>
>So I don't see the need for git suvmodules either. Hell I don't think a normal user even needs to know about this other repo 

>+1 have a repo, fudor/broker-vendor for vendor built as an rpm: broker-vendor.rpm. Make that as a dependency of the asb rpm. This is how we handled this in candlepin. Keeps main repo clean, vendor gets updated when rpm built.

jmrodri:
Also, it was stated origin commits the vendor directory. My response is:

>Yes, origin does, so does kubernetes. I still don't like it.

tsanders-rh:
Understand the concerns.  The plan  and decision is to check the vendor directory in for Tech Preview.  We will address alternative options and tooling to lessen the impact on development afterwards.  ACK for merge.

jmrodri:
It will be much harder to reduce the size of the repo once this is in. The git index will always have it in the history.

PR #182, opened 2017-06-14
Stop running glide for every build and ignore compiled binaries

jmontleon:


Makefile
```diff
@@ -8,7 +8,7 @@ INSTALL_DIR      ?= /usr/local
 vendor:
```
rthallisey: Another thing I've also noticed is not having perms to run the binary in ```/usr/local```. Maybe we should just go with ```.``` or make a local dir?

eriknelson: When are you running it out of /usr/local? Not sure what this is in reference to.

jmontleon: Right now it's INSTALL_DIR. I can always change it to PREFIX. I was looking at --prefix but that would be a function of configure rather than make and I wasn't ready to go there.

Right now you can just run `make build` or `make install`. Running make build && make install will end up building twice as install does a build (as build did a vendor). I can undo that if it's undesirable. Then it's just necessary to do `make build && sudo make install` or something like `make build && INSTALL_DIR=~ make install`.

Currently you'd just want to do `sudo -E make install`


Per the notes in the Targets section:

```
  * Notes for install, run, and uninstall:
    * The default install prefix is /usr/local. Use `sudo -E make install` to obtain proper permissions and preserve env vars like GOPATH
    * Alternatively you can alter the installation directory by using INSTALL_DIR, e.g. `INSTALL_DIR=~ make run`
```

eriknelson: > make build && sudo make install or something like make build && INSTALL_DIR=~ make install

This is what I'd expect coming to a Makefile based project, I don't know if the go community does things differently. `make install` normally installs to a a system location to be overridden by PREFIX for c/c++ stuff.

eriknelson:
ACK pending Travis.

PR #188, opened 2017-06-16
Clean Registry URL prefix before attempting a request

dymurray:


jmontleon:
ACK. I think this is a good idea. When switching from dockerhub to rhcc it was unclear that I had to change the url from docker.io to http://registry.access.redhat.com. @dymurray explained why that is, and it makes sense, but I think this will save a lot of misery.

pkg/apb/rhcc_registry.go
```diff
@@ -37,6 +38,20 @@ func (r *RHCCRegistry) Init(config RegistryConfig, log *logging.Logger) error {
...
 }
 
+// This function is used because our code expects an HTTP Url for talking to RHCC
+func (r RHCCRegistry) cleanHttpUrl(url string) string {
+	if strings.HasPrefix(url, "http://") == true {
+		return url
+	}
+
+	if strings.HasPrefix(url, "https://") == true {
+		url = strings.TrimPrefix(url, "https://")
```
eriknelson: Do we have a specific reason for doing this? I think it's ill-advised to quietly swap https -> http if https was knowingly specified.

PR #190, opened 2017-06-16
Set default for LAUNCH_APB_ON_BIND to 'false' for TP

johnkim76:


jwmatthews:
ACK

I'd like to keep this as false for tech-preview.

PR #192, opened 2017-06-18
Local Development Changes:  Simulates broker running InCluster while executing locally

jwmatthews:
Assumes running oc cluster up on localhost.

Allows running broker local on host with:
  - service account credentials
  - etcd route
  - modifies asb route to point to local broker

With this PR we can do the following:

1) Use catasb 'dev' branch to bring up a cluster
2) make build
3) cd scripts && ./run_local.sh

The local broker now replaces the broker the Service Catalog was talking to.

I've had success with this doing provision/bind with mediawiki & rhscl postgres example.

Note the broker will not have access to the service network, for example it couldn't reach the running etcd so we created a route to expose it.  We are not talking to the service network for anything else currently, in future this may be a limitation for us.

Also...I've tested this with building broker on macOS and running locally on Mac talking to cluster.

### DIFF
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -1,0 +1,28 @@
+# Recommended Local Development and Debugging Workflow
+
+## Overview
+
+We recommend 2 modes of running the broker for development work.
+
+1. Build the broker executable and run it locally connected to a local oc cluster up from [fusor/catasb 'dev' branch](https://github.com/fusor/catasb/tree/dev)
+2. Build a local image and deploy from image
+
+## Running the broker executable locally
+
+1. Build the Ansible Service Broker executable with the command: ```make build```
+2. Ensure you have a local oc cluster up environment running with [fusor/catasb 'dev' branch](https://github.com/fusor/catasb/tree/dev)
+3. ```cp scripts/my_local_dev_vars.example scripts/my_local_dev_vars```
+4. Edit 'scripts/my_local_dev_vars' 
+5. Prepare your local development environment by running: ```make prepare-local-env```
+    * This will remove the running 'asb' pod and replace the endpoint of the asb route with the locally running broker executable.
+    * You __must__ rerun this whenever you have reset your cluster.
+    * It is safe to run this multiple times
+6. Run the broker executable locally:  ```make run```
+    * Use the cluster as you normally would for testing.
+    * When you want to make a change and rebuild the broker.  CTRL-C, rebuild and re-run
+
+## Build a local image and deploy from image
+
+1. Build the Ansible Service Broker executable with the command: ```make build ```
+3. Build a development image:  ```make build-image BROKER_IMAGE_NAME=asb-dev TAG=local```
+4. Deploy from [templates/deploy-ansible-service-broker.template.yaml](https://github.com/openshift/ansible-service-broker/blob/master/templates/deploy-ansible-service-broker.template.yaml) setting the parameter "-p BROKER_IMAGE=asb-dev:local" to your local image name/tag 
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -3,7 +3,6 @@ package broker
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"reflect"
 
 	"github.com/coreos/etcd/client"
@@ -12,6 +11,7 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 	"github.com/openshift/ansible-service-broker/pkg/dao"
 	"github.com/pborman/uuid"
+	k8srestclient "k8s.io/client-go/rest"
 )
 
 var (
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
 
 }
 func (a AnsibleBroker) Login() error {
-	a.log.Debug("Retrieving serviceaccount token")
-	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	clientConfig, err := k8srestclient.InClusterConfig()
 	if err != nil {
-		a.log.Debug("Error reading serviceaccount token")
+		a.log.Error("Failed to create a restclient.InClusterConfig: %v.", err)
 		return err
 	}
 
-	return apb.OcLogin(a.log, "https://kubernetes.default",
-		"--certificate-authority", "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
-		"--token", string(token),
+	return apb.OcLogin(a.log, clientConfig.Host,
+		"--certificate-authority", clientConfig.CAFile,
+		"--token", clientConfig.BearerToken,
 	)
 }
 
--- a/scripts/my_local_dev_vars.example
+++ b/scripts/my_local_dev_vars.example
@@ -1,0 +1,20 @@
+#####
+## CHANGE THESE TO REFLECT YOUR ENVIRONMENT
+#####
+
+# Typically the below values are:
+#  For Linux, use OPENSHIFT_SERVER_HOST=172.17.0.1
+#  For macOS, use OPENSHIFT_SERVER_HOST=192.168.37.1
+OPENSHIFT_SERVER_HOST=172.17.0.1
+OPENSHIFT_SERVER_PORT=8443
+
+# BROKER_IP_ADDR must be the IP address of where to reach broker
+#   it should not be 127.0.0.1, needs to be an address the pods will be able to reach
+BROKER_IP_ADDR=${OPENSHIFT_SERVER_HOST}
+DOCKERHUB_USERNAME=""
+DOCKERHUB_PASSWORD=""
+DOCKERHUB_ORG="ansibleplaybookbundle
+
+
+BROKER_CMD="../broker"
+GENERATED_BROKER_CONFIG="../etc/generated_local_development.yaml"
--- a/scripts/run_local.sh
+++ b/scripts/run_local.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+MY_VARS="./my_local_dev_vars"
+if [ ! -f $MY_VARS ]; then 
+  echo "Please create $MY_VARS"
+  echo "cp $MY_VARS.example $MY_VARS"
+  echo "then edit as needed"
+  exit 1
+fi 
+
+source ./${MY_VARS}
+if [ "$?" -ne "0" ]; then
+  echo "Error reading in ${MY_VARS}"
+  exit 1
+fi 
+
+if [ -z "${BROKER_CMD}" ]; then 
+  echo "Please ensure BROKER_CMD is defined in ${MY_VARS}"
+  exit 1 
+fi 
+
+export KUBERNETES_SERVICE_HOST=${OPENSHIFT_SERVER_HOST}
+export KUBERNETES_SERVICE_PORT=${OPENSHIFT_SERVER_PORT}
+
+BROKER_CONFIG=${GENERATED_BROKER_CONFIG}
+if [ ! -z "$1" ]; then
+  BROKER_CONFIG="$1"
+fi 
+
+if [ -z "${BROKER_CONFIG}" ]; then 
+  echo "Please specify a broker configuration file to run"
+  exit 1
+fi 
+
+echo "Running ${BROKER_CMD} --config ${BROKER_CONFIG}"
+${BROKER_CMD} --config ${BROKER_CONFIG}
--- a/scripts/test_local_server_connection.sh
+++ b/scripts/test_local_server_connection.sh
@@ -1,0 +1,6 @@
+CERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+TOKEN=`cat /var/run/secrets/kubernetes.io/serviceaccount/token`
+export KUBERNETES_SERVICE_HOST=192.168.37.1
+export KUBERNETES_SERVICE_PORT=8443
+
+curl --cacert ${CERT} -H "Authorization: Bearer ${TOKEN}" https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/version
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
 REGISTRY         ?= docker.io
 PROJECT          ?= ansibleplaybookbundle
 TAG              ?= latest
-BROKER_IMAGE     = $(REGISTRY)/$(PROJECT)/ansible-service-broker
+BROKER_IMAGE     ?= $(REGISTRY)/$(PROJECT)/ansible-service-broker
 BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/build"
 PREFIX           ?= /usr/local
+BROKER_CONFIG    ?= $(PWD)/etc/generated_local_development.yaml
 
 vendor:
 	@glide install -v
@@ -17,8 +18,8 @@ install:
 	mkdir -p ${PREFIX}/etc/ansible-service-broker
 	cp etc/example-broker-config.yaml ${PREFIX}/etc/ansible-service-broker/broker-config.yaml
 
-run:
-	${PREFIX}/bin/ansible-service-broker --config ${PREFIX}/etc/ansible-service-broker/broker-config.yaml
+run: 
+	cd scripts && ./run_local.sh ${BROKER_CONFIG}
 
 uninstall:
 	rm  -f ${PREFIX}/bin/ansible-service-broker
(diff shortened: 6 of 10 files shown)


docs/local_development.md
```diff
@@ -0,0 +1,27 @@
...
+
+We recommend 2 modes of running the broker for development work.
+
+1. Build a local image and deploy from image
+2. Build the broker executable and run it locally connected to a local oc cluster up from [fusor/catasb 'dev' branch](https://github.com/fusor/catasb/tree/dev)
+
+## Build a local image and deploy from image
+
+1. Build the Ansible Service Broker executable with the command: ```make build```
+2. Edit the Makefile to reflect the BROKER_IMAGE_NAME and TAG you want to use for local development
```
jmrodri: Do not require editing the Makefile. If we need to change the image name and tag, then those should be passed in as parameters.

docs/local_development.md
```diff
@@ -0,0 +1,27 @@
...
+3. Build a development image:  ```make build-image```
```
jmrodri: merge step 2 and 3. `make build-image BROKER_IMAGE_NAME=asb-dev TAG=johnny5`. Editing the `Makefile` is a bad idea.

docs/local_development.md
```diff
@@ -0,0 +1,27 @@
...
+4. Deploy from [templates/deploy-ansible-service-broker.template.yaml](https://github.com/openshift/ansible-service-broker/blob/master/templates/deploy-ansible-service-broker.template.yaml) setting the parameter "-p BROKER_IMAGE" to your local image name/tag 
+
+## Running the broker executable locally
+
+1. Build the Ansible Service Broker executable with the command: ```make build```
+2. Ensure you have a local oc cluster up environment running with [fusor/catasb 'dev' branch](https://github.com/fusor/catasb/tree/dev)
```
jmrodri: do I need to run catasb? I used to just run `oc cluster up` when I ran it locally.

docs/local_development.md
```diff
@@ -0,0 +1,27 @@
...
+3. ```cd scripts```
+4. ```cp my_local_dev_vars.example my_local_dev_vars```
+5. Edit 'my_local_dev_vars' 
+6. Run: ```./run_local.sh```
+  * This will remove the running 'asb' pod and replace the endpoint of the asb route with the locally running broker executable. 
+7. Use the cluster as you normally would for testing.
+  * When you want to make a change and rebuild the broker.  CTRL-C ```run_local.sh```, rebuild and re-run ```run_local.sh```
```
jmrodri: I really hate the need for the scripts. I would prefer `make run` handle all of this.

pkg/broker/broker.go
```diff
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
 
 }
 func (a AnsibleBroker) Login() error {
-	a.log.Debug("Retrieving serviceaccount token")
-	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
```
jmrodri: hrm, I was totally going to hack this to have the token file locally on my system. Using the `InClusterConfig` will probably negate that.

pkg/broker/broker.go
```diff
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
...
+	clientConfig, err := k8srestclient.InClusterConfig()
 	if err != nil {
-		a.log.Debug("Error reading serviceaccount token")
+		a.log.Warning("Failed to create a restclient.InClusterConfig: %v.", err)
```
jmrodri: :+1: to the warning.

pkg/broker/broker.go
```diff
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
...
 		return err
 	}
 
-	return apb.OcLogin(a.log, "https://kubernetes.default",
-		"--certificate-authority", "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
-		"--token", string(token),
+	return apb.OcLogin(a.log, clientConfig.Host,
+		"--certificate-authority", clientConfig.CAFile,
+		"--token", clientConfig.BearerToken,
```
jmrodri: Ok I think I see what happened. Instead of using hardcoded values, simply use a configuration setting. :ok:

scripts/test_local_server_connection.sh
```diff
@@ -0,0 +1,6 @@
+CERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+TOKEN=`cat /var/run/secrets/kubernetes.io/serviceaccount/token`
+export KUBERNETES_SERVICE_HOST=192.168.37.1
+export KUBERNETES_SERVICE_PORT=8443
+
+curl --cacert ${CERT} -H "Authorization: Bearer ${TOKEN}" https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/version
```
jmrodri: Why can't i put these tokens on my file system and run the broker outside the cluster?

scripts/run_local.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
+###
```
jmrodri: How long does it take to go from change to build to redeploy?

docs/local_development.md
jwmatthews: The only requirement is a working cluster that is running locally.

pkg/broker/broker.go
jwmatthews: Yes, this is using the same mechanism we use for creating the client to talk to k8s API, namely using the environment variables KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT.

scripts/test_local_server_connection.sh
jwmatthews: You can and that is what I am doing.
The whole point of the script is to fetch the needed info and store it locally.

scripts/run_local.sh
jwmatthews: ~15 seconds maybe?

jmrodri:
Script won't work with plain `oc cluster up`. There are too many things missing. i.e. no `ansible-service-broker` namespace, no `asb` service account, etc. Those are created by catasb.

jmrodri:
catasb/local deploys a broker image, then this script deletes it. I would prefer we skipped that step somehow. I typically destroy my cluster `oc cluster down` and if I'm using catasb, having to wait for it to deploy something we are going to delete will eat up time.

jwmatthews:
"Script won't work with plain oc cluster up. There are too many things missing. i.e. no ansible-service-broker namespace, no asb service account, etc. Those are created by catasb."

To be clear, the resources you refer to are created from our openshift template, it's not just a catasb thing:
https://github.com/openshift/ansible-service-broker/blob/master/templates/deploy-ansible-service-broker.template.yaml

The openshift template is how the broker is deployed to a cluster.

This run_local.sh script assumes the template was used and then it replaces the broker with the local version.  It's intended to run to complement catasb, yet technically it's requirement is the openshift template was used.

Testing the broker with a plan oc cluster up implies the below:

oc cluster up 
oc new-project ansible-service-broker
oc process deploy-ansible-service-broker.template.yaml -n ansible-service-broker | oc create -f -

scripts/run_local.sh
```diff
@@ -0,0 +1,170 @@
...
+echo "Broker Service Account Token is in secret: ${BROKER_SVC_ACCT_SECRET_NAME}"
+
+###
+# Fetch the service-ca.crt for the service account
+###
+SVC_ACCT_CA_CRT_DATA=`oc get secret ${BROKER_SVC_ACCT_SECRET_NAME} -n ${ASB_PROJECT} -o json | jq -c '.data["service-ca.crt"]'`
+# Remove quotes from variable
+SVC_ACCT_CA_CRT_DATA=( $(eval echo ${SVC_ACCT_CA_CRT_DATA[@]}) )
+# Base64 Decode
+SVC_ACCT_CA_CRT_DATA=`echo ${SVC_ACCT_CA_CRT_DATA} | base64 -D `
```
shawn-hurley: ```echo ${SVC_ACCT_CA_CRT_DATA} | base64 --decode ``` will work on both Linux and Darwin.  It appears that linux wants -d and darwin wants -D

jwmatthews: good catch will get that updated

jwmatthews: Pushed change for --decode

eriknelson:
The workflow I'd like to get to is basically:

```
-> Bring up cluster with catasb/local/run_environment.sh once to get myself a cluster
-> Configure local config values in scripts/my_local_dev_vars_once
-> make run
-> loop(dev some changes, make run, test changes) Needs to be as fast as we can get
```

My suggested changes:
* Add dockerhub org the broker will discover apbs from to the configurable options to my_local_dev_vars, because I'm often testing my own apbs out of my org
* Put run_local.sh behind make run so I can repeatedly do make run, Ctl+C, make some changes, repeat
* Put the logic in run_local.sh behind some kind of check to see if that's already been done. In theory, it will only tear down the ansible-service-broker that was brought up by catasb once, and then every subsequent run will skip this. Alternatively, pull it out of the run_local.sh script entirely and add a flag in catasb to skip creating these resources in the first place.

I don't really understand the objection to using the cluster's etcd, seems arbitrary to me. Is it because it's easier to query a local etcd for debugging purposes?

Apart from that, this is really close. I've been hacking around my development environment since we started, so I'm all in favor of finally having a standardized dev env.

jmrodri:
Using local catasb (after a reboot, and deleting all my old docker images) I was able to get catasb running locally. And run_local.sh deployed broker.

jmrodri:
Using oc cluster up does NOT work. There are too many things missing that catasb sets up and assumed to be there by run_local.sh.

eriknelson:
Is it correct that the broker needs to be deployed once ahead of time to generate the cert/token so that can be extracted locally?

scripts/run_local.sh
```diff
@@ -0,0 +1,170 @@
...
+  color: true
+openshift: {}
+broker:
+  dev_broker: true
+  launch_apb_on_bind: false
+  recovery: true
+  output_request: true
+EOF
+
+${BROKER_CMD} -c ${GENERATED_BROKER_CONFIG}
```
jmrodri: I would rather not run the broker in the script. I would favor the script do all of the setup. Then let me run the broker by hand OR via `make run`. That way it is obvious I don't need to re-run the script.

jmrodri: Or better yet, have a setup task that runs the script from the Makefile. Then make run launches the broker:
```
setup:
    scripts/run_local.sh
run: setup
    ./broker -c $CONFIG
```

eriknelson: Separating the targets is not a bad idea; setup would basically reconfigure everything for a true local setup, and then run literally just runs a binary. changes, make run, Ctl+C, changes, make run loop is about as short as it can get, it's limited to binary build times.

jmrodri:
I started a branch for make run, but it makes sense to be in your PR:

```
BROKER_CONFIG    ?= etc/generated_local_development.yaml
```

```
run:
    broker --config $(BROKER_CONFIG)
```

pkg/broker/broker.go
```diff
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
...
-		a.log.Debug("Error reading serviceaccount token")
+		a.log.Warning("Failed to create a restclient.InClusterConfig: %v.", err)
 		return err
 	}
 
-	return apb.OcLogin(a.log, "https://kubernetes.default",
-		"--certificate-authority", "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
-		"--token", string(token),
+	return apb.OcLogin(a.log, clientConfig.Host,
+		"--certificate-authority", clientConfig.CAFile,
```
rthallisey: I'm -1 to merge this part.  OcLogin is only used by a local broker so we can't use an incluster config object.  In addition, I think we should remove RefreshLoginToken, which is the function that calls this. IMO we shouldn't change anything here, but follow up with a patch to remove RefreshLoginToken and OcLogin for the TP. I'll come up with a patch for this shortly.

rthallisey: Nevermind this gets used in two places. 1) When we create a broker, if we're incluster use the incluster config 2) refresh login.  If we Remove 2), then I think we can use it the way you propose here.

jwmatthews: @rthallisey please look again, this OcLogin is actually opposite of what you said:

https://github.com/jwmatthews/ansible-service-broker/blob/34f23d72cef22bef8dbf221df41b05a5b6c838b2/pkg/broker/broker.go#L80

OcLogin for this case is used when running with InCluster.
It is intended to get a ~/.kube/config so that the oc create commands for role bindings and the like for dynamic service account creation will function.

This call is __not__ called from RefreshLoginToken, this is called from NewAnsibleBroker()

jwmatthews: If this code path was used for #2 of refresh token, then that was broken.
Running locally wouldn't have been able to resolve the hardcoded kubernetes.default DNS

jwmatthews: The other refresh login for the local case was calling apb/client.go:

https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/client.go#L247

rthallisey: I have the names a little mixed up.  I thought this was OcLogin.

pkg/broker/broker.go
```diff
@@ -101,16 +101,15 @@ func (a AnsibleBroker) getServiceInstance(instanceUUID uuid.UUID) (*apb.ServiceI
 
 }
 func (a AnsibleBroker) Login() error {
-	a.log.Debug("Retrieving serviceaccount token")
-	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	clientConfig, err := k8srestclient.InClusterConfig()
```
rthallisey: To get the client use ```a.clusterConfig```

jwmatthews: I don't think that will work.
I'm assuming a.clusterConfig is an instance of apb.ClusterConfig.

apb.ClusterConfig doesn't contain the variables we need.
https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/client.go#L45


I was looking for a way to reference this:
https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/client.go#L82

rthallisey: This looks like a larger cleanup patch is needed.  To get the client you would call ```apb.Client.ClusterClient```.  The problem is the client is created during a provision, bind, unbind, or deprovision.  So we either stay with what you have and make note of the tech debt (I did already in the post TP etherpad) or create the clients before the broker.

https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/provision.go#L61

jwmatthews: Thanks Ryan for helping to research into this.

I like the idea of sticking with the code as-is for now and revisit later as per your etherpad note.

scripts/my_local_dev_vars.example
```diff
@@ -0,0 +1,20 @@
...
+OPENSHIFT_SERVER_HOST=172.17.0.1
+OPENSHIFT_SERVER_PORT=8443
+
+# BROKER_IP_ADDR must be the IP address of where to reach broker
+#   it should not be 127.0.0.1, needs to be an address the pods will be able to reach
+BROKER_IP_ADDR=${OPENSHIFT_SERVER_HOST}
+DOCKERHUB_USERNAME=""
+DOCKERHUB_PASSWORD=""
+DOCKERHUB_ORG="ansibleplaybookbundle
+
```
rthallisey: I think we may be able to merge this with broker-config.yaml.  We can use ```function parse_yaml``` in ```scripts/deploy.sh``` to take any variables and pull them into bash. Your script ```prep_local_devel_env.sh``` can make the call to parse the yaml. Again, this could be tech debt after we cut the release.

jwmatthews: Let's hold off on that change.

jwmatthews:
Updated the workflow as per our meeting today, broke out the targets in Makefile to:
  prepare-local-env
  run

Ready to merge if others are good with this.

scripts/run_local.sh
```diff
@@ -0,0 +1,35 @@
...
+  BROKER_CONFIG="$1"
+fi 
+
+if [ -z "${BROKER_CONFIG}" ]; then 
+  echo "Please specify a broker configuration file to run"
+  exit 1
+fi 
+
+echo "Running ${BROKER_CMD} --config ${BROKER_CONFIG}"
+${BROKER_CMD} --config ${BROKER_CONFIG}
```
jmrodri: I would put the my vars in `etc/` and write out the generated config to `etc/generated_...`

jmrodri:
nevermind, it's close enough. ACK

shawn-hurley:
ACK

PR #195, opened 2017-06-19
Check for empty spec dir when querying for catalog.

cfchase:


jmrodri:
run:

```
gofmt -d ./cmd
gofmt -d ./pkg
```

pkg/dao/dao.go
```diff
@@ -126,7 +126,12 @@ func (d *Dao) BatchGetRaw(dir string) (*[]string, error) {
 	var err error
 
 	opts := &client.GetOptions{Recursive: true}
-	if res, err = d.kapi.Get(context.Background(), dir, opts); err != nil {
+	res, err = d.kapi.Get(context.Background(), dir, opts)
+
+	if client.IsKeyNotFound(err) {
+		emptySpec := make([]string, 0)
+		return  &emptySpec, nil
+	} else if err != nil {
```
jmrodri: I would rather do the check in the specific `BatchGet`s. Because `BatchGetRaw` is intended to be used by more than just specs.

Change `BatchGetSpecs` to do the error checking.

```golang
// BatchGetSpecs - Retrieve all the specs for dir.
func (d *Dao) BatchGetSpecs(dir string) ([]*apb.Spec, error) {
    payloads, err := d.BatchGetRaw(dir)
        // handling key error
        return []*apb.Spec{}, nil
    } else if err != nil {
        return []*apb.Spec{}, err
    }

    specs := make([]*apb.Spec, len(*payloads))
    for i, payload := range *payloads {
        spec := &apb.Spec{}
        apb.LoadJSON(payload, spec)
        specs[i] = spec
        d.log.Debug("Batch idx [ %d ] -> [ %s ]", i, spec.Id)
    }

    return specs, nil
}
```

jmrodri:
You might want to configure the IDE to run `gofmt` or `goimports` on the files to avoid failing travis.

cfchase:
@jmrodri made the requested changes

PR #197, opened 2017-06-19
Remove refresh login function

### LINKED ISSUE #196: Breakdown the different code paths between a local and incluster broker
Right now there is too much variation between the local and incluster code paths.  We want to use nearly identical code paths so that when we develop the broker locally, it will reflect how it will run incluster.

|          | Config files | RefreshLoginToken  | APB environment vars |
| ------------- |-------------|:-----:|:------:|
| **Local**     |  ~/.kube/config | Code exists, but not needed | Code exists, but not needed |
| **InCluster**     |  /var/run/serviceaccounts/... | not used  | not used |

rthallisey:
The RefreshLoginToken function is only used with a local
broker. Let's remove this extra code path and have developers
refresh their login token manually when necessary.

Partially-fixes #196

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -125,17 +125,6 @@ func (c *Client) RunImage(
 		return "", err
 	}
 
-	if !clusterConfig.InCluster {
-		err = c.RefreshLoginToken(clusterConfig)
-
-		if err != nil {
-			c.log.Error("Error occurred while refreshing login token! Aborting apb run.")
-			c.log.Error(err.Error())
-			return "", err
-		}
-		c.log.Notice("Login token successfully refreshed.")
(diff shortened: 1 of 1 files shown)


rthallisey:
Do we want this post TP?

eriknelson:
ACK. I'm a proponent of removing code that expects a user/pass combo. It's only been a source of trouble since we moved to ServiceAccounts. If we want to support user impersonation, we should think about the design for that and implement accordingly.

rthallisey:
[Remove RefreshLoginToken](https://trello.com/c/fhSFBRN6/294-remove-refreshlogintoken)

eriknelson:
2 ACKs on this, merging.

PR #199, opened 2017-06-20
162: show errors during failed format runs

### LINKED ISSUE #162: fix travis.sh lint to output errors on failure
The `scripts/travis.sh` script will eat linter errors. We should probably output the errors so that we can see them in the logs when it fails. Also print a suggestion to the user that they should run:

```
gofmt -d ./cmd
gofmt -d ./pkg
```

jmrodri:
* add vet and golint. Vet will fail build.

Fixes #162.

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -49,7 +49,7 @@ elif [[ "$action" == "format" ]]; then
   echo "================================="
   echo "             Format              "
   echo "================================="
-  if [ -n "$(gofmt -l cmd pkg)" ]; then
+  if [ -n "$(gofmt -l $PROJECT_ROOT/cmd $PROJECT_ROOT/pkg)" ]; then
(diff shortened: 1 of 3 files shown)


scripts/travis.sh
```diff
@@ -38,16 +38,27 @@ if [[ "$action" == "install" ]]; then
...
+  # to have lint fail build add -set_exit_status option
+  # to each golint command
+  golint $PROJECT_ROOT/cmd/...
+  golint $PROJECT_ROOT/pkg/...
+  exit $?
+elif [[ "$action" == "format" ]]; then
+  echo "================================="
+  echo "             Format              "
+  echo "================================="
+  if [ -n "$(gofmt -l cmd pkg)" ]; then
```
shawn-hurley: This returns values example
```stat cmd: no such file or directory```
```stat pkg: no such file or directory```

any reason not to always just use $PROJECT_ROOT ?

PR #200, opened 2017-06-20
removing unnecessary function that just add's indirection.

shawn-hurley:
Really want to open a discussion on this one. 

I think that it does not make sense to wrap package calls like this unless you are doing more logic for every time you call that method. 

If we do want to continue, then we should update the function 
```
func LoadYAML(yaml []byte, obj interface{}) {
        return yaml.Unmarshal(yaml, obj)
}
```


eriknelson:
@jwmatthews need another call here, TP1 or post TP branch? No objection to this in TP1 either.

PR #208, opened 2017-06-21
Cleanup local pod split

### LINKED ISSUE #196: Breakdown the different code paths between a local and incluster broker

rthallisey:
We don't need to pass in environment vars anymore
because of service accounts. We're always running the apb
incluster so oc login will use the service account to login.

Partially-fixes #196

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -166,66 +166,26 @@ func (c *Client) RunImage(
 		return apbId, err
 	}
 
-	// TODO(rhallisey): Cleanup the local vs incluster code paths so there
-	// is almost no variation between the two.
-	// TODO(rhallisey): The Kubernetes API objects should be in JSON structs
-	// so it's easier to manipulate them. This is for a rebase so we don't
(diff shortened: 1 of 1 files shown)


rthallisey:
[Remove the incluster pod creation path](https://trello.com/c/fCZ7fyah/295-remove-the-incluster-pod-creation-path)

rthallisey:
Depends-on: https://github.com/fusor/apb-examples/pull/58

eriknelson:
Merged https://github.com/fusor/apb-examples/pull/58, 2 ACKs on this, merging.

Issue #59, opened 2017-03-31
Refresh the catalog when bootstrapping the broker


rthallisey:
If you leave etcd running it will build up a catalog of services from every bootstrap that has been run in the past. Since it's likely we will be provided etcd in an environment, we don't want bootstrap to only pull from what's currently in etcd, but to update etcd to the latest set of images in the docker registry.

shawn-hurley:
Closed by PR #221

PR #209, opened 2017-06-21
Dev remove spec

shawn-hurley:
This PR is needed by https://github.com/fusor/ansible-playbook-bundle/pull/58

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -46,6 +46,12 @@ type BrokerConfig struct {
 	OutputRequest   bool `yaml:"output_request"`
 }
 
+type DevBroker interface {
+	AddSpec(spec apb.Spec) (*CatalogResponse, error)
+	RemoveSpec(specID string) error
+	RemoveSpecs() error
+}
+
 // AnsibleBroker - Broker using ansible and images to interact with oc/kubernetes/etcd
 type AnsibleBroker struct {
 	dao           *dao.Dao
@@ -730,4 +736,36 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
 	}
 	service := SpecToService(&spec)
 	return &CatalogResponse{Services: []Service{service}}, nil
+}
+
+func (a AnsibleBroker) RemoveSpec(specID string) error {
+	spec, err := a.dao.GetSpec(specID)
+	if client.IsKeyNotFound(err) {
+		return ErrorNotFound
+	}
+	if err != nil {
+		a.log.Error("Something went real bad trying to retrieve spec for deletion... - %v", err)
+		return err
+	}
+	err = a.dao.DeleteSpec(spec.Id)
+	if err != nil {
+		a.log.Error("Something went real bad trying to delete spec... - %v", err)
+		return err
+	}
+	return nil
+}
+
+func (a AnsibleBroker) RemoveSpecs() error {
+	dir := "/spec"
+	specs, err := a.dao.BatchGetSpecs(dir)
+	if err != nil {
+		a.log.Error("Something went real bad trying to retrieve batch specs for deletion... - %v", err)
+		return err
+	}
+	err = a.dao.BatchDeleteSpecs(specs)
+	if err != nil {
+		a.log.Error("Something went real bad trying to delete batch specs... - %v", err)
+		return err
+	}
+	return nil
 }
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -69,6 +69,8 @@ func NewHandler(b broker.Broker, log *logging.Logger, brokerConfig broker.Broker
 
 	if h.brokerConfig.DevBroker {
 		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbAddSpec)).Methods("POST")
+		h.router.HandleFunc("/apb/spec/{spec_id}", createVarHandler(h.apbRemoveSpec)).Methods("DELETE")
+		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbRemoveSpecs)).Methods("DELETE")
 	}
 
 	return handlers.LoggingHandler(os.Stdout, h)
@@ -306,7 +308,7 @@ func (h handler) apbAddSpec(w http.ResponseWriter, r *http.Request, params map[s
 	//Read Request for an image name
 
 	// create helper method from MockRegistry
-	ansibleBroker, ok := h.broker.(*broker.AnsibleBroker)
+	ansibleBroker, ok := h.broker.(broker.DevBroker)
 	if !ok {
 		h.log.Errorf("unable to use broker - %T as ansible service broker", h.broker)
 		writeResponse(w, http.StatusInternalServerError, broker.ErrorResponse{Description: "Internal server error"})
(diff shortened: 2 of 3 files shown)


rthallisey:
@shawn-hurley is https://github.com/openshift/ansible-service-broker/issues/59 related?

shawn-hurley:
@rthallisey it is and it is not. This is only adding the ability for brokers in development mode to remove specs from the broker. I have a different branch with a different commit that will fix that issue as well as the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1463798). I will be using the methods from this commit though, so want to wait till this is merged in and then submit the PR for the Bugzilla

rthallisey:
@shawn-hurley ok sounds good

eriknelson:
As I understood today, we're looking to accept this for TP1. @jwmatthews, can you ACK that?

pkg/broker/broker.go
```diff
@@ -718,3 +724,35 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
...
 	return &CatalogResponse{Services: []Service{service}}, nil
 }
+
+func (a AnsibleBroker) RemoveSpec(specID string) error {
+	spec, err := a.dao.GetSpec(specID)
+	if client.IsKeyNotFound(err) {
+		return ErrorNotFound
+	}
+	if err != nil {
+		a.log.Error("Something went real bad trying to retrieve spec for deletion... - %v", err)
```
eriknelson: I still regret my shitty error messages :)

pkg/broker/broker.go
```diff
@@ -46,6 +46,12 @@ type BrokerConfig struct {
 	OutputRequest   bool `yaml:"output_request"`
 }
 
+type DevBroker interface {
```
eriknelson: +1, nicely done.

pkg/dao/dao.go
```diff
@@ -191,6 +197,18 @@ func (d *Dao) BatchGetSpecs(dir string) ([]*apb.Spec, error) {
 	return specs, nil
 }
 
+// BatchDeleteSpecs - set specs based on SpecManifest in the kvp API.
+func (d *Dao) BatchDeleteSpecs(specs []*apb.Spec) error {
+	// TODO: Is there no batch insert in the etcd api?
```
eriknelson: I remember searching for this and coming up with nothing. Would love to hear anything to the contrary.

pkg/handler/handler.go
```diff
@@ -335,6 +337,37 @@ func (h handler) apbAddSpec(w http.ResponseWriter, r *http.Request, params map[s
 	writeDefaultResponse(w, http.StatusOK, resp, err)
 }
 
+func (h handler) apbRemoveSpec(w http.ResponseWriter, r *http.Request, params map[string]string) {
+	ansibleBroker, ok := h.broker.(broker.DevBroker)
+	if !ok {
+		h.log.Errorf("unable to use broker - %T as ansible service broker", h.broker)
```
eriknelson: I don't think `Errorf` is actually required. I originally thought I had to `fmt.Sprintf`, and realized they just accept format strings. `Error("thing %s", "fubar")`, unless I'm off my rocker.

We should just try to remain consistent.

eriknelson: It's actually not clear to me why there is a *f variant.

https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/svc_acct.go#L127

Is an example of the non *f variant still accepting a format string. The docs literally use the same docs string: https://godoc.org/github.com/op/go-logging#Logger.Error

Anyone else have an opinion here?

pkg/broker/broker.go
jmrodri: :+1:

pkg/broker/broker.go
```diff
@@ -718,3 +724,35 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
 	service := SpecToService(&spec)
 	return &CatalogResponse{Services: []Service{service}}, nil
 }
+
+func (a AnsibleBroker) RemoveSpec(specID string) error {
+	spec, err := a.dao.GetSpec(specID)
+	if client.IsKeyNotFound(err) {
+		return ErrorNotFound
+	}
```
jmrodri: Is this really an error case? Removing a non existent spec seems like a noop instead of an error to me.

pkg/broker/broker.go
```diff
@@ -718,3 +724,35 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
...
+	return nil
+}
+
+func (a AnsibleBroker) RemoveSpecs() error {
+	dir := "/spec"
+	specs, err := a.dao.BatchGetSpecs(dir)
+	if err != nil {
+		a.log.Error("Something went real bad trying to retrieve batch specs for deletion... - %v", err)
+		return err
+	}
```
jmrodri: This make sense

pkg/handler/handler.go
```diff
@@ -69,6 +69,8 @@ func NewHandler(b broker.Broker, log *logging.Logger, brokerConfig broker.Broker
 
 	if h.brokerConfig.DevBroker {
 		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbAddSpec)).Methods("POST")
+		h.router.HandleFunc("/apb/spec/{spec_id}", createVarHandler(h.apbRemoveSpec)).Methods("DELETE")
+		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbRemoveSpecs)).Methods("DELETE")
 	}
```
jmrodri: :+1:

pkg/handler/handler_test.go
```diff
@@ -95,6 +111,25 @@ func TestNewHandler(t *testing.T) {
...
+	}
+	form := url.Values{}
+	form.Add("apbSpec", base64TestSpec)
+	req.PostForm = form
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	w := httptest.NewRecorder()
+	testhandler.ServeHTTP(w, req)
+	ft.AssertEqual(t, w.Result().StatusCode, http.StatusNotFound, fmt.Sprintf("resulting status was not 404 - %v", w.Result().Status))
+	ft.AssertNotNil(t, testhandler, "handler wasn't created")
+}
```
jmrodri: :tada: :trophy: :clap: :fireworks: :champagne: adding a test

pkg/handler/handler_test.go
```diff
@@ -261,23 +352,22 @@ func TestMissingOperation(t *testing.T) {
...
 	testhandler.lastoperation(w, r, params)
 	ft.AssertEqual(t, w.Code, 200, "lastoperation should've returned 200")
 	ft.AssertState(t, w.Body, "in progress")
 }
 
 // utility functions
 
-func BuildBootstrapHandler(err error) (
-	handler, *httptest.ResponseRecorder, *http.Request) {
+func buildBootstrapHandler(err error) (handler, *httptest.ResponseRecorder, *http.Request) {
```
jmrodri: the change to `build*` makes sense. Good catch.

pkg/handler/handler.go
shawn-hurley: I have opinions. The f variant is what is correct according to the docs IMO
https://github.com/op/go-logging/blob/master/logger.go#L253

pkg/handler/handler_test.go
shawn-hurley: :+1:  to golint :)

pkg/handler/handler.go
eriknelson: @shawn-hurley good enough for me. I like explicit. @jmrodri any objection? We should require this going forwards and consider any code that doesn't conform to be out of date.

pkg/broker/broker.go
shawn-hurley: I was thinking that alerting the user to a spec not being found from the tool, could give the users valuable feedback. But I do see where this could be considered a noop as well. Does anyone have super strong opinions?

eriknelson: If someone explicitly requests the removal of a spec and it doesn't exit, I think that's totally a valid error case.

"Please remove spec X", "We have no knowledge at all of spec X, bad request"

jmrodri: Ok that's fair.

eriknelson:
Got a bunch of acks on it and we've now cut the 3.6 release branch. Merging.

PR #210, opened 2017-06-21
strip makefile whitespace

eriknelson:


eriknelson:
Self-merging. It's trivial.

PR #211, opened 2017-06-21
Expect a config file to be mounted 

fabianvf:
rather than create one with sed based on environment variables


jwmatthews:
We need this for tech-preview.

jwmatthews:
ACK

jmrodri:
this breaks make build-image, and make deploy. It also breaks the rpm build since we renamed build-config.xml

Issue #213, opened 2017-06-21
Break apart all the clients so they are easier to consume


rthallisey:
All the clients: docker, etcd, kubernetes, and coming soon Openshift, are all created in different locations. In some cases, we're even creating the clients each time we perform an action. Let's pool all the client initialization into one directory so they can all be called at once.

PR #214, opened 2017-06-21
Break apart the Kubernetes Client

rthallisey:
Make a separate directory for clients so they are not part of the APB code.  This will make the clients
easy to initialize, a single place to create the object, and adds the ability to easily create new ones.

Partially-fixes: https://github.com/openshift/ansible-service-broker/issues/213

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -11,9 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/pborman/uuid"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/util/homedir"
 	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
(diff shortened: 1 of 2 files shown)


rthallisey:
Testing: Run the normal broker tests.

pkg/apb/client.go
```diff
@@ -11,9 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 
+	"github.com/openshift/ansible-service-broker/pkg/clients"
```
jmrodri: Wow I never realized how ugly our import is :( not really a problem with this PR, just something I noticed finally. Would be cool if we could get: `openshift.io/asb/`

jmrodri: On the plus side importing our a `clients` package is good.

pkg/apb/client.go
```diff
@@ -56,52 +55,23 @@ type Client struct {
...
-		log.Debug("Checking for a local Cluster Config")
-		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
-		if err != nil {
-			log.Error("Failed to create LocalClientSet")
-			return nil, err
-		}
-	}
-
-	clientset, err := clientset.NewForConfig(clientConfig)
+	k8s, err := clients.Kubernetes(log)
```
jmrodri: I like the simplicity of this. I do not like how the `Client` struct has a bunch of clients in it. Will keep looking at this and other PRs to see if I understand the need for it. I know originally it was
```
type Client struct {
    dockerClient ...
}
```

pkg/clients/kubernetes.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	config, err := clientcmd.NewDefaultClientConfig(*clientConfig, &clientcmd.ConfigOverrides{}).ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+func Kubernetes(log *logging.Logger) (*clientset.Clientset, error) {
```
jmrodri: This is definitely cleaner. :+1:

jmrodri:
I'm going to approve this on my side. The main concern is similar to @eriknelson need to remove Clients struct. But the changes here seem reasonable.

eriknelson:
@jmrodri Plan is for me to follow up this series with a PR that removes the Client, which is reasonable because this is a good step forwards and looks stable.

rthallisey:
Working in #222

PR #219, opened 2017-06-22
Fix `make deploy`

fabianvf:
also make deploy accept all keys so it's configurable for up and downstream.

Haven't had a chance to test yet due to busted environment, but verified the generated openshift yaml file looked correct.

### DIFF
--- a/templates/broker-deployment_template.yaml
+++ b/templates/broker-deployment_template.yaml
@@ -18,16 +18,52 @@ spec:
     spec:
       restartPolicy: Always
       containers:
-        - image: "{{dockerhub_org}}/ansible-service-broker:latest"
+        - image: {{registry_org}}/ansible-service-broker:latest
           name: asb
           imagePullPolicy: Always
           ports:
             - containerPort: 1338
               protocol: TCP
           env:
-            - name: DOCKERHUB_PASS
-              value: "{{dockerhub_pass}}"
-            - name: DOCKERHUB_USER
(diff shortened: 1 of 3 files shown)


jwmatthews:
What if we changed 'make deploy' to use the same template we use for deploying elsewhere

https://github.com/openshift/ansible-service-broker/blob/master/templates/deploy-ansible-service-broker.template.yaml

templates/broker-deployment_template.yaml
```diff
@@ -18,16 +18,52 @@ spec:
     spec:
       restartPolicy: Always
       containers:
-        - image: "{{dockerhub_org}}/ansible-service-broker:latest"
+        - image: {{registry_org}}/ansible-service-broker:latest
```
shawn-hurley: What if I want to pull ansible-service-broker from my org but get the apbs from ansibleplaybookbundle? I think re-using registry_org might make that hard?

eriknelson: +1, I was frequently doing this.

fabianvf:
@eriknelson comments addressed.

eriknelson:
Thanks @fabianvf, will review and try to get this merged today.

fabianvf:
Going with a completely different approach, these changes are no longer useful.

jmrodri:
Please do not work on deploy. I'm working on card that will update it.

PR #220, opened 2017-06-22
use the correct source name in the rpm spec

jmontleon:


jmontleon:
It looks like now that we have a proper version it's creating the archive with the version number included:
https://copr-be.cloud.fedoraproject.org/results/%40ansible-service-broker/ansible-service-broker/epel-7-x86_64/00569698-ansible-service-broker/builder-live.log

ansible-service-broker-0.9.3.tar.gz: OK

PR #222, opened 2017-06-22
Breakup all the Broker Clients into a clients pkg

rthallisey:
Refactor the broker clients into their own package.

Right now, we have clients scattered around in different packages. This is causing the clients to be initialized on each action (bind, unbind, etc...) as well as other places because the clients aren't available for use.

Let's initialize the clients one time and distribute those clients to the places that need it.  This patch set creates a global struct 'Clients' that gets initialized by app.go before the broker object is created.  Any function that runs after the broker object creation, will have access to all the clients.

Fixes: #213

### DIFF
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -54,23 +54,15 @@ type Client struct {
 }
 
 func NewClient(log *logging.Logger) (*Client, error) {
-	dockerClient, err := clients.Docker(log)
-	if err != nil {
-		log.Error("Could not load docker client")
-		return nil, err
-	}
-
-	k8s, err := clients.Kubernetes(log)
-	if err != nil {
-		return nil, err
-	}
-
-	rest := k8s.CoreV1().RESTClient()
+	//TODO: This object gets created each provision, bind, deprovision,
+	// and unbind.  Instead, those functions should be using the global
+	// clients were needed and this class needs to be reworked.
+	k8s := clients.Clients.KubernetesClient
 
 	client := &Client{
-		dockerClient:  dockerClient,
+		dockerClient:  clients.Clients.DockerClient,
 		ClusterClient: k8s,
-		RESTClient:    rest,
+		RESTClient:    k8s.CoreV1().RESTClient(),
 		log:           log,
 	}
 
@@ -146,14 +138,14 @@ func (c *Client) RunImage(
 	}
 
 	c.log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, ns))
-	_, err = c.ClusterClient.CoreV1().Pods(ns).Create(pod)
+	_, err = clients.Clients.KubernetesClient.CoreV1().Pods(ns).Create(pod)
 
 	return apbId, err
 }
 
 func (c *Client) PullImage(imageName string) error {
 	// Under what circumstances does this error out?
-	c.dockerClient.PullImage(docker.PullImageOptions{
+	clients.Clients.DockerClient.PullImage(docker.PullImageOptions{
 		Repository:   imageName,
 		OutputStream: os.Stdout,
 	}, docker.AuthConfiguration{})
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -39,8 +39,6 @@ First cut might have to pass kubecfg from broker. FIRST SPRINT broker passes use
 admin/admin
 */
 
-var DockerSocket = "unix:///var/run/docker.sock"
-
 type ClusterConfig struct {
 	InCluster bool
 	Target    string
@@ -56,7 +54,7 @@ type Client struct {
 }
 
 func NewClient(log *logging.Logger) (*Client, error) {
-	dockerClient, err := docker.NewClient(DockerSocket)
+	dockerClient, err := clients.Docker(log)
 	if err != nil {
 		log.Error("Could not load docker client")
 		return nil, err
--- a/pkg/apb/client.go
+++ b/pkg/apb/client.go
@@ -11,9 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/pborman/uuid"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/util/homedir"
 	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
@@ -56,19 +55,6 @@ type Client struct {
 	log           *logging.Logger
 }
 
-func createClientConfigFromFile(configPath string) (*restclient.Config, error) {
-	clientConfig, err := clientcmd.LoadFromFile(configPath)
-	if err != nil {
-		return nil, err
-	}
-
-	config, err := clientcmd.NewDefaultClientConfig(*clientConfig, &clientcmd.ConfigOverrides{}).ClientConfig()
-	if err != nil {
-		return nil, err
-	}
-	return config, nil
-}
-
 func NewClient(log *logging.Logger) (*Client, error) {
 	dockerClient, err := docker.NewClient(DockerSocket)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


rthallisey:
Testing: Run the normal broker tests.

pkg/apb/client.go
```diff
@@ -56,52 +53,23 @@ type Client struct {
...
-	}
-
-	config, err := clientcmd.NewDefaultClientConfig(*clientConfig, &clientcmd.ConfigOverrides{}).ClientConfig()
-	if err != nil {
-		return nil, err
-	}
-	return config, nil
-}
-
 func NewClient(log *logging.Logger) (*Client, error) {
```
eriknelson: Can we push this into the clients package? I think it makes more sense there. Related to my overall comment on apb/client.go

rthallisey: @eriknelson I agree, but this is a much larger change since Client is a class.  So things like RunImage need to be distributed out of client.go.  For now, I'm going to keep this in place, but adding a TODO in the final patch for this series and we can return to it later.

eriknelson:
Is this PR a strict superset of the docker and k8s ones? Should all the changes just get make and addressed here and we can close the others, pulling it all in from this?

rthallisey:
@eriknelson They are each a super set of the previous patch.  Personally, I'd like for us to discuss in each patch the code that each one touches. I think it's much easier to review because this entire patch set is large and I have one more on the way ;). Also, I was careful not to have overlap, so anything in the docker patch only touches docker client code. The large patch that's coming last will touch a bit of everything.

If we want to talk about overall structure, then the final patch at the end will make sense since it will have the final picture.

pkg/clients/etcd.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"time"
+
+	logging "github.com/op/go-logging"
+
+	"github.com/coreos/etcd/client"
+)
+
+type EtcdConfig struct {
```
shawn-hurley: > Speaking of clients.() methods, do we want any of these to be singletons? 

singleton object I think that we just create them during initialization of application with the configuration of the clients. Allows us to error early if something is going to go wrong imo.

> I think I would have preferred for the PRs to have come in as one unit. Made it difficult to review the set since it looks like C builds on B builds on A, and they all introduce new features on top of the others.
 
wish that at least on comments that told us the order, which could be added.

> We should get rid of [this struct](https://github.com/rthallisey/ansible-service-broker/blob/fe5a7c1ee999858d4d30bd55bfc1761f7357b26f/pkg/apb/client.go#L49)
:+1: 

> I like the clients pkg a lot, decouples things in a way that makes changes a lot less brittle (client init code isn't just thrown about where it's used; it's got a home). Also going to make it real easy and natural to add an Origin() client.
:+1: 

>  Semantically, it's the part of the apb package that's responsible for actually executing apb methods. Could almost be apb/apb.go? Curious what you guys think.

To me, it looks like a lot of the ABP code is in their own files. and client.go should all be in the new package. I am wondering what methods you were thinking should be left in apb?

jmrodri:
@rthallisey the individual patches actually make it harder to see on github. Since the diff of the last one is ALL of them. It's hard to see what was just for the last patch vs the first one. And while looking at the first one, it's hard to see where you are going with it in the final implementation. :smile:

pkg/app/config.go
```diff
@@ -8,12 +8,12 @@ import (
...
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 	"github.com/openshift/ansible-service-broker/pkg/broker"
-	"github.com/openshift/ansible-service-broker/pkg/dao"
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 )
 
 type Config struct {
 	Registry   apb.RegistryConfig
-	Dao        dao.Config
+	Dao        clients.EtcdConfig
```
jmrodri: This will mean that if we ever want to add some configuration to the dao that is not etcd related we'd have to redo this. dao.Config was easier so we could configure all of the dao. I see how you arrived here since the only configuration items on the dao were etcd related, but it paints in a corner of only being etcd.

pkg/clients/etcd.go
jmrodri: @eriknelson +1 to `EtcdConfig` remaining `Config`. I just added that as a comment to my review.

eriknelson: > To me, it looks like a lot of the ABP code is in their own files. and client.go should all be in the new package. I am wondering what methods you were thinking should be left in apb?

I don't think code that's responsible for the execution of an apb belongs in the `clients` pkg, it's a natural fit for the `apb` pkg. It's effectively `ExecuteApb(method, parameters)`, and we should hide the guts of **how** that actually gets done behind that function (in this case, it's running the pod). `ExecuteApb` IMO is a better name for `RunImage`. Run image bleeds the abstraction with "Image" in the name. It suggests "run this apb method with these parameters, I don't care how you do that" vs "run this image because that's how we happen to run apbs".

jmrodri: > Speaking of clients.() methods, do we want any of these to be singletons?

This sounds ok, we need to make sure we test multiple users of the SAME client at the SAME time. Will that work ok? IS it really a 'single' connection? The old way we created a new connection so if there were multiple provisions going on each worker would have their own connection.

jmrodri: > We should get rid of [this struct](https://github.com/rthallisey/ansible-service-broker/blob/fe5a7c1ee999858d4d30bd55bfc1761f7357b26f/pkg/apb/client.go#L49)

I also concur with this :+1:

shawn-hurley:
I am wondering if we are even using the docker client. Pull Image appears to be the only thing that uses the dockerClient and that is not being called anywhere. I think we should consider removing this.

I also would like to note that we should add doc comments on the new exported structs/functions/constants et all.

Lastly, and this is just an idea/thing I want to throw out there. using individual clients should be in their own package. example client/kubernetes/client.go the client object could import this and do the above workflow but just a better speration of the clients.

pkg/dao/dao.go
```diff
@@ -7,52 +7,31 @@ import (
...
-}
-
-// Dao - retreive, create, and update  objects from etcd
 type Dao struct {
-	config    Config
-	log       *logging.Logger
-	endpoints []string
-	client    client.Client
-	kapi      client.KeysAPI // Used to interact with kvp API over HTTP
+	config clients.EtcdConfig
```
jmrodri: should be dao.Config

pkg/apb/client.go
eriknelson: @rthallisey since I'm requesting the change, mind if I take a crack at it and submit for your review? Think I could probably just explain myself better with code anyways.

rthallisey: @eriknelson That would be fantastic 👍

rthallisey:
@jmrodri wfm. One more on the way

eriknelson:
@shawn-hurley
> singleton object I think that we just create them during initialization of application with the configuration of the clients. Allows us to error early if something is going to go wrong imo.

Good argument over lazy singleton, fail fast and early.

> I am wondering if we are even using the docker client. Pull Image appears to be the only thing that uses the dockerClient and that is not being called anywhere. I think we should consider removing this.

I think you're right. It's a vestige from when we pulled images before running then with `oc run`. I don't remember why we needed the pull in the first place, it's possible we never did. It's a use case entirely covered by the Pod `pullPolicy`. Huge bonus points if it means we can shed a dependency on docker entirely. We should definitely look into this, probably appropriate for an issue and follow up PR.

> I also would like to note that we should add doc comments on the new exported structs/functions/constants et all.
:+1: , this a style thing for the linter? Obviously a good habit.

> Lastly, and this is just an idea/thing I want to throw out there. using individual clients should be in their own package. example client/kubernetes/client.go the client object could import this and do the above workflow but just a better speration of the clients.

What's the benefit of having a package for every client? I understand it would separate things more, just don't see a case for something I can't do with them all in the same package. It also feels like we're blowing up our package count, I'm not familiar enough with go conventions to know if this is frowned upon or it's normal matter of course, but it feels excessive.

shawn-hurley:
I think the major benefit is the code breaking up into their own little place. because of the way go does packages just be in their own file doesn't break it up for the caller. So, for instance, Origin client has a struct that we want to call ServiceRoleBinding but so does Kubernetes and they are different. Rather than having to name them OriginServiceRoleBinding and KuberneetesServiceRoleBinding you name it ServiceRoleBinding and use the package to tell which one to use. I think it makes for cleaner code. If this is not a use case we are worried about then yes I agree it is not worth the extra packages.

eriknelson:
@shawn-hurley I'd propose we start with one package and if we start to see naming collisions, we can break things out.

rthallisey:
git didn't like the branch rename :). We'll leave it as is

rthallisey:
Sorry for the confusion on the PRs :).  That experiment didn't work

rthallisey:
> I think the major benefit is the code breaking up into their own little place. because of the way go does packages just be in their own file doesn't break it up for the caller. So, for instance, Origin client has a struct that we want to call ServiceRoleBinding but so does Kubernetes and they are different. Rather than having to name them OriginServiceRoleBinding and KuberneetesServiceRoleBinding you name it ServiceRoleBinding and use the package to tell which one to use. I think it makes for cleaner code. If this is not a use case we are >  worried about then yes I agree it is not worth the extra packages.

I think that the clients should stay in the clients package because they're just objects.  They aren't classes. We aren't building a kubernetes or docker client, we're creating a variable to call the clients efficiently.

eriknelson:
Going to test this now in an integrated environment to make sure things are still running smoothly, but the current thinking after speak with the guys offline is we're ready to merge this into master. Doc strings can come in with a follow up PR (there's one out there to become golint compliant), and I will be following this up with a change that already has the DaoConfig/EtcdConfig requested change.

eriknelson:
ACK, tested and was able to provision things fine with this on top of master. Merging.

PR #223, opened 2017-06-22
With a newer Etcd, we can use the GetVersion function

rthallisey:
The GetEtcdVersion function was a temporary workaround to contact Etcd until we moved to a newer etcd. The change we needed merged in v3.1.1.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -68,14 +69,14 @@ func CreateApp() App {
 	}
 
 	// Initializing clients as soon as we have deps ready.
-	initClients(app.log.Logger, app.config.Dao.GetEtcdConfig())
-
-	app.log.Debug("Connecting Dao")
-	app.dao, err = dao.NewDao(app.config.Dao, app.log.Logger)
+	err = initClients(app.log.Logger, app.config.Dao.GetEtcdConfig())
 	if err != nil {
 		app.log.Error(err.Error())
 		os.Exit(1)
 	}
+
+	app.log.Debug("Connecting Dao")
+	app.dao, err = dao.NewDao(app.config.Dao, app.log.Logger)
 
 	k8scli, err := clients.Kubernetes(app.log.Logger)
 	if err != nil {
@@ -164,7 +165,7 @@ func (a *App) Start() {
 	}
 }
 
-func initClients(log *logging.Logger, ec clients.EtcdConfig) {
+func initClients(log *logging.Logger, ec clients.EtcdConfig) error {
 	// Designed to panic early if we cannot construct required clients.
 	// this likely means we're in an unrecoverable configuration or enviornment.
 	// Best we can do is alert the operator as early as possible.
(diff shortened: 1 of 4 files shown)


rthallisey:
Testing: Run the broker normally.  You should see the Etcd server and client versions in the logs.

pkg/dao/dao.go
```diff
@@ -27,7 +23,7 @@ type Dao struct {
 	config    Config
```
rthallisey: maybe I should capitalize everything in here..

jwmatthews:
Confirmed that we are OK assuming we have 3.1.1+ in regard to what we use for production deployments.

The version of etcd currently delivered from RHCC images is 3.1.7
https://access.redhat.com/containers/#/registry.access.redhat.com/rhel7/etcd

pkg/dao/dao.go
eriknelson: The etcd client was deliberately private because the Dao is intended to be an abstraction layer on top of the models with controlled public access. I don't think business logic should be talking to etcd directly; it's an encapsulated implementation detail. I'd prefer to keep it private to prevent breaking that design.

rthallisey: @eriknelson Sounds good. I'll find a way around it.

eriknelson: Thinking about it more, the app.go code should delegate the validation of a data connection to the dao and the dao determines what that means since it really belongs in the domain of that package. A good indicator of that abstraction leak is if other code in the broker is importing etcd/client.

shawn-hurley: I disagree slightly, the broker looks at the errors that are returned from the DAO as it could be etcd errors. 

We could get around that by adding another layer of predefined errors in the dao and in the broker.

pkg/app/app.go
```diff
@@ -66,13 +67,16 @@ func CreateApp() App {
 		app.log.Error(err.Error())
 		os.Exit(1)
 	}
-	serv, clust, err := app.dao.GetEtcdVersion(app.config.Dao)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	version, err := app.dao.Client.GetVersion(ctx)
```
jmrodri: You could wrap this in a public method on dao.
```
func GetEtcdVersion(ctx...) {
return client.GetVersion(ctx)
}
```
That way you don't have to leak the client out only expose what you need to validate.

pkg/dao/dao.go
jmrodri: > Thinking about it more, the app.go code should delegate the validation of a data connection to the dao and the dao determines what that means since it really belongs in the domain of that package. A good indicator of that abstraction leak is if other code in the broker is importing etcd/client.

HEH I had a similar thought. Though I know like @shawn-hurley said the broker now imports the `etcd/client` simply to verify the errors that come back from it. I think it might be a good idea to define errors the broker knows about and keep the low level errors down in their own layers. i.e. broker shouldn't get an etcd error? but that would require some nasty handling since there are no Exceptions :)

eriknelson: > I disagree slightly, the broker looks at the errors that are returned from the DAO as it could be etcd errors.

Fair. I don't think we should be doing that either since it's another abstraction leak for the same reasons :)

> I think it might be a good idea to define errors the broker knows about and keep the low level errors down in their own layers. i.e. broker shouldn't get an etcd error? but that would require some nasty handling since there are no Exceptions :)

I like this. Don't like go error handling, but we don't have a choice if this is the road we want to head down ¯\_(ツ)_/¯

Dao errors make sense to me. `NoConnection`, `FailedToRetrieve`, `NotFound`, things on that level.

pkg/app/app.go
eriknelson: Exactly what I had in mind.

rthallisey: This is going to end up using the new clients patch #222.  I'll wait for that to merge so I can avoid exposing dao.

rthallisey:
Depends-on: #222

jmrodri:
@eriknelson @rthallisey how is this compared to PR #238

rthallisey:
@jmrodri This uses the etcd client to call GetVersion instead of the GetEtcdVersion function.

eriknelson:
@jmrodri @rthallisey I'd like to get #238 into master and rebase this if you guys are cool with that. Main differences were:

* No reason for GetEtcdVersion to be a stateful method
* Because of ^, semantically, it really belongs in clients/etcd.go, so I moved it

eriknelson:
#238 merged, needs rebase.

rthallisey:
@eriknelson All set!  I added some error checking and changed the etcd var since it can't be used as a pointer.

pkg/clients/clients.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 var instances struct {
-	Etcd       *etcd.Client
+	Etcd       etcd.Client
```
eriknelson: Why is this not a pointer? I moved all the clients to be pointers for reference and consistency. If it's a light object, I don't care, IMO the pointer was cheap and we're referring to the same object anyways with a singleton.

rthallisey: In order to call GetVersion, you need the client interface object and not a pointer.

jmrodri: Ok, looks like etcd `Client` is an interface. 

```
type Client interface {
```

And when you change the above to a pointer, you get:

```
$ make 
# github.com/openshift/ansible-service-broker/pkg/app
pkg/app/app.go:187: etcdClient.GetVersion undefined (type *client.Client is pointer to interface, not interface)
Makefile:17: recipe for target 'broker' failed
make: *** [broker] Error 2
```

So +1 to this staying `etcd.Client`.

eriknelson: Gotcha! That makes sense.

PR #225, opened 2017-06-23
Golint fixes

shawn-hurley:
* removing errors because it is not used.

I am attempting to fix the overall golint issues that we have.


pkg/app/app.go
```diff
@@ -15,8 +15,13 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/handler"
 )
 
+// MsgBufferSize - The buffer for the message channel.
 const MsgBufferSize = 20
 
+// Version - the current version of the ansbile service broker. Should be updated by tito.
+var Version = "v0.1.0"
```
jmrodri: This line was replaced in master. Are you up to date?

jmrodri:
I clicked too soon. More comments on the way

pkg/app/app.go
shawn-hurley: No, I pushing up more commits as I go this is very much a WIP

jmrodri:
Is this still WIP? if not please rename the PR.

pkg/apb/svc_acct.go
```diff
@@ -86,11 +86,7 @@ func (s *ServiceAccountManager) createResources(rFilePath string, namespace stri
 	return nil
 }
 
-func (s *ServiceAccountManager) writeResourceFile(
-	handle string,
-	svcAcctM *map[string]interface{},
-	roleBindingM *map[string]interface{},
-) (string, error) {
+func (s *ServiceAccountManager) writeResourceFile(handle string, svcAcctM *map[string]interface{}, roleBindingM *map[string]interface{}) (string, error) {
```
eriknelson: I really dislike long signatures on a single line like this. Anyone else feel strongly about that? @jmrodri @rthallisey ?

eriknelson: I really dislike long signatures on a single line like this. Anyone else feel strongly about that? @jmrodri @rthallisey? If this is a linting problem, I'd like to disable it.

pkg/app/args.go
```diff
@@ -1,42 +1,29 @@
...
-
-	err = validateArgs(&args)
-	if err != nil {
-		return args, err
-	}
-
 	return args, nil
 }
 
-func validateArgs(args *Args) error {
```
eriknelson: :+1: , this was before we had a default value.

pkg/app/logger.go
```diff
@@ -8,20 +8,24 @@ import (
...
 }
 
+// Log - Logging struct that will contain https://godoc.org/os#File and the logging object.
 type Log struct {
 	*logging.Logger
 	file *os.File
 }
 
+// MODULE - Module for the logger.
 const MODULE = "asb"
```
eriknelson: Should this be `Module` instead of `MODULE`? What's our stance on const identifier style?

pkg/apb/svc_acct.go
rthallisey: I agree @eriknelson. I liked it the way it was.

jmrodri:
@shawn-hurley update the travis.sh to fail on golint errors:
https://github.com/openshift/ansible-service-broker/blob/master/scripts/travis.sh#L46

pkg/apb/svc_acct.go
jmrodri: I don't think golint should fail on long signatures. I usually write the long signature but I do not rewrite the shorter one if it already exists. I don't think we should fail on this. If we want to fail, then let's fail on LONG ones. I.e. make it shorter.

jmrodri:
On Satellite 5 we used 95 :) because some of us wanted 80 (me) and others wanted 120. We found the longest current line in the source, 95 and used that.

eriknelson:
95 works for me. I basically use 80 as a guideline, if I'm going over, I'll finish what I'm doing and newline or break things up.

jmrodri:
Longest line in `pkg` dir is **1521**. Longest line in `pkg` that is NOT a test is **155**. Longest line in cmd` is **103**.

```
$ find pkg -name '*.go'| grep -v _test | xargs wc -L | grep total
   155 total
```
```
$ find cmd -name '*.go'| grep -v _test | xargs wc -L | grep total
 103 total
```
```
$ find pkg -name '*.go'|  xargs wc -L | grep total
  1521 total

```

pkg/apb/svc_acct.go
shawn-hurley: I like the long lines better personally but it seems I am outvoted here. I'll change it back and update my environment to alert me.

shawn-hurley:
[(2) Fix golint errors in Broker source code](https://trello.com/c/i8L9H6U2/309-2-fix-golint-errors-in-broker-source-code)

shawn-hurley:
[(2) Fix golint errors in Broker source code](https://trello.com/c/i8L9H6U2/309-2-fix-golint-errors-in-broker-source-code)

PR #215, opened 2017-06-21
Pull out the docker client

rthallisey:
Take out the docker client creation from the apbs package and put it along side the kubernetes client.

Partially-fixes: #213 
Depends-on: #214

PR #228, opened 2017-06-23
Single client object

rthallisey:
Pull together all the client split work by using a global variable to pass around the clients.

Partially-fixes: #213


eriknelson:
Can we close this, push the last commit to #222, and close #214 and #215, then rename #222 to something like "Introduce Clients Package"? This is getting really confusing with all these PRs that share commits.

rthallisey:
Working in #222

PR #230, opened 2017-06-23
remove the mock-registry service

jmrodri:
We used to have a mock-registry service you could run with the broker
that was driven by a data file. This service ran seperately from the
broker.

This is NOT the same as the current mock_registry located IN the broker.


Makefile
```diff
@@ -65,12 +50,11 @@ push:
...
-	@rm -f mock-registry
 
 deploy:
 	@${GOPATH}/src/github.com/openshift/ansible-service-broker/scripts/deploy.sh
 
 test:
 	go test ./pkg/...
 
-.PHONY: vendor build install run uninstall build-mock-registry install-mock-registry run-mock-registry uninstall-mock-registry prepare-build-image build-image release-image release push clean deploy test
+.PHONY: vendor build install run uninstall -registry prepare-build-image build-image release-image release push clean deploy test
```
eriknelson: Little leftover "-registry" from "uninstall-mock-registry"

jmrodri:
@rthallisey the mock registry referenced in config.md is a *different* mock registry :) that registry uses dockerhub registry except sourced from a yaml file instead of talking to a remote server. The mock registry I was removing was a very very old registry which was a separate service. It has been defunct for quite some time.

jmrodri:
Closing this because it's not picking up my latest commit: https://github.com/openshift/ansible-service-broker/commit/377309d789f41ae0dab061f49c997c4eb92d18fc

Issue #227, opened 2017-06-23
Investigate removing dockerClient and dependency


eriknelson:
From discussion of #222 

> It's a vestige from when we pulled images before running then with oc run. I don't remember why we needed the pull in the first place, it's possible we never did. It's a use case entirely covered by the Pod pullPolicy. Huge bonus points if it means we can shed a dependency on docker entirely. We should definitely look into this, probably appropriate for an issue and follow up PR.

shawn-hurley:
This was closed with #232

PR #232, opened 2017-06-25
removing go-dockerclient

shawn-hurley:
Took first stab at #227. Leaving as do not merge, because I need some sanity checks.


rthallisey:
👍

PR #234, opened 2017-06-26
techdebt: speed up builds

jmrodri:


jmrodri:
The `-i` option will speed up subsequent builds dramatically. There is still a time hit for the very first build.
For each of the tests, I did a `make clean` before running `make build`. The first two were *not* using the `-i` option. The last 3 tests use `-i` option.

### long builds
#### first build after clean
```
$ time make build
go build -ldflags="-s -w" ./cmd/broker

real	0m28.204s
user	1m50.538s
sys	0m9.607s
```

#### second build after clean
```
$ time make build
go build -ldflags="-s -w" ./cmd/broker

real	0m28.027s
user	1m52.214s
sys	0m9.470s
```

### using -i 
#### first build after clean
```
$ time make build

real	0m28.493s
user	1m51.544s
sys	0m10.655s
```

#### second build after clean
```
$ time make build

real	0m3.205s
user	0m3.507s
sys	0m0.829s
```

#### third build after clean
```
$ time make build

real	0m3.194s
user	0m3.508s
sys	0m0.800s
```

jmrodri:
### TEST
clean and run build twice. The second build should be dramatically faster than the first.

`make clean`
`time make build`
`make clean`
`time make build`

Issue #235, opened 2017-06-26
Add configuration option to control Image Pull Policy for APBs


jwmatthews:
Let's create a configuration option so we can control this image pull policy for ABPs
https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/client.go#L172

IfNotPresent would help in several developer testing scenarios

PR #237, opened 2017-06-27
Make the ImagePullPolicy Configurable

rthallisey:
The pull policy of an APB should be set by the broker.
To configure it, set image_pull_policy in the config file.

The image_pull_policy setting will be verified that it's
a correct setting. The option is not case sensetive and will
error if an incorrect setting is used.

Fixes https://github.com/openshift/ansible-service-broker/issues/235

### DIFF
--- a/etc/example-config.yaml
+++ b/etc/example-config.yaml
@@ -17,6 +17,7 @@ openshift:
   target: localhost:8443
   user: OPENSHIFT_USER
   pass: OPENSHIFT_PASS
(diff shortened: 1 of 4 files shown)


jwmatthews:
Let's default this to IfNotPresent

rthallisey:
@jwmatthews Made the broker config default to IfNotPresent and the local config I left as Always. Atleast for me with dev work, I use Always because I often change around APBs to point at my own namespace. Let me know if you agree.

Issue #229, opened 2017-06-23
Consider singleton clients


eriknelson:
Arose out of #222 discussion.

It sounds like most people are in favor of singleton clients.

Considerations:
* Needs testing WRT concurrent usage of a since client instance (for all relevant clients).
* Should the instances be initialized and validated on first use, or on app startup? I think @shawn-hurley made some good arguments for the latter, fail fast and early.

rthallisey:
I think it's better to validate the clients before we create the broker.  It's a simple pre-flight check to make sure your environment is in order.  I also think these checks could be added to the client package so we can reduce the size of app.go.

eriknelson:
> I think it's better to validate the clients before we create the broker. It's a simple pre-flight check to make sure your environment is in order. I also think these checks could be added to the client package so we can reduce the size of app.go.

Agreed. When I was looking at refactoring things it seemed like there were a lot of blocks in `App.Start` that were written to instantiate/init/validate indirectly by just creating references. Think those init's and checks should be behind a func named appropriately like `InitAndValidate` called from `App.Start`, and then we can get rid of all of the random inits and exit 1 if anything blows up.

shawn-hurley:
@eriknelson I think we can close this issue yes?

Issue #231, opened 2017-06-23
Remove apb/client.go Client struct and related refactoring.


eriknelson:
Arose out of #222 discussion.

Client struct in apb/client.go simply aggregates references to the various clients, which doesn't provide much value. It should be removed along with the `NewClient` constructor.

Additionally, pushing all the client related logic into the `clients` pkg brings into question whether `apb/client.go` is really an appropriate name for the file. Given the work being done (actually executing apb methods), consider a better name `RunApb(method, params)`.

jmrodri:
Fixed by PR #238

PR #238, opened 2017-06-27
Refactor apb/client contents

eriknelson:
~~Marking as DO NOT MERGE because I need to make sure the rebase on top of the docker removal didn't corrupt anything. Still open for comment since hopefully nothing major will change.~~
**Ready for review**

First commit is a squashed version of what I had in my personal org, prior discussion: https://github.com/eriknelson/ansible-service-broker/pull/2

Turns out, embedding EtcdConfig into the DaoConfig will not play nice with the yaml unmarshalling; it doesn't respect the annotations of the embedded type (correct me if I'm wrong!). I just kept them as fields on the Dao and added a convenience getter for EtcdConfigs.

Needed to manually remove some things as a result of the rebase for the third commit. Making sure this still makes sense.

Still looking to add one more commit with feedback items from my linked PR.


eriknelson:
Rebased to pick up golinter compliance PR.

eriknelson:
Of particular attention on this one is moving towards panicing the client constructors if errors are reported, and initializing everything as early as we can in `App::Start`. Thanks to @shawn-hurley for the suggestion, nicely cleans up `clients` package. If you can't get at your clients, you're probably FUBAR anyway and need to let someone know to go fix things.

pkg/app/app.go
```diff
@@ -170,3 +154,26 @@ func (a *App) Start() {
...
+	//
+	// Deliberately forcing the injection of deps here instead of running as a
+	// method on the app. Forces developers at authorship time to think about
+	// dependencies / make sure things are ready.
+	log.Notice("Initializing clients...")
+	log.Debug("Trying to connect to etcd")
+	serv, clust, err := clients.GetEtcdVersion(ec)
+	if err != nil {
+		log.Error("Failed to connect to Etcd\n")
+		log.Error(err.Error())
```
jmrodri: Why not make it the same log? You did that in another place `log.Error("Failed to connect to Etcd: %s", err.Error())`

pkg/clients/etcd.go
```diff
@@ -15,6 +19,34 @@ type EtcdConfig struct {
 	EtcdPort string `yaml:"etcd_port"`
 }
 
+func GetEtcdVersion(ec EtcdConfig) (string, string, error) {
+	// The next etcd release (1.4) will have client.GetVersion()
```
jmrodri: :+1: `etcd.go` is the proper place for `GetEtcdVersion` I like it.

pkg/clients/etcd.go
```diff
@@ -15,6 +19,34 @@ type EtcdConfig struct {
...
+		if err := json.Unmarshal(body, &vresp); err != nil {
+			return "", "", err
+		}
+		return vresp.Server, vresp.Cluster, nil
+	default:
+		var connectErr error
+		if err := json.Unmarshal(body, &connectErr); err != nil {
+			return "", "", err
+		}
+		return "", "", connectErr
```
jmrodri: I would've returned "unknown" instead of "".

jmrodri: Hrm just realized the error will be handled and the empty strings won't even be looked at. Disregard last comment :)

pkg/apb/provision.go
```diff
@@ -55,17 +55,13 @@ func Provision(
...
+	podName, err := ExecuteApb(
+		"provision", clusterConfig, instance.Spec,
+		instance.Context, instance.Parameters, log,
+	)
 
 	if err != nil {
-		log.Error("Problem running image")
+		log.Error("Problem executing apb")
 		log.Error(string(podName))
 		log.Error(err.Error())
```
jmrodri: I would collapse these:

```
log.Error("Problem executing abp [%s]. Error occurred: %s", string(podName), err.Error())
```

pkg/clients/clients.go
```diff
@@ -1,14 +1,17 @@
 package clients
 
 import (
-	"github.com/coreos/etcd/client"
-	"k8s.io/client-go/rest"
-	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
+	etcd "github.com/coreos/etcd/client"
+	k8s "k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
+	"sync"
```
jmrodri: `goimports` would actually put "sync" at the top because it's from the stdlib. 
```
import (
    "sync"
    
    etcd "github.com/coreos/etcd/client"
    k8s "k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
)
```

pkg/clients/etcd.go
```diff
@@ -1,12 +1,16 @@
 package clients
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/coreos/etcd/version"
+	"io/ioutil"
+	"net/http"
 	"time"
```
jmrodri: goimports reordered these for me too.
```
import (
    "encoding/json"
    "fmt"
    "io/ioutil"
    "net/http"
    "time"

    "github.com/coreos/etcd/version"

    etcd "github.com/coreos/etcd/client"
)
```

pkg/app/app.go
shawn-hurley: +1 if you make it ```log.Errorf```

pkg/clients/kubernetes.go
```diff
@@ -9,6 +9,24 @@ import (
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
 
+// Kubernetes - Create a new kubernetes client if needed, returns reference
+func Kubernetes(log *logging.Logger) *clientset.Clientset {
```
shawn-hurley: The only other thing that I might want to change is to add an error to this call. 
I could see a time when we might be needing a pool, and we would use this method to retrieve values and would want errors. I also think that making sure the value is not nil, would be something that could be useful right now.
Anyone have strong feelings this is just an idea I don't fee strongly about it.

Makefile
```diff
@@ -6,11 +6,12 @@ BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/bu
 PREFIX           ?= /usr/local
 BROKER_CONFIG    ?= $(PWD)/etc/generated_local_development.yaml
 
+vendor:
```
eriknelson: HAH. This was me putting build first so I could lazily run `make` to rebuild. Then you educated me on DEFAULT_GOAL, sensei.

pkg/app/app.go
eriknelson: Honestly I way prefer to see things on new lines so I'm not reading shitty wrapped stack traces or errors. Will update to Errorf.

pkg/clients/clients.go
eriknelson: I literally can't save the file and format it this way with my current setup. vim-go orders alphabetically. I'll follow up with you to get my env using goimports, it's always been broken for me.

eriknelson:
Added PR addressing review feedback. Mainly, @shawn-hurley had some good suggestions for the interface accessors to nil check and return errors.

pkg/clients/clients.go
eriknelson: @jmrodri if only I had some kind if Integrated Development Environment to help me with these mundane tasks...

pkg/apb/bind.go
```diff
@@ -22,17 +22,13 @@ func Bind(instance *ServiceInstance, parameters *Parameters, clusterConfig Clust
...
-
-	podName, err := client.RunImage("bind", clusterConfig, instance.Spec, instance.Context, parameters)
+	podName, err := ExecuteApb(
+		"bind", clusterConfig, instance.Spec,
+		instance.Context, parameters, log,
+	)
 
 	if err != nil {
-		log.Error("Problem running image", err)
+		log.Error("Problem executing apb: %s", err)
```
rthallisey: We will print out the error message later on, but we won't have podName.  I think it would be more beneficial to print out podName here instead of err. Copy from provision.go

pkg/apb/executor.go
```diff
@@ -0,0 +1,124 @@
...
+	log.Debug("action:[ %s ]", action)
+
+	// It's a critical error if a Namespace is not provided to the
+	// broker because its required to know where to execute the pods and
+	// sandbox them based on that Namespace. Should fail fast and loud,
+	// with controlled error handling.
+	if context.Namespace == "" {
+		errStr := "Namespace not found within request context. Cannot perform requested " + action
+		log.Error(errStr)
+		return "", errors.New(errStr)
```
rthallisey: Could replace with ```return "", fmt.Errorf("Namespace not found within request context. Cannot perform requested %s", action)```.  I think no matter what this message will wrap over to the next line.  Either way I'm fine with it.

eriknelson: +1. I don't like the length and normally endeavor to deal with it, but this is not a big deal.

eriknelson:
Merging this based on ACKS and IRC OKs. @jmrodri if you got a problem, yell at me :)

jmrodri:
Fixes Issue #229 and #231

PR #240, opened 2017-06-27
Automated builds from Dockerhub

rthallisey:
The docker registry has ability to automatically build containers.
But, the container expects the latest source. We'll use automated-build
to represent the automated build Dockerfile.

Tested this change out here: https://hub.docker.com/r/rthallisey/ansible-service-broker-test-builds/builds/

### DIFF
--- a/build/automated-build
+++ b/build/automated-build
@@ -33,15 +33,12 @@ RUN yum -y install golang make device-mapper-devel btrfs-progs-devel etcd
 
 ENV GOPATH=/go
 ENV BROKER_PATH=${GOPATH}/src/github.com/openshift
-ENV GLIDE_VERSION=v0.12.3 \
-    GLIDE_HOME=/go/src/github.com/openshift/ansible-service-broker/.glide
-RUN curl -sSL https://github.com/Masterminds/glide/releases/download/$GLIDE_VERSION/glide-$GLIDE_VERSION-linux-amd64.tar.gz | tar -vxz -C /usr/local/bin --strip=1
 
 RUN mkdir -p /go/src/github.com/openshift/ansible-service-broker
 RUN git clone https://github.com/openshift/ansible-service-broker ${BROKER_PATH}/ansible-service-broker
 WORKDIR /go/src/github.com/openshift/ansible-service-broker
 
-RUN go build -ldflags="-s -w" ./cmd/broker
+RUN go build -i -ldflags="-s -w" ./cmd/broker
 RUN mv broker /usr/bin/asbd
 ######################
 # BUILD BROKER SOURCE
--- a/build/automated-build
+++ b/build/automated-build
@@ -1,0 +1,59 @@
+FROM centos:7
+MAINTAINER Ansible Service Broker Community
+
+ENV USER_NAME=ansibleservicebroker \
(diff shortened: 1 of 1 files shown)


build/automated-build
```diff
@@ -0,0 +1,59 @@
...
+ENV BROKER_PATH=${GOPATH}/src/github.com/openshift
+ENV GLIDE_VERSION=v0.12.3 \
+    GLIDE_HOME=/go/src/github.com/openshift/ansible-service-broker/.glide
+RUN curl -sSL https://github.com/Masterminds/glide/releases/download/$GLIDE_VERSION/glide-$GLIDE_VERSION-linux-amd64.tar.gz | tar -vxz -C /usr/local/bin --strip=1
+
+RUN mkdir -p /go/src/github.com/openshift/ansible-service-broker
+RUN git clone https://github.com/openshift/ansible-service-broker ${BROKER_PATH}/ansible-service-broker
+WORKDIR /go/src/github.com/openshift/ansible-service-broker
+
+RUN go build -ldflags="-s -w" ./cmd/broker
```
jmrodri: `go build -i -ldflags="-s -w" ./cmd/broker`

build/automated-build
```diff
@@ -0,0 +1,59 @@
...
+######################
+RUN rpm --import https://mirror.go-repo.io/centos/RPM-GPG-KEY-GO-REPO
+RUN curl -s https://mirror.go-repo.io/centos/go-repo.repo | tee /etc/yum.repos.d/go-repo.repo
+RUN yum -y install golang make device-mapper-devel btrfs-progs-devel etcd
+
+ENV GOPATH=/go
+ENV BROKER_PATH=${GOPATH}/src/github.com/openshift
+ENV GLIDE_VERSION=v0.12.3 \
+    GLIDE_HOME=/go/src/github.com/openshift/ansible-service-broker/.glide
+RUN curl -sSL https://github.com/Masterminds/glide/releases/download/$GLIDE_VERSION/glide-$GLIDE_VERSION-linux-amd64.tar.gz | tar -vxz -C /usr/local/bin --strip=1
```
jmrodri: Unfortunately we probably don't need any of the glide stuff in here since the damn vendor directory was committed. Maybe comment them out in case we need to re-enable.

rthallisey: true. Forgot to remove it.

rthallisey:
The downside to this is it's another Dockerfile we have to maintain. I'm all for iterating on this though to close that gap between Dockerfile-src and automated-build.

build/automated-build
eriknelson: @jmrodri can you push to get a card on a sprint to remove the committed vendor and take that on? I'm really not happy with it either and you've certainly been the loudest opponent :)

jmontleon:
ACK.

Related on a tangent... Do we want to push stable images periodically? Like 0.9.x when we're good, built from an RPM? I think it would be ok to have latest be from source? Certainly good with any automated builds being from source however all the repo:tag stuff pans out.

rthallisey:
@jmontleon The automated build also has a trigger.  So whenever a copr rebuild is triggered maybe we can also trigger a binary build in the dockerhub.

PR #243, opened 2017-06-28
Fixes BZ#1466031 add Accept header with application/json to RHCC get

jmontleon:


jmontleon:
Thanks @cfchase for pointing me in the right direction.

jwmatthews:
I'm good with this going into release-0.9

Issue #256, opened 2017-07-03
Proposal: Multiple Registry’s


shawn-hurley:
# Proposal: Multiple Registry’s 
 
## Goal: To allow users to provide multiple registries from the config. We should pull APB’s from all of the registries configured 
Trello Card: https://trello.com/c/1KqTacLK/304-8-multiple-registry-adapters-configured-in-one-broker-instance
 
 
### Changes to be made:
* app.App will be changed to have a slice of registries
* broker.Broker will be changed to have a slice of registries
* app.Config will contain a list of Registry Configs 
* Bootstrap will loop through each configured registry and Load Specs
* Change registry config to have a Type and Name therefore users can name their registries.
* Combine logic of registries to simplify workflow - #180 
 
 
### Outstanding Questions
Questions with my thoughts underneath. I am looking for feedback on the options proposed.
1. #### How to deal with apb id collision
  * Easiest solution is to error out and blow up. I don’t think that an ID collision should cause the broker to die
  * Another idea is to warn if this occurs and take the one that is being entered.
  * Redesign the data storage, so that registry -> spec. This will change the ID’s of everything away from UUID OR we create a mapping in etcd 
2. #### What happens if a registry fails to load.
  * Initial thought is to error out.
  * We can add a configuration that will allow an person to determine if a registry should fail the broker or not.
  * We can continue and warn if and only error if all registries have an error.
3. #### Should loading of specs be parallelised 
  * Initial thought is simpler is easier to understand, we can go back and add goroutines at a later time.
  * If this is the case then we need to make sure 2 makes sense.

shawn-hurley:
![](https://github.trello.services/images/mini-trello-icon.png) [(8) Multiple Registry Adapters configured in one Broker Instance](https://trello.com/c/1KqTacLK/304-8-multiple-registry-adapters-configured-in-one-broker-instance)

rthallisey:
> How to deal with apb id collision
>
>    Easiest solution is to error out and blow up. I don’t think that an ID collision should cause the broker to die
    Another idea is to warn if this occurs and take the one that is being entered.
    Redesign the data storage, so that registry -> spec. This will change the ID’s of everything away from UUID OR we create a mapping in etcd
>
How much do we care about the IDs? In other words, are they being used for anything? If we load the same apb twice, that's the user's doing, but will it break anything?

shawn-hurley:
I agree it won't break anything that we are doing unless we make it break. 

[...]

PR #257, opened 2017-07-03
[WIP] initial attempt at multiple registries.

shawn-hurley:
This is for issue #256 It is the first bit of changes called out in the proposal.

### DIFF
--- a/pkg/apb/registry/registry.go
+++ b/pkg/apb/registry/registry.go
@@ -1,7 +1,12 @@
 package registry
 
 import (
+	b64 "encoding/base64"
+	"encoding/json"
 	"fmt"
+	"net/http"
+
+	yaml "gopkg.in/yaml.v1"
 
 	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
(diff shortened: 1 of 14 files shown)


pkg/apb/registry/registry.go
```diff
@@ -0,0 +1,133 @@
...
+	err := reg.Init(config, log)
+	if err != nil {
+		return nil, err
+	}
+
+	return reg, err
+}
+
+// Retrieve the spec from a registry manifest request
+func imageToSpec(log *logging.Logger, req *http.Request) (*apb.Spec, error) {
```
dymurray: +1 to moving this to a common file.

pkg/broker/broker.go
```diff
@@ -185,14 +186,37 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 		a.log.Error("Something went real bad trying to delete batch specs... - %v", err)
 		return nil, err
 	}
+	specs = []*apb.Spec{}
 
-	if specs, imageCount, err = a.registry.LoadSpecs(); err != nil {
-		return nil, err
+	//Load Specs for each registry
+	registryErrors := []error{}
+	for _, r := range a.registry {
```
dymurray: Any thoughts on making this a channel? This could take a long time depending on which registries are configured.

shawn-hurley: I agree it could. I think the idea is we can go back and do this, while the way it is implemented now is easier and more readable for the first pass at the feature.

shawn-hurley:
This is going to moved to the feature branch.

PR #258, opened 2017-07-05
1467905 - Added error handling for images with improper APB spec

dymurray:


jmontleon:
ACK.

pkg/apb/rhcc_registry.go
```diff
@@ -66,10 +67,10 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
 	r.log.Debug("Found %d images in RHCC", numResults)
 	for _, image := range imageList.Results {
 		spec, err := r.imageToSpec(image)
-		if err != nil {
-			return []*Spec{}, 0, err
+		// Want to ignore errors here because some images will not be an APB
+		if err == nil {
+			specs = append(specs, spec)
 		}
```
jmrodri: I would change this to be:

```
if spec, ok := r.imageToSpec(image); ok {
    specs = append(specs, spec)
}
```

pkg/apb/rhcc_registry.go
```diff
@@ -117,8 +118,8 @@ func (r RHCCRegistry) imageToSpec(image *Image) (*Spec, error) {
 	}
 
 	if hist.History == nil {
-		r.log.Error("V1 Schema Manifest does not exist in registry")
-		return nil, nil
+		r.log.Error("V1 Schema Manifest history does not exist in registry")
+		return nil, errors.New("Error: Image history does not exist")
```
jmrodri: +1

pkg/apb/rhcc_registry.go
```diff
@@ -127,7 +128,17 @@ func (r RHCCRegistry) imageToSpec(image *Image) (*Spec, error) {
...
+		r.log.Error("Did not find v1 Manifest in image history. Skipping.")
+		return nil, errors.New("Error: v1 Manifest does not exist for this image")
+	}
+
 	encodedSpec := conf.Config.Label.Spec
+	if len(encodedSpec) == 0 {
+		r.log.Error("Didn't find encoded Spec label. Assuming image is not APB and skipping.")
+		return nil, errors.New("Error: Spec not found")
+	}
+
```
jmrodri: +1

pkg/apb/types.go
```diff
@@ -114,6 +114,9 @@ func specsLogDump(specs []*Spec, log *logging.Logger) {
 func NewSpecManifest(specs []*Spec) SpecManifest {
 	manifest := make(map[string]*Spec)
 	for _, spec := range specs {
+		if spec == nil {
+			return nil
+		}
```
jmrodri: 👍

pkg/apb/rhcc_registry.go
dymurray: This gives me an error because `ok` is a non-bool (type error). I have what you put here but instead kept it as `err` (since ok being nil implies not ok) and said err == nil.

jmrodri:
Are you sure? this doesn't compile for me:

```
$ git checkout -b dymurray-bz1467905 release-0.9
$ git pull git://github.com/dymurray/ansible-service-broker.git bz1467905
$ make clean
$ make build
go build -ldflags="-s -w" ./cmd/broker
# github.com/openshift/ansible-service-broker/pkg/apb
pkg/apb/rhcc_registry.go:70: syntax error: unexpected =, expecting comma or }
make: *** [Makefile:13: build] Error 2
```

pkg/apb/rhcc_registry.go
```diff
@@ -57,19 +58,17 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
 		return []*Spec{}, 0, err
 	}
 
 	numResults := imageList.NumResults
 	r.log.Debug("Found %d images in RHCC", numResults)
 	for _, image := range imageList.Results {
-		spec, err := r.imageToSpec(image)
-		if err != nil {
-			return []*Spec{}, 0, err
+		spec, ok := r.imageToSpec(image); ok {
```
jmrodri: missing an if statement. I apologize if this is what I gave you :(

rthallisey: ok isn't a bool.  ```if spec, ok := r.imageToSpec(image); ok != nil```

pkg/apb/rhcc_registry.go
```diff
@@ -127,7 +126,17 @@ func (r RHCCRegistry) imageToSpec(image *Image) (*Spec, error) {
 		return nil, err
 	}
 
+	if conf.Config == nil {
+		r.log.Error("Did not find v1 Manifest in image history. Skipping.")
+		return nil, errors.New("Error: v1 Manifest does not exist for this image")
```
rthallisey: The error you're returning isn't being used by the caller. Since the apb is getting skipped, I think it should be ```return nil, nil```

pkg/apb/rhcc_registry.go
```diff
@@ -57,19 +58,17 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
 	}
 
 	numResults := imageList.NumResults
 	r.log.Debug("Found %d images in RHCC", numResults)
 	for _, image := range imageList.Results {
-		spec, err := r.imageToSpec(image)
-		if err != nil {
-			return []*Spec{}, 0, err
+		if spec, err := r.imageToSpec(image); err == nil {
+			specs = append(specs, spec)
```
rthallisey: Since we're adding valid images and skipping invalid ones, should the log that identifies an invalid image be ```log.Info``` instead of ```log.Error```? I think it's great we identify there's an issue with an apb, but it's not a broker error.

pkg/apb/rhcc_registry_test.go
```diff
@@ -0,0 +1,130 @@
...
+            "alg": "ES256"
+         },
+         "signature": "4DuXenHioqVKeqdaqqYHmflj1zJu0ZSlRbfYv8xlFahVP_ZllFBpDjU8CX4DHNGov4BoEuXLfRqqbFlpN4NzFw",
+         "protected": "eyJmb3JtYXRMZW5ndGgiOjkyMzQsImZvcm1hdFRhaWwiOiJDbjAiLCJ0aW1lIjoiMjAxNy0wNi0yMFQwNTowNzoxOVoifQ"
+      }
+   ]
+}
+`
+
+func TestLoadSpecs(t *testing.T) {
```
shawn-hurley: :+1:

pkg/apb/rhcc_registry.go
rthallisey: It doesn't look like there are any fatal errors in this function. I think it's a legitimate assumption that if there is an invalid apb  we 1) alert the user of the problem 2) skip loading that apb. With that assumption, we should only have to log errors, but not return any. So you can rid of error as a return. 

Let me know if folks agree with that assumption.

pkg/apb/rhcc_registry_test.go
```diff
@@ -0,0 +1,130 @@
...
+	var log = &logging.Logger{}
+	config := RegistryConfig{Name: "rhcc", Url: serv.URL}
+	reg, err := NewRegistry(config, log)
+	specs, num, err := reg.LoadSpecs()
+	ft.AssertEqual(t, num, 3)
+	ft.AssertNotNil(t, specs)
+	if err != nil {
+		t.Fatal("ERROR: ", err)
+	}
+
```
jmrodri: 🎆 👍 thanks for adding the test.

pkg/apb/rhcc_registry.go
dymurray: Good point Ryan. I will update.

pkg/apb/rhcc_registry.go
```diff
@@ -113,34 +111,44 @@ func (r RHCCRegistry) imageToSpec(image *Image) (*Spec, error) {
...
 	if err != nil {
 		r.log.Error("Error grabbing JSON body from response: %s", err)
-		return nil, err
+		return nil
 	}
 
 	if hist.History == nil {
-		r.log.Error("V1 Schema Manifest does not exist in registry")
-		return nil, nil
+		r.log.Error("V1 Schema Manifest history does not exist in registry")
```
rthallisey: Change ```r.log.Errorr``` to ```r.log.Info``` for all these checks.

dymurray: @rthallisey Done. Thanks.

pkg/apb/rhcc_registry.go
```diff
@@ -57,39 +57,37 @@ func (r RHCCRegistry) LoadSpecs() ([]*Spec, int, error) {
...
+			specs = append(specs, spec)
 		}
-		specs = append(specs, spec)
 	}
 
 	return specs, numResults, nil
 }
 
-func (r RHCCRegistry) imageToSpec(image *Image) (*Spec, error) {
+func (r RHCCRegistry) imageToSpec(image *Image) (*Spec) {
```
rthallisey: I think travis doesn't like you have parenthesis around *Spec in the return statement.

shawn-hurley: should be ```func (r RHCCRegistry) imageToSpec(image *Image) *Spec```

shawn-hurley: format error that is causing the build failure

PR #262, opened 2017-07-11
Asbcli bind

rthallisey:
I did a little hacking on asbcli so I figured I'd share this.


eriknelson:
Thanks for this @rthallisey, I had a couple conversations with people today who were looking for ways to test running apbs outside the catalog but still exercise the broker, asbcli continues to be useful in this space.

I tried briefly to provision something today and wasn't able to. May try to submit something to fix that and keep this thing alive.

PR #263, opened 2017-07-11
1467852 - add ENV HOME to Dockerfile

jmontleon:


jmontleon:
@eriknelson Yes. Should I do a separate PR or will it get merged forward?

eriknelson:
@jmontleon yeah I think we just need another PR against master and maybe reference this one.

PR #264, opened 2017-07-11
1469401 - Print out error messages only once

rthallisey:
We are printing error messages and the help menu
twice. Also, clean up some unused options.

Closes-bz1469401

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -33,10 +33,7 @@ func CreateApp() App {
 
 	// Writing directly to stderr because log has not been bootstrapped
 	if app.args, err = CreateArgs(); err != nil {
-		os.Stderr.WriteString("ERROR: Failed to validate input\n")
-		os.Stderr.WriteString(err.Error() + "\n")
-		ArgsUsage()
-		os.Exit(127)
+		os.Exit(1)
 	}
 
 	if app.args.Version {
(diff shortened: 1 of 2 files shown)


pkg/app/args.go
```diff
@@ -31,13 +30,8 @@ func CreateArgs() (Args, error) {
 func validateArgs(args *Args) error {
 	var err error
 	if args.ConfigFile == "" {
```
shawn-hurley: We have a default for the argument, this should never error out, I would rather remove this to avoid confusion IMO.

pkg/app/app.go
```diff
@@ -33,9 +33,6 @@ func CreateApp() App {
 
 	// Writing directly to stderr because log has not been bootstrapped
 	if app.args, err = CreateArgs(); err != nil {
-		os.Stderr.WriteString("ERROR: Failed to validate input\n")
-		os.Stderr.WriteString(err.Error() + "\n")
-		ArgsUsage()
 		os.Exit(127)
```
shawn-hurley: I would also worry about returning 127 for the help option. I would think that we would want the more general 1 at least.

pkg/app/args.go
rthallisey: Yup this never gets used. Turns out we never even get to this code if there's any error in the args.

pkg/app/app.go
rthallisey: Sure. 1 wfm.

shawn-hurley:
ACK

eriknelson:
@rthallisey Can you please open a PR to merge this to master as well?

PR #268, opened 2017-07-12
Fix BZ1467852 add ENV HOME to Dockerfile (#263)

jmontleon:
1467852 - add ENV HOME to Dockerfile


jmontleon:
@fabianvf are they writing a kube config? If not it might not be necessary. Thus far I haven't seen them fail, not to say there isn't necessarily a problem.

PR #271, opened 2017-07-13
White/Black List Filtering and Multiple Registries Refactor

shawn-hurley:


etc/example-config.yaml
```diff
@@ -23,3 +29,5 @@ broker:
   launch_apb_on_bind: false
   recovery: true
   output_request: true
+  whitelist: "/etc/ansible-service-broker/whitelist.yaml"
```
shawn-hurley: @eriknelson I think this needs to be moved to the registry here.

pkg/apb/types_test.go
```diff
@@ -20,6 +20,7 @@ const SpecImage = "fusor/etherpad-apb"
 const SpecBindable = false
 const SpecAsync = "optional"
 const SpecDescription = "A note taking webapp"
+const SpecRegistryName = "test"
```
shawn-hurley: This should be removed missed it in the cleanup.

eriknelson:
This is obviously pretty big, I'm going to try to follow this up with a test plan and try to point out the stuff that needs review attention. It's a little misleading in +/- because reorg'd a bunch of files into different packages.

pkg/registries/registry.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if len(validNames) != 0 {
+		r.log.Debugf("Passing APBs:", r.config.Name)
+		for _, name := range validNames {
+			r.log.Debugf("-> %s", name)
+		}
+	}
+
+	if len(filteredNames) != 0 {
+		r.log.Info("Filtered APBs:")
```
shawn-hurley: Some of the images might not be APB's?

pkg/registries/registry.go
```diff
@@ -0,0 +1,166 @@
...
+			r.config.Name, err)
+		return []*apb.Spec{}, 0, err
+	}
+	validNames, filteredNames := r.filter.Run(imageNames)
+
+	r.log.Debug("Filter applied against registry: %s", r.config.Name)
+
+	if len(validNames) != 0 {
+		r.log.Debugf("Passing APBs:", r.config.Name)
+		for _, name := range validNames {
```
shawn-hurley: What about putting in a goroutine. We could do the whole loop off by itself with little effort and would not block if slice gets big

pkg/registries/registry.go
```diff
@@ -0,0 +1,166 @@
...
+	specs, err := r.adapter.FetchSpecs(validNames)
+	if err != nil {
+		r.log.Errorf("unable to fetch specs for registry %v - %v",
+			r.config.Name, err)
+		return []*apb.Spec{}, 0, err
+	}
+	return specs, len(imageNames), nil
+}
+
+// Fail - should this registry and error cause a failure.
```
rthallisey: Can you rephrase this comment

pkg/broker/broker.go
```diff
@@ -185,18 +189,51 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	}
+	if err := a.dao.BatchSetSpecs(specManifest); err != nil {
 		return nil, err
 	}
+	a.log.Debugf("specs -> %v", specs)
 
 	return &BootstrapResponse{SpecCount: len(specs), ImageCount: imageCount}, nil
 }
 
+func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
```
rthallisey: I think lint may yell at you for not adding a comment before this function

etc/example-config.yaml
eriknelson: Ah, missed that one.

pkg/apb/types_test.go
eriknelson: +1

pkg/registries/registry.go
eriknelson: I don't want this loop's logs to get interlaced with other logs running concurrently. What do you think about building up a single string with line breaks inside a goroutine and outputting once, so you get the full list displayed atomically?

pkg/registries/registry.go
eriknelson: It feels very wrong to me for the adapter to return a list of image names for filtering that contain images that are *not* APBs. I suppose it could happen, and the adapter's LoadSpecs will validate the labels and metadata further, but if `GetImageNames` is returning non-APBs, something is going awry.

docs/apb-filter-design.md
```diff
@@ -0,0 +1,80 @@
+# APB Filter Design Doc
+
```
eriknelson: This is distinctly a "Spec" or a "Design" doc that we started doing this sprint for the major features. The intent is to have a plan that has been vetted and reviewed before implementation. It was a bit awkward because we were still figuring out how we wanted to accomplish this, but its the same type of document as the one @shawn-hurley sent out for multi-reg and @jmrodri sent out for auth. I think we're trying to settle on posting them as PRs against his repo moving forwards, the auth one being a good reference.

I do still need to document white/blacklist usage in the relevant config doc though.

docs/apb-filter-design.md
```diff
@@ -0,0 +1,80 @@
...
+If only whitelist is present, strictest mode of operation. Any APB that does *not*
+match on a regex in the whitelist will be filtered.
+
+If only blacklist is present, all discovered APBs will be eligible *unless*
+they match on a regex found in the blacklist.
+
+If both are present, same behavior as whitelist, but if a value matches on white
+and black, blacklist will take precedence and block the whitelist match.
+
+These can be easily configured by cluster operators via config maps.
```
eriknelson: This is not intended to be considered operator documentation, I need to follow this up with that. It was mostly scratch design thoughts prior to writing any code.

pkg/registries/registry.go
shawn-hurley: It feels that wrong then we need to discuss how to change the dockerhub adapter. Currently, the dockerhub adapter will get all the images in an org. This does not know anything about APB's because we can not check the label at this time. We could do something like if the name does not contain "apb" then do not return it. I am having trouble coming up with another solution that does not make another network request. I also think that this might want to be an issue/card that we prioritize for next sprint.

pkg/broker/broker.go
shawn-hurley: this is an unexported method, so it wont complain. Do you think there needs to be a comment?

pkg/registries/registry.go
eriknelson: I'm realizing my initial assumption was wrong, but I still think the set of image names that hits the filter should *ideally* be only APBs. I originally thought the adapter was pushing the query serverside (GET /<whatever>?name=*-apb) or something to that effect, but that's actually not how its *ever* worked. It has always fetched every name under an org, and we filtered on our side based on *-apb. It was always a limitation of their API, from the work we did back in CAP days.

My thoughts on an ideal scenario:
1) Adapters can query a service and get back *only APBs*
2) Registry is able to validate this set and confirm that the names are indeed apbs
3) Vetted set of names hits filter
4) By this point, we're confident we have a set of images we know we care about, and we load the specs. (Possibly validate additional labels like apb.spec version).

Obviously we need to make some concessions because:
1) The service a particular adapter is talking to for fetching APB names may or may not support server-side filtering. In Dockerhub's case, it definitely does not.
2) Since we're strictly dealing with names, we don't have a lot of options for determining whether or not an image is actually an APB. Until now, it has always been `-apb` suffix.

I'd propose `Registry.LoadSpecs` to  filter any image names that do not match `-apb$` before they hit `Filter.Run`. It's a filter that applies to all registries, so it belongs at the Registry level and not in the adapters. Potential drawback here is possible apbs will be filtered if they don't follow this naming convention, but that has always been the case until now, and to your point, we should put a card on the backlog to think about how we can accomplish this better.

shawn-hurley: Ok, I will work on this now

pkg/registries/registry.go
```diff
@@ -0,0 +1,183 @@
...
+	specs, err := r.adapter.FetchSpecs(validNames)
+	if err != nil {
+		r.log.Errorf("unable to fetch specs for registry %v - %v",
+			r.config.Name, err)
+		return []*apb.Spec{}, 0, err
+	}
+	return specs, len(imageNames), nil
+}
+
+func registryFilterImagesForAPBs(imageNames []string) []string {
```
shawn-hurley: @eriknelson can you make sure this is what you were thinking?

eriknelson: @shawn-hurley perfect, thanks @shawn-hurley. We should discuss your thoughts more too, think you're right; there's a better way to handle this.

pkg/app/config.go
```diff
@@ -44,14 +46,24 @@ func CreateConfig(configFile string) (Config, error) {
...
+	registryName := map[string]bool{}
+	for _, rc := range c.Registry {
+		if !rc.Validate() {
+			return fmt.Errorf("registry config is not valid - %v", rc.Name)
+		}
+		if _, ok := registryName[rc.Name]; ok {
+			return fmt.Errorf("registry name must be unique")
+		}
+		registryName[rc.Name] = true
+	}
```
jmrodri: good validation.

pkg/broker/broker.go
```diff
@@ -186,18 +190,53 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
 		return nil, err
 	}
+	specs = []*apb.Spec{}
 
-	if specs, imageCount, err = a.registry.LoadSpecs(); err != nil {
-		return nil, err
+	//Load Specs for each registry
+	registryErrors := []error{}
+	for _, r := range a.registry {
+		s, i, err := r.LoadSpecs()
```
jmrodri: I would've used `c` or `cnt` for count instead of `i`. I thought the i was an index but based on the imageCount it's a total.

pkg/broker/types.go
```diff
@@ -71,7 +71,7 @@ type CatalogResponse struct {
 // LastOperationRequest - Request to obtain state information about an action that was taken
 // Defined in more detail here https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#polling-last-operation
 type LastOperationRequest struct {
-	ServiceID uuid.UUID `json:"service_id"`
```
jmrodri: I thought they were still UUIDs?

pkg/fusortest/util.go
```diff
@@ -94,6 +94,23 @@ func AssertNotNil(t *testing.T, a interface{}, message ...string) {
 	t.Fatal(msg)
 }
 
+// AssertNil - Assert input is nil
+func AssertNil(t *testing.T, a interface{}, message ...string) {
```
jmrodri: Thanks I was lazy and kept using `ft.AssertTrue(t, obj == nil, "msg")`. +1 to `AssertNil`

pkg/handler/handler.go
```diff
@@ -290,7 +290,7 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
 
 	// service_id is optional
 	if val, ok := queryparams["service_id"]; ok {
-		req.ServiceID = uuid.Parse(val[0])
+		req.ServiceID = val[0]
```
jmrodri: so how do we guarantee this isn't empty string? maybe we don't care.

pkg/registries/adapters/adapter.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/http"
+	"net/url"
+
+	logging "github.com/op/go-logging"
+	"github.com/openshift/ansible-service-broker/pkg/apb"
+	"github.com/pborman/uuid"
+	yaml "gopkg.in/yaml.v1"
+)
+
+//Adapter - Adapter will wrap the methods that a registry needs to fully manage images.
```
jmrodri: Space after the `//`.

pkg/registries/adapters/adapter.go
```diff
@@ -0,0 +1,108 @@
...
+	//FetchSpecs will retrieve all the specs for the list of images names.
+	FetchSpecs([]string) ([]*apb.Spec, error)
+}
+
+// BundleSpecLabel - label on the image that we should use to pull out the abp spec.
+// TODO: needs to remain ansibleapp UNTIL we redo the apps in dockerhub
+const BundleSpecLabel = "com.redhat.apb.spec"
+
+// Configuration - Adapter configuration. Contains the info that the adapter
+// would need to complete it's request to the images.
```
jmrodri: TYPO: it's => its

pkg/registries/filter.go
```diff
@@ -0,0 +1,227 @@
+package registries
+
+import (
+	"regexp"
+	"sync"
+)
+
+type filterMode uint8
```
jmrodri: what is this C? 😆

pkg/registries/filter.go
```diff
@@ -0,0 +1,227 @@
...
+	// Join workers and close chunks channel when finished
+	go func() {
+		wg.Wait()
+		close(matchChunksChan)
+	}()
+
+	return reduceChunks(matchChunksChan)
+}
+
+func reduceChunks(chunks <-chan []string) <-chan matchSetT {
```
jmrodri: cool, feels very CompSci-y

pkg/registries/filter.go
```diff
@@ -0,0 +1,227 @@
...
+	// Fan-in chunks, reduce to single set of matches
+	out := make(chan matchSetT)
+	go func() {
+		matchSet := make(matchSetT)
+		// Read and process chunks as they complete until input chan is cloesd
```
jmrodri: TYPO: cloesd -> closed

pkg/registries/filter_test.go
```diff
@@ -0,0 +1,209 @@
...
+	yaml "gopkg.in/yaml.v1"
+
+	ft "github.com/openshift/ansible-service-broker/pkg/fusortest"
+)
+
+const testWhitelistFile = "whitelist.yaml"
+const testBlacklistFile = "blacklist.yaml"
+const testBlacklistOverrideFile = "blacklist_override.yaml"
+
+func testGetRegexFromFile(file string) []string {
```
jmrodri: Does this actually run? I always thought they had to be `TestXYZ`. I learned something new today.

pkg/registries/registry.go
```diff
@@ -0,0 +1,183 @@
...
+var regex = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)
+
+// Config - Configuration for the registry
+type Config struct {
+	URL  string
+	User string
+	Pass string
+	Org  string
+	Type string
+	Name string
```
jmrodri: do the optional fields need to be marked `omitempty`?

pkg/registries/registry.go
jmrodri: +1 when I ask the adapter for the list of images, I expect the list of APBs. I understand if the adapter needs to get ALL of them because of an api limitation but it should filter it down by name first.

pkg/broker/types.go
shawn-hurley: ServiceID is now the hash, and they will not conform with UUID. The instance/job and other ID's will still be generated UUIDs

docs/apb-filter-design.md
jmrodri: I'm conflicted because I think the spec should ultimately reflect the implementation as was committed. If we go with the PR and specifications directory approach, we can have a history of the conversation. Seeing how it began and how it ultimately ended.

I was planning on updating my auth spec to reflect the implementation. Then leave it be.

pkg/registries/registry.go
shawn-hurley: We are never marshaling registry to JSON so I don't see a need. does anyone else have thoughts?

docs/apb-filter-design.md
eriknelson: @jmrodri @rthallisey so what action should I take based on this? Does this document need to get updated, or is it okay to leave it reflective of design thoughts *before* the implementation and update separate official usage documentation for operators?

docs/apb-filter-design.md
```diff
@@ -0,0 +1,80 @@
...
+
+## High Level Summary
+
+Introduce 2 new fields to broker section of config:
+
+`whitelist: "/etc/ansible-service-broker/whitelist.yaml"`
+`blacklist: "/etc/ansible-service-broker/blacklist.yaml"`
+
+Both are optional file paths that are yaml files containing arrays
+of regexes. Broker will construct the set of eligible APBs from this list.
```
eriknelson: This is part of the problem I mentioned to you. These aren't files anymore, they're arrays in the main configuration file on registries. What should I do now that implementation has changed since spec design?

pkg/registries/filter_test.go
eriknelson: Believe it's a utility function, not a test. Think it deserves a name change?

pkg/handler/handler.go
shawn-hurley: Honestly, I don't think that we are even using it currently in the broker except as a log statement also we allow the operation field above to be an empty string. do you feel strongly about this one

eriknelson:
Spoke with @rthallisey and @jmrodri re: `apb-filter-design.md`. Ultimately, it was supposed to be an implementation proposal that didn't belong in `docs/` and fell very much out of date with that actual implementation. Plan is to remove the file since it's entirely inaccurate and follow up this PR with proper usage documentation.

pkg/registries/filter.go
eriknelson: Fixed

docs/config.md
```diff
@@ -28,14 +28,64 @@ registry:
...
+
+### RHCC Registry
+```yaml
+registry:
+  - name: rhcc
+    type: rhcc
+    url: <rhcc url>
+```
+
+### Multiple Registries Example
```
shawn-hurley: Fixed.

docs/config.md
```diff
@@ -28,14 +28,64 @@ registry:
...
 name of the registry to 'mock' to use the Mock registry.
 
 ```
 registry:
-  name: mock
+  - name: mock
+    type: mock
+```
+
+### Dockerhub Registry
```
shawn-hurley: Fixed

pkg/registries/adapters/rhcc_adapter_test.go
```diff
@@ -118,13 +120,17 @@ func TestLoadSpecs(t *testing.T) {
...
-	ft.AssertEqual(t, num, 3)
-	ft.AssertNotNil(t, specs)
+	u, err := url.Parse(serv.URL)
+	if err != nil {
+		t.Fatal("ERROR: ", err)
+	}
+	config := Configuration{URL: u}
+	adapter := RHCCAdapter{Config: config, Log: log}
+	imageNames, err := adapter.GetImageNames()
+	//specs, num, err := reg.LoadSpecs()
```
rthallisey: Looks like a left over comment

pkg/registries/registry.go
```diff
@@ -0,0 +1,183 @@
...
+			r.config.Name, err)
+		return []*apb.Spec{}, 0, err
+	}
+	return specs, len(imageNames), nil
+}
+
+func registryFilterImagesForAPBs(imageNames []string) []string {
+	newNames := []string{}
+	for _, imagesName := range imageNames {
+		if strings.HasSuffix(strings.ToLower(imagesName), "-apb") {
```
rthallisey: This filters for the *-apb tag in the name. In the past we're we reading labels for an image? Do we document that '-apb' is required per image?

pkg/registries/filter.go
```diff
@@ -0,0 +1,227 @@
...
+type filterMode uint8
+
+const (
+	filterModeBoth filterMode = iota
+	filterModeWhite
+	filterModeBlack
+	filterModeNone
+)
+
+type failedRegexp struct {
```
rthallisey: Style nit: Make a type.go for all the structs and consts.

rthallisey:
@eriknelson ack from me once my comments are addressed

pkg/registries/filter.go
eriknelson: I actually am not a huge fan of this, but I could be swayed. @shawn-hurley what's most golang-like, in your opinion?

pkg/registries/registry.go
eriknelson: The Dockerhub adapter has always pulled the entire population of image names from the API and processed them broker-side in this way (due to a limitation of the API), then loaded the metadata from a separate endpoint and confirmed the label. We had a discussion about this elsewhere on the PR I'll try to link, but this was the conclusion until we can address this directly.

+1 re: documentation, that's a must.

shawn-hurley: This should be followed up and documented here https://github.com/fusor/ansible-playbook-bundle imo. @eriknelson @jmrodri @rthallisey  thoughts?

pkg/registries/filter.go
shawn-hurley: Kubernetes seems to use the type file. the standard lib does not. a good example is the net/http package. https://golang.org/src/net/http/header.go. Usually, when talking about good style/idioms  I side with the standard library.

shawn-hurley:
[(8) Multiple Registry Adapters configured in one Broker Instance](https://trello.com/c/1KqTacLK/304-8-multiple-registry-adapters-configured-in-one-broker-instance)

shawn-hurley:
[(8) Multiple Registry Adapters configured in one Broker Instance](https://trello.com/c/1KqTacLK/304-8-multiple-registry-adapters-configured-in-one-broker-instance)

PR #273, opened 2017-07-13
1470860 - Remove broker project creation 

eriknelson:
If the project does not exist, broker will report and return error


rthallisey:
Is there a specific reason to remove this? Are we adding back in the future?

rthallisey:
Saw the email thread nevermind.

jmrodri:
Why is there no bugzilla number in the commit?

eriknelson:
@jmrodri meant to add that back in and forgot. Updated.

PR #275, opened 2017-07-14
1470860 - Remove broker project creation

eriknelson:
Forward port of fix from release-0.9 #273

### DIFF
--- a/pkg/apb/provision.go
+++ b/pkg/apb/provision.go
@@ -41,18 +41,8 @@ func Provision(
 	ns := instance.Context.Namespace
 	log.Info("Checking if project %s exists...", ns)
 	if !projectExists(ns) {
-		log.Info("Project %s does NOT exist, creating project...", ns)
-		output, err := createProject(ns)
-		if err != nil {
-			log.Error("Something went wrong creating project %s!", ns)
-			log.Error(err.Error())
-			return "", nil, err
-		}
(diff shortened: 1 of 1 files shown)


pkg/apb/provision.go
```diff
@@ -41,18 +41,8 @@ func Provision(
...
-			log.Error(err.Error())
-			return "", nil, err
-		}
-		log.Info("Successfully created project %s", ns)
-		log.Debug("oc new-project output:")
-		log.Debug(string(output))
-	} else {
-		log.Info("Project %s already exists!", ns)
+		log.Error("Project %s does NOT exist! Cannot provision requested %s", ns, instance.Spec.Name)
+		return "", nil, errors.New(fmt.Sprintf("Project %s does not exist", ns))
```
rthallisey: Travis doesn't like this line. ```log.Errorf``` instead of ```errors.New(fmt.Sprintf(..))```

jmrodri: @eriknelson @rthallisey meant replace this:
```golang
return "", nil, errors.New(fmt.Sprintf("Project %s does not exist", ns))
```
with 
```golang
return "", nil, fmt.Errorf("Project %s does not exist", ns)
```

eriknelson:
@rthallisey @jmrodri had no idea about `fmt.Errorf`. Good to know! Updated.

PR #279, opened 2017-07-14
Filtering documentation

eriknelson:


eriknelson:
Proper filtering documentation, following up implementation #271

Thank you @jmrodri for the table :+1:

PR #281, opened 2017-07-17
HTTPS for asb route

shawn-hurley:


README.md
```diff
@@ -75,7 +75,9 @@ configurations.
 * `make vendor`: Installs or updates the dependencies
 * `make build`: Builds the binary from source
 * `make install`: Installs the built binary.
-* `make run`: Runs the broker with the default profile, configured via `/etc/dev.config.yaml`
+* `make prepare-local-env`: will set up the local environemt to test and run the broker against a local deployment of [catasb](https://github.com/fusor/catasb)
+* `make run`: Runs the broker with the default profile, configured via `/etc/generated_local_development.yaml`
```
jmrodri: it's not `/etc/` it's just the `etc` in the project repo.

shawn-hurley:
Need to update catasb and merge requests together.

dymurray:
[(8) Create an Adapter Registry in Broker to Communicate with AtomicRegistry portion of RHCC (ISV Images)](https://trello.com/c/TuUsP9bL/303-8-create-an-adapter-registry-in-broker-to-communicate-with-atomicregistry-portion-of-rhcc-isv-images)

johnkim76:
[(5) Configure https with Broker's route.](https://trello.com/c/DngTXbRR/324-5-configure-https-with-brokers-route)

etc/example-config.yaml
```diff
@@ -35,3 +35,5 @@ broker:
   launch_apb_on_bind: false
   recovery: true
   output_request: true
+  ssl_cert_key: /path/to/key
```
rthallisey: Should we default this to ```""```? dev_broker is being set to true, which means this should get defaulted to an empty string.

scripts/run_local.sh
```diff
@@ -38,5 +38,14 @@ if [ -z "${BROKER_CONFIG}" ]; then
   exit 1
 fi
 
-echo "Running ${BROKER_CMD} --config ${BROKER_CONFIG}"
-${BROKER_CMD} --config ${BROKER_CONFIG}
+if [ "${BROKER_INSECURE}" == "true" ]; then
+  echo "${BROKER_INSECURE}"
```
rthallisey: What's the point of this? Why not do the check ```"${BROKER_INSECURE}" == "true"``` below?

etc/example-config.yaml
shawn-hurley: what does this have to do with the dev_broker? I would rather never show this as being able to accept an empty string as that could be confusing.

rthallisey: IIRC when we're using the broker in dev mode we should skip ssl and run insecure.  But it looks like you covered that with the insecure flag so really what goes here won't matter if we use that flag.

PR #290, opened 2017-07-19
add tls files to really-clean

jmrodri:


jmrodri:
# TEST

After you have a prepared local env you should have files in secrets directory:

```
ls /var/run/secrets/kubernetes.io/serviceaccount/
ca.crt  tls.crt  tls.key  token
```

Now run `make really-clean`, those files should be gone now. 
```
ls /var/run/secrets/kubernetes.io/serviceaccount/
```

`really-clean` is needed when you want to force `make run` to rerun the `prepare-local-env` for you.

PR #291, opened 2017-07-20
Broker CI with Travis

rthallisey:
Building a simple CI framework that uses catasb to create
a cluster for testing. The CI will provision mediawiki and postgresql,
then bind them.

depends-on: https://github.com/fusor/catasb/pull/99


.travis.yml
```diff
@@ -1,16 +1,44 @@
...
+  - sudo service docker restart
 script:
   - ./scripts/travis.sh format
   - ./scripts/travis.sh vet
   - ./scripts/travis.sh lint
   - ./scripts/travis.sh build
   - ./scripts/travis.sh test
+  - ./scripts/travis.sh ci
+  - sudo docker pull docker.io/openshift/origin:v3.6.0-rc.0
+  - oc cluster up --image=docker.io/openshift/origin --version=v3.6.0-rc.0 --service-catalog=true
```
jwmatthews: If you don't specify the routing_suffix then I think you will see problems with the ServiceCatalog talking to the Ansible Service Broker.  The routing suffix defaults to 127.0.0.1.nip.io

The controller-manager pod will then attempt to access something like asb-1338-openshift-ansible-service-broker.apps.127.0.0.1.nip.io which goes back to itself 127.0.0.1 and is never able to connect to the broker.

We use the IP of the docker bridge on the host typically.

It's also possible we might be able to change how the broker registers to the service catalog, maybe we can use the kube dns between namespaces...not sure  but this implies it's possible to resolve services between namespaces:  https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

Yet, also worth considering, even if the resolution works, there may be cases where namespaces are fully locked down and can not communicate at the SDN layer, isolated networks, so the route option perhaps is safest.

rthallisey: Thanks for the info John

rthallisey: I'm going to add back in the routing_suffix and try and get that working. It was failing to work before so I moved on, but it looks like we're going to need it.

jwmatthews:
Excellent work @rthallisey 
Great to have this.

ACK

rthallisey:
Quay is still struggling: https://status.quay.io/

rthallisey:
The registries have stabilized. Merging now that it's back to green.

PR #293, opened 2017-07-21
[Proposal]: New Bind and Unbind Workflow

shawn-hurley:
Detailing the responsibility of the broker and APB to complete the new workflow.

### DIFF
--- a/docs/specs/bind-unbind-workflow.md
+++ b/docs/specs/bind-unbind-workflow.md
@@ -14,6 +14,7 @@ NOTE: The launch_apb_on_bind will eventually be deprecated.
 * The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
 * The broker will be responsible for providing the provisioned credentials (extracted credentials given at provision time) to the APB as extra vars during bind and unbind. Example
   - `--extra-vars '{..., "_apb_provision_creds:{}`
+  - during unbind `{..., "_apb_provision_creds:{}, "_apb_bind_creds:{}}`
 * The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
 
 ### APB Responsibility
--- a/docs/specs/bind-unbind-workflow.md
+++ b/docs/specs/bind-unbind-workflow.md
@@ -7,18 +7,18 @@ While this is still using existing workflow pieces (i.e extracting credentials, 
 ### Default Settings Needed
 * set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
 
+NOTE: The launch_apb_on_bind will eventually be deprecated.
+
 
 ### Brokers Responsibility
 * The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
-* At provision time, the broker should inform the APB of the launch_apb_on_bind value. The apb can then choose what it wants to do. 
 * The broker will be responsible for providing the provisioned credentials (extracted credentials given at provision time) to the APB as extra vars during bind and unbind. Example
-  - `--extra-vars '{..., "_apb_db_params:{}`
+  - `--extra-vars '{..., "_apb_provision_creds:{}`
 * The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
-* The broker will be responsible for deleting the bindings extracted credentials in etcd during deprovision.
-
 
 ### APB Responsibility
-* The APB is responsible for encoding new or already created parameters into `/var/tmp/bind-creds`.
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
 * The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
 * The APB is responsible for handling the deletion/creation of users for their image.
 
--- a/docs/specs/bind-unbind-workflow.md
+++ b/docs/specs/bind-unbind-workflow.md
@@ -1,6 +1,8 @@
 ## Bind / Unbind New Work flow
 
-The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched and to run the bind or unbind actions. 
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched and to run the bind or unbind actions.
+
+While this is still using existing workflow pieces (i.e extracting credentials, saving those credentials), it is also creating a new workflow. This is an enhancement on the current design and will allow for the APB to choose which workflow it would prefer to follow. 
 
 ### Default Settings Needed
 * set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
@@ -8,11 +10,9 @@ The goal of this new work flow is to give more control over the actions that the
 
 ### Brokers Responsibility
 * The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* At provision time, the broker should inform the APB of the launch_apb_on_bind value. The apb can then choose what it wants to do. 
 * The broker will be responsible for providing the provisioned credentials (extracted credentials given at provision time) to the APB as extra vars during bind and unbind. Example
-* `--extra-vars '{"provision_params": {}, "bind_params":{}, "db_params:{}`
-  - Note: provision_params and bind_params are currently being defined in the APB and the service catalog is passing them to the broker, to be sent along to the APB.
-  - Example: on APB provision, the APB decides to create an admin user & password. This admin user should not be given out during binding. Instead the APB author would like to create new user accounts (by using the already saved/created admin user & password), and pass those back as credentials during the bind. The same is true for unbind. The APB author could want to delete the bind credentials and would need to use the admin user & password. 
-  - Example: APB author would like to keep only 1 account, and use that 1 account for all the bindings(current behavior). APB authors can make no guarantees on how the broker is running. (launch_apb_on_bind is true or false). That means their bind would need to pass back the credentials created at provision time. They can now do this by just taking the parameters passed in and sending them back to the broker. 
+  - `--extra-vars '{..., "_apb_db_params:{}`
 * The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
 * The broker will be responsible for deleting the bindings extracted credentials in etcd during deprovision.
 
@@ -21,3 +21,12 @@ The goal of this new work flow is to give more control over the actions that the
 * The APB is responsible for encoding new or already created parameters into `/var/tmp/bind-creds`.
 * The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
 * The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed): 
+1. On APB provision, the APB decides to create an admin user & password. 
+2. This admin user should not be given out during binding. Instead the APB author would like to create new user accounts (by using the already saved/created admin user & password), and pass those back as credentials during the bind. 
+3. The same is true for unbind. The APB author could want to delete the bind credentials and would need to use the admin user & password. 
+
+#### Example Usage: 
+1. APB author would like to keep only 1 account, and use that 1 account for all the bindings(current behavior).
+2. That means their bind would need to pass back the credentials created at provision time. They can now do this by just taking the parameters passed in and sending them back to the broker. 
--- a/docs/specs/bind-unbind-workflow.md
+++ b/docs/specs/bind-unbind-workflow.md
@@ -1,6 +1,6 @@
 ## Bind / Unbind New Work flow
 
-The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched on on bind and unbind. 
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched and to run the bind or unbind actions. 
 
 ### Default Settings Needed
 * set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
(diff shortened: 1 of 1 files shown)


shawn-hurley:
[Update Bind Operation in postgres APB so that 'bind' will create credentials and unbind will revoke](https://trello.com/c/iDQUZzpW/327-update-bind-operation-in-postgres-apb-so-that-bind-will-create-credentials-and-unbind-will-revoke)

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
+## Bind / Unbind New Work flow
+
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched on on bind and unbind.
```
rthallisey: ```on on bind```

rthallisey: This means we launch an APB on bind and run the bind playbook?

jmrodri: you have two "on"s

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
...
+
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched on on bind and unbind. 
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* The broker will be responsible for providing the provisioned credentials to the APB as extra vars during bind and unbind. Example
```
rthallisey: What are the provisioned credentials? Are those bind credentials created during a provision?

docs/specs/bind-unbind-workflow.md
jmrodri: Also what is the *new* workflow? because right now I thought we had a setting that if enabled will actually make a call to the apbs to do a bind (whatever that bind wants to do). If unset, we return what's in etcd probably from a previous provision call.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
+## Bind / Unbind New Work flow
+
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched on on bind and unbind. 
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
```
jmrodri: This was the setting I was referring to in my comment above.

docs/specs/bind-unbind-workflow.md
jmrodri: Isn't this what the broker already does?

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* The broker will be responsible for providing the provisioned credentials to the APB as extra vars during bind and unbind. Example
+```
+--extra-vars '{"provision_params": {}, "bind_params":{}, "db_params:{}'
```
rthallisey: Can you create a section explaining what each of these key value pairs are and where they come from?

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
...
+```
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during deprovision 
+
+
+### APB Responsibility
+* The APB is responsible for encoding new or old parameters into `/var/tmp/bind-creds`.
+* The APB is responsible for handling the use of superuser vs. regular user. 
+* The APB is responsible for handling the deletion/creation of users for their image.
```
rthallisey: Can you create a list of ```## Work Items``` so folks can track the work that needs to go into this?

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,22 @@
...
+The goal of this new work flow is to give more control over the actions that the database takes on bind and unbind to the APB author. This means that we want the apb to be launched on on bind and unbind. 
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* The broker will be responsible for providing the provisioned credentials to the APB as extra vars during bind and unbind. Example
+```
```
rthallisey: Why would bind and unbind need to receive params? Can you go into more detail about that?

docs/specs/bind-unbind-workflow.md
shawn-hurley: It does not provide the credentials. It provides provision and binding params, but that is not the extracted credentials IMO. Also, the broker does not ever say credentials for a binding, it only ever deals in the provisioned credentials. I think these make it a significant change to bind and unbind.

docs/specs/bind-unbind-workflow.md
shawn-hurley: The new part is that we have no way of telling the apb how to talk to the database. See comment below on what we need to change. I can change the name to update or enhance if you would like?

docs/specs/bind-unbind-workflow.md
shawn-hurley: I want this to be a spec that does not need to be updated with work/outstanding questions.  If you want I can create the work that needs to happen on the trello card?

docs/specs/bind-unbind-workflow.md
shawn-hurley: These keys are from the parameter spec that is passed between the service catalog and the broker. This is documented in the openservicebrokerapi imo.

docs/specs/bind-unbind-workflow.md
shawn-hurley: I attempted to explain this better in the update, let me know if it makes more sense and the difference is more clear.

docs/specs/bind-unbind-workflow.md
shawn-hurley: I added a note in the update to explain those parameters. let me know if that is sufficient

docs/specs/bind-unbind-workflow.md
shawn-hurley: I tried to create examples that would show what was needed. Let me know if that addressed your question of if it still needs clarification.

docs/specs/bind-unbind-workflow.md
shawn-hurley: Re-worded this line, let me know if it makes more sense or not please.

docs/specs/bind-unbind-workflow.md
rthallisey: Work Items is a list of a few high level items that will be required to implement this feature.  The goal is to layout all things that need to get done with a few words in list format.

For example:
 - Default luanch_apb_bind to true
 - Add a new field 'APB_dao_creds' to the dao object
 - Update APBs so they accept new params

shawn-hurley:
@jmrodri @rthallisey I updated the doc again.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+While this is still using existing workflow pieces (i.e extracting credentials, saving those credentials), it is also creating a new workflow. This is an enhancement on the current design and will allow for the APB to choose which workflow it would prefer to follow. 
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* At provision time, the broker should inform the APB of the launch_apb_on_bind value. The apb can then choose what it wants to do.
```
eriknelson: Why would an APB need to be told such? Doesn't it already know its own behavior for all actions? What is the use-case to be told ahead of time? It's not stateful.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* At provision time, the broker should inform the APB of the launch_apb_on_bind value. The apb can then choose what it wants to do. 
+* The broker will be responsible for providing the provisioned credentials (extracted credentials given at provision time) to the APB as extra vars during bind and unbind. Example
+  - `--extra-vars '{..., "_apb_db_params:{}`
```
eriknelson: Love the `_apb_` prefix, right inline with my thinking. Contents should be standard and generic, i.e. not `db_params`. I would prefer something like `_apb_provision_params`.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during deprovision.
```
eriknelson: Our concerns on this were raised in IRC. The broker should never receive requests to delete potentially referenced resources. I.E. Delete ServiceInstance with outstanding bindings. That sounds like a hard error to me. At least until the spec suggests otherwise.

If this occurs, 4xx. Bad request.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+### APB Responsibility
+* The APB is responsible for encoding new or already created parameters into `/var/tmp/bind-creds`.
```
eriknelson: I would really like to see us start using the bind module @cfchase wrote for this. We need to encapsulate this action.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
+
+### APB Responsibility
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
+* The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
+* The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed): 
+1. On APB provision, the APB decides to create an admin user & password.
```
rthallisey: ```the APB decides to create an admin user & password```. A user and pw in the cluster? Or a new user and pw in the app? Can you be more specific?  Try plugging mediawiki and postgresql into your example.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+### APB Responsibility
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
+* The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
+* The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed): 
+1. On APB provision, the APB decides to create an admin user & password. 
+2. This admin user should not be given out during binding. Instead the APB author would like to create new user accounts (by using the already saved/created admin user & password), and pass those back as credentials during the bind.
```
rthallisey: Can you be more specific about ```the admin user```? Is this relating to the cluster or the app? I guess I'm confused because both have users and the APB can touch both of then.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+### APB Responsibility
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
+* The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
+* The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed): 
+1. On APB provision, the APB decides to create an admin user & password. 
+2. This admin user should not be given out during binding. Instead the APB author would like to create new user accounts (by using the already saved/created admin user & password), and pass those back as credentials during the bind. 
+3. The same is true for unbind. The APB author could want to delete the bind credentials and would need to use the admin user & password.
```
rthallisey: This is my understanding after reading the examples: We want the APB to have the ability to:
 1) Add a new user & pw to an app
 2) Delete credentials from the broker

shawn-hurley: The APB should have the ability to manage the bindings that are requested of it is a better way of thinking about it. 

The broker should manage the deletion and creation of credentials from etcd for a binding on unbinding, not the APB. 

The APB could want to once again manage that unbind request of itself and delete the new user that the APB created for its application.

docs/specs/bind-unbind-workflow.md
shawn-hurley: updated to make more clear that we only talking about what ever the APB is managing.

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+  - `--extra-vars '{..., "_apb_provision_creds:{}`
+* The broker will be responsible for deleting the bindings extracted credentials in etcd during unbind.
+
+### APB Responsibility
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
+* The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
+* The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed):
```
rthallisey: > The APB should have the ability to manage the bindings that are requested of it is a better way of thinking about it.
>
_How can the APBs manage the bindings?_
The APB can tell the broker about certain users while keeping the admin user to itself. This is useful if ...

> The broker should manage the deletion and creation of credentials from etcd for a binding on unbinding, not the APB.
>
The broker's behaviour doesn't change.

>The APB could want to once again manage that unbind request of itself and delete the new user that the APB created for its application.
>
_What does this look like?_
The broker deletes the credentials from etcd. The APB deletes the user and password in the application.

The three sentences you wrote are a good way to explain it. Maybe we can fit these in somewhere? I added in some follow up sentences with questions I think we should answer.

shawn-hurley: Updated the document to include these exact wordings (as best as possible). I think that the broker is still changing quite a bit, (deleting credentials during unbind is a new thing, saving credentials for a binding id is a new thing).

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Default Settings Needed
+* set `launch_apb_on_bind: true` - If this is set to false we will behave the exact same as today, handing back the credentials received at provision to the service catalog.
+
+NOTE: The launch_apb_on_bind will eventually be deprecated.
+
+
+### Brokers Responsibility
+* The broker will save the credentials for each bind. this will be saved in etcd under the `extracted_credentials` directory. The key will be the binding id.
+* The broker will be responsible for providing the provisioned credentials (extracted credentials given at provision time) to the APB as extra vars during bind and unbind. Example
```
rthallisey: Line 15 is a change in the broker.
 - Pass provision credentials tto an APB duing a bind

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+
+### APB Responsibility
+* The APB is responsible for encoding credentails. Using the [asb_encode_binding](https://github.com/fusor/apb-examples/pull/93/files/3d444b778e27ac3fb266fc5cc55d55eee211fb50#diff-c0c3dd5820ea9b91bd5f865af6a41f67) ansible module.
+- Note: will be unused until PR is merged and we everything can use this method.
+* The APB is responsible for handling the use of superuser/admin user vs. regular user to manage their image. 
+* The APB is responsible for handling the deletion/creation of users for their image.
+
+#### Example Usage (workflow proposed): 
+1. On APB provision, the APB decides that, for what ever application it is responsible for deploying, to create an admin user & password. 
+2. The APB does not want to hand out the admin user and password, for the application that it is responsible for, to all the other applications that could be bound. Instead the APB author would like to create new user accounts (by using the already saved/created admin user & password), and pass those back as credentials during the bind. The new credentials are for the application that the APB is responsible for.
```
rthallisey: What would have to be added to get this feature? Is this change documenting what you can do in an APB? Or do we want to make a change in all our existing APBs?
 - Document how an APB can control user & password creation

docs/specs/bind-unbind-workflow.md
```diff
@@ -0,0 +1,32 @@
...
+3. The same is true for unbind. The APB author could want to delete the user and password, for the application that it manages, credentials and would need to use the admin user & password. The broker would alse be deleting credentials from etcd at this time.
```
rthallisey: On an unbind, the broker would delete the credentials for application x. Then the APB would delete the user & pw associated with those credentials. I think this means the broker needs to pass in the bind creds to the APB so it's aware of which user & pw to remove. Do you agree?
 - Broker passes APBs the bind creds on unbind then deletes credentials.

Are there any failure scenarios here we should consider?

jmrodri:
Sounds ok, the biggest problem with this will be the fact that we do not have async binds yet. This means we might not return within the time frame allotted by the service-catalog. Other than that, looks good.

Issue #297, opened 2017-07-22
spec logs spec ids instead of useful info


jmrodri:
```
[2017-07-22T17:24:13.402Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:13.595Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/jenkins-apb into Spec
[2017-07-22T17:24:13.63Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:13.837Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/pyzip-demo-db-apb into Spec
[2017-07-22T17:24:13.879Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:14.144Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/rhscl-postgresql-apb into Spec
[2017-07-22T17:24:14.182Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:14.336Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/pyzip-demo-apb into Spec
[2017-07-22T17:24:14.383Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:14.757Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/mediawiki123-apb into Spec
[2017-07-22T17:24:14.792Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:15.166Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/wordpress-ha-apb into Spec
[2017-07-22T17:24:15.204Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:15.339Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/hello-world-apb into Spec
[2017-07-22T17:24:15.377Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:15.576Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/rhscl-mariadb-apb into Spec
[2017-07-22T17:24:15.621Z] [DEBUG] Registry::imageToSpec
[2017-07-22T17:24:15.988Z] [DEBUG] Successfully converted Image ansibleplaybookbundle/rocketchat-apb into Spec
[2017-07-22T17:24:16.023Z] [DEBUG] Registry::imageToSpec
[...]

PR #298, opened 2017-07-24
Multi-plan support

eriknelson:


pkg/handler/handler.go
```diff
@@ -174,14 +174,22 @@ func (h handler) deprovision(w http.ResponseWriter, r *http.Request, params map[
 		writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: "invalid instance_uuid"})
 		return
 	}
+
 	var async bool
 	queryparams := r.URL.Query()
```
shawn-hurley: So small nit, just because your editing this.

I would prefer to use the [request.FormValue()](https://golang.org/pkg/net/http/#Request.FormValue). I personally think it is nicer and easier to use.

eriknelson:
Going to fix the ol' linter

pkg/registries/registry.go
```diff
@@ -181,3 +195,63 @@ func createFilter(config Config, log *logging.Logger) Filter {
 
 	return filter
 }
+
+func validateSpecs(log *logging.Logger, inSpecs []*apb.Spec) []*apb.Spec {
+	var wg sync.WaitGroup
+	wg.Add(len(inSpecs))
```
shawn-hurley: I prefer to call the addition to the wait group as I call the go routines. To me, it makes it more clear what we are supposed to be waiting on. This is still really clear, just throwing it out there for the future.

pkg/registries/registry.go
```diff
@@ -61,7 +62,7 @@ func (r Registry) LoadSpecs() ([]*apb.Spec, int, error) {
 	r.log.Debug("Filter applied against registry: %s", r.config.Name)
 
 	if len(validNames) != 0 {
-		r.log.Debugf("Passing APBs:")
+		r.log.Debugf("APBs passing white/blacklist filter:")
```
shawn-hurley: :+1:

pkg/registries/registry.go
```diff
@@ -181,3 +195,63 @@ func createFilter(config Config, log *logging.Logger) Filter {
...
+		spec       *apb.Spec
+		failReason string
+	}
+
+	out := make(chan resultT)
+	for _, spec := range inSpecs {
+		go func(s *apb.Spec) {
+			ok, failReason := validateSpecPlans(s)
+			out <- resultT{ok, s, failReason}
+			wg.Done()
```
shawn-hurley: You should defer this, things could break if an unexpected panic happens.

pkg/handler/handler.go
eriknelson: Much better, didn't know about this. TY @shawn-hurley

eriknelson:
Updated based on feedback.

eriknelson:
Realizing this requires some decent testing changes. Working on that.

pkg/handler/handler.go
eriknelson: Logged an issue to update the request layer, https://github.com/openshift/ansible-service-broker/issues/303

pkg/registries/registry.go
eriknelson: I think I like saying "I expect this to run for every spec" up front.

eriknelson: @shawn-hurley I'd like to also talk about this method. It's not in the scope of this PR, but where do we want to actually validate what we actually bring in, and how?

pkg/handler/handler.go
jmrodri: But only if they are coming in as a form. If they are json or coming in on the actual url hook. I'd be careful making that change blindly.

pkg/apb/types.go
```diff
@@ -39,9 +47,8 @@ type Spec struct {
 	Metadata    map[string]interface{} `json:"metadata,omitempty"`
 
 	// required, optional, unsupported
```
rthallisey: This comment looks out of place now.

pkg/broker/broker.go
```diff
@@ -459,14 +456,27 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
...
 	context := &req.Context
-	parameters := &req.Parameters
+	parameters := req.Parameters
+
+	if req.PlanID == "" {
+		errMsg :=
+			"PlanID from provision request is blank. " +
+				"Provision requests must specify PlanIDs"
+		a.log.Error(errMsg)
+		return nil, errors.New(errMsg)
```
rthallisey: This is going to print ```errMsg``` twice.  Here's where err gets passed to: https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/provision_job.go#L50. Same thing for all the custom error messages below.

Because we have the caller capturing and printing error everywhere, I think the only places it makes sense to print a custom error message is if you ```return nil, err```.  That way we print both the custom errMsg and the actually error.

rthallisey:
Closes-bug: https://github.com/openshift/ansible-service-broker/issues/297

pkg/handler/handler.go
eriknelson: @jmrodri Are you satisfied with this after your investigation?

pkg/registries/registry.go
eriknelson: @shawn-hurley coming back to this now that the deprovision fire is (hopefully) out. I think we're talking about how the broker should validate Specs that come in, which IMO is outside the scope of this PR. I would like to lodge a card to consider spec validation explicitly, because it's something we need to address. The broker needs to defend against adapters sending bunk Specs.

pkg/handler/handler.go
jmrodri: @eriknelson yes FormValue is ok. Weird but it seems to work.

PR #302, opened 2017-07-25
Adding ability to pass credentials to bind and unbind actions.

shawn-hurley:
Implementation for #293. Just getting thoughts down. This allows you test with the the test apps and APB  
https://github.com/shawn-hurley/rhscl-postgresql-apb
https://github.com/shawn-hurley/bottle-todo-demo

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -100,10 +100,12 @@ The broker config section will tell the broker what functionality should be enab
 and disabled. It will also tell the broker where to find files on disk that will
 enable the full functionality.
 
+*Note: with the absence of async bind, setting launch_apb_on_bind to true can cause the bind action to timeout and will span a retry. The broker will handle with with 409 Conflicts because it is the same bind request with different parameters.*
+
 **field**|**description**|**default value**|**required**
 :-----:|:-----:|:-----:|:-----:
 dev_broker|Allow development routes to be accessable|false|N
-launch_apb_on_bind|Allow bind be be no op|false|N
+launch_apb_on_bind|Allow bind be be no op|true|N
 bootstrap_on_startup|Allow the broker attempt to bootstrap itself on start up. Will retrieve the APBs from configured registries|false|N
 recovery|Allow the broker to attempt to recover itself by dealing with pending jobs noted in etcd|false|N
 output_request|Allow the broker to output the requests to the log file as they come in for easier debugging|false|N
--- a/docs/config.md
+++ b/docs/config.md
@@ -105,7 +105,7 @@ enable the full functionality.
 **field**|**description**|**default value**|**required**
 :-----:|:-----:|:-----:|:-----:
 dev_broker|Allow development routes to be accessable|false|N
-launch_apb_on_bind|Allow bind be be no op|true|N
+launch_apb_on_bind|Allow bind be be no op|false|N
 bootstrap_on_startup|Allow the broker attempt to bootstrap itself on start up. Will retrieve the APBs from configured registries|false|N
 recovery|Allow the broker to attempt to recover itself by dealing with pending jobs noted in etcd|false|N
 output_request|Allow the broker to output the requests to the log file as they come in for easier debugging|false|N
--- a/pkg/apb/unbind.go
+++ b/pkg/apb/unbind.go
@@ -10,7 +10,7 @@ import (
 // github.com/op/go-logging, which is used all over the broker
 // Maybe apb defines its own interface and accepts that optionally
 // Little looser, but still not great
-func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging.Logger, parameters *Parameters) error {
+func Unbind(instance *ServiceInstance, parameters *Parameters, clusterConfig ClusterConfig, log *logging.Logger) error {
 	log.Notice("============================================================")
 	log.Notice("                       UNBINDING                              ")
 	log.Notice("============================================================")
--- a/pkg/apb/unbind.go
+++ b/pkg/apb/unbind.go
@@ -10,7 +10,7 @@ import (
 // github.com/op/go-logging, which is used all over the broker
 // Maybe apb defines its own interface and accepts that optionally
 // Little looser, but still not great
-func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging.Logger) error {
+func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging.Logger, parameters *Parameters) error {
 	log.Notice("============================================================")
 	log.Notice("                       UNBINDING                              ")
 	log.Notice("============================================================")
@@ -18,7 +18,7 @@ func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging
 	// podName, err
 	_, err := ExecuteApb(
 		"unbind", clusterConfig, instance.Spec,
-		instance.Context, instance.Parameters, log,
+		instance.Context, parameters, log,
 	)
 
 	if err != nil {
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -782,6 +782,11 @@ func (a AnsibleBroker) Unbind(
 	}
 	// Add the credentials to the parameters so that an APB can choose what
 	// it would like to do.
+	if provExtCreds == nil && bindExtCreds == nil {
+		a.log.Warningf("Unable to find credentials for instance id: %v and binding id: %v"+
+			" something may have gone wrong. Proceeding with unbind.",
+			instanceUUID, bindingUUID)
+	}
 	if provExtCreds != nil {
 		params[provisionCredentialsKey] = provExtCreds.Credentials
 	}
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -717,7 +717,7 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
 		a.log.Warningf("unable to retrieve provision time credentials - %v", err)
 	}
 
-	//Add the DB Credentials this will allow the
+	// Add the DB Credentials this will allow the apb to use these credentials if it so chooses.
 	params[provisionCredentialsKey] = provExtCreds.Credentials
 
 	// NOTE: We are currently disabling running an APB on bind via 'LaunchApbOnBind'
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -734,7 +734,7 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
 			return nil, err
 		}
 	} else {
-		a.log.Warningf("Broker configured to *NOT* launch and run APB bind")
+		a.log.Warning("Broker configured to *NOT* launch and run APB bind")
 	}
 	instance.AddBinding(bindingUUID)
 	if err := a.dao.SetServiceInstance(instanceUUID.String(), instance); err != nil {
@@ -793,7 +793,7 @@ func (a AnsibleBroker) Unbind(
 	}
 	// only launch apb if we are always launching the APB.
 	if a.brokerConfig.LaunchApbOnBind {
-		err = apb.Unbind(serviceInstance, a.clusterConfig, a.log, &params)
+		err = apb.Unbind(serviceInstance, &params, a.clusterConfig, a.log)
 		if err != nil {
 			return nil, err
 		}
@@ -810,6 +810,7 @@ func (a AnsibleBroker) Unbind(
 	if err != nil {
 		return nil, err
 	}
+
 	serviceInstance.RemoveBinding(bindingUUID)
 	err = a.dao.SetServiceInstance(instanceUUID.String(), serviceInstance)
 	if err != nil {
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -30,8 +30,10 @@ var (
 )
 
 const (
+	// provisionCredentialsKey - Key used to pass credentials to apb.
 	provisionCredentialsKey = "_apb_provision_creds"
-	bindCredentialsKey      = "_apb_bind_creds"
+	// bindCredentialsKey - Key used to pas bind credentials to apb.
+	bindCredentialsKey = "_apb_bind_creds"
 )
 
 // Broker - A broker is used to to compelete all the tasks that a broker must be able to do.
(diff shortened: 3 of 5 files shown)


shawn-hurley:
## TESTING New Workflow
*Needs to be run on a machine that spins up containers quickly*

Use: 
APB -  https://github.com/shawn-hurley/rhscl-postgresql-apb
Application - https://github.com/shawn-hurley/bottle-todo-demo use S2I to build to pod.

The test APB will give you the ability to select which workflow you would like the APB to do.

The options are `multiple` or `single`. `multiple` will tell the APB to generate a new user for each bind. The credentials created at provision time should be for the postgres user. The `single` option will tell the APB to generate a new user, and then use that user for subsequent bind requests. 

### Test 1

Launch catasb point to broker built with the changes. Leave `launch_apb_on_bind` to false.
* Use the media-wiki and RHSCL postgres images to test that the original workflow still works. 
- deploy postgresql apb and media-wiki apb 
- should be able to bind mediawiki-apb and postgres-apb
- unbind should delete the secret and remove the pod preset


### Test 2

Launch catasb point to broker built with the changes. set `launch_apb_on_bind` to true.
* Use test postgres-apb and test application
- deploy S2I python application and test APB Postgresql. Select the `multiple` user option
- application should display no database connection. 
- Create a binding between application and postgresql.
- APB should launch.
- New generated user and password should be created in etcd /extracted_credentials directory with key of binding id.
- secret should be created with new user and password.
- pod preset should be created  with new user and password.
- redeploy application and reload application page. Then navigate to /todo
- Should see data, can update tasks. You can create new todo's by navigating to /new and creating one

### Test 3
Launch catasb point to broker built with the changes. set `launch_apb_on_bind` to true.
* Use test postgres-apb and test application
- deploy S2I python application and test APB Postgresql. Select the `single` user option
- application should display no database connection. 
- Create a binding between application and postgresql.
- APB should launch.
- same user and password should be created in etcd /extracted_credentials directory with key of binding id.
- secret should be created with new user and password.
- pod preset should be created  with new user and password.
- redeploy application and reload application page. Then navigate to /todo
- Should see data, can update tasks. You can create new todo's by navigating to /new and creating one



**PLEASE NOTE**: that if it takes too long for the APB to run the service catalog will fail to get the data and will not create the secret. This is due to lack of async binding support.

shawn-hurley:
[Update Bind Operation in postgres APB so that 'bind' will create credentials and unbind will revoke](https://trello.com/c/iDQUZzpW/327-update-bind-operation-in-postgres-apb-so-that-bind-will-create-credentials-and-unbind-will-revoke)

shawn-hurley:
[Update Bind Operation in postgres APB so that 'bind' will create credentials and unbind will revoke](https://trello.com/c/iDQUZzpW/327-update-bind-operation-in-postgres-apb-so-that-bind-will-create-credentials-and-unbind-will-revoke)

rthallisey:
LGTM. I haven't tested yet.

docs/config.md
```diff
@@ -100,6 +100,8 @@ The broker config section will tell the broker what functionality should be enab
 and disabled. It will also tell the broker where to find files on disk that will
 enable the full functionality.
 
+*Note: with the absence of async bind, setting launch_apb_on_bind to true can cause the bind action to timeout and will span a retry. The broker will handle with with 409 Conflicts because it is the same bind request with different parameters.*
```
jmrodri: Good note.

pkg/apb/unbind.go
```diff
@@ -10,15 +10,15 @@ import (
...
 	log.Notice("============================================================")
 	log.Notice("                       UNBINDING                              ")
 	log.Notice("============================================================")
 
 	// podName, err
 	_, err := ExecuteApb(
 		"unbind", clusterConfig, instance.Spec,
-		instance.Context, instance.Parameters, log,
+		instance.Context, parameters, log,
 	)
```
jmrodri: +1

pkg/apb/unbind.go
```diff
@@ -10,15 +10,15 @@ import (
 // github.com/op/go-logging, which is used all over the broker
 // Maybe apb defines its own interface and accepts that optionally
 // Little looser, but still not great
-func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging.Logger) error {
+func Unbind(instance *ServiceInstance, clusterConfig ClusterConfig, log *logging.Logger, parameters *Parameters) error {
```
jmrodri: It would be nice to use the same order as `Bind`:

```
func Bind(
    instance *ServiceInstance,
    parameters *Parameters,
    clusterConfig ClusterConfig,
    log *logging.Logger,
) (string, *ExtractedCredentials, error) {
```

pkg/broker/broker.go
```diff
@@ -751,31 +734,28 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
 			return nil, err
 		}
 	} else {
-		a.log.Warning("Broker configured to *NOT* launch and run APB bind")
+		a.log.Warningf("Broker configured to *NOT* launch and run APB bind")
```
jmrodri: I like using the f version, but do we need to change it if we're not passing in variables. I think it's ok, but I wouldn't change them unless we're passing in a variable.

pkg/broker/broker.go
```diff
@@ -751,31 +734,28 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
...
 	instance.AddBinding(bindingUUID)
 	if err := a.dao.SetServiceInstance(instanceUUID.String(), instance); err != nil {
 		return nil, err
 	}
 	// Can't bind to anything if we have nothing to return to the catalog
 	if provExtCreds == nil && bindExtCreds == nil {
-		a.log.Error("No extracted credentials found from provision or bind")
-		a.log.Error("Instance ID: %s", instanceUUID.String())
+		a.log.Errorf("No extracted credentials found from provision or bind instance ID: %s",
+			instanceUUID.String())
```
jmrodri: This is a prime example of requiring the change as we are passing the variable in.

pkg/broker/broker.go
```diff
@@ -29,6 +29,13 @@ var (
 	ErrorBindingExists = errors.New("binding exists")
 )
 
+const (
+	// provisionCredentialsKey - Key used to pass credentials to apb.
+	provisionCredentialsKey = "_apb_provision_creds"
+	// bindCredentialsKey - Key used to pas bind credentials to apb.
+	bindCredentialsKey = "_apb_bind_creds"
```
jmrodri: These are more descriptive but make for longer key names when used as a map key. 

```
params[provisionCredentialsKey]
```

vs
```
params[provisionCredsKey]
```

I guess it's fine.

pkg/broker/broker.go
```diff
@@ -789,21 +769,47 @@ func (a AnsibleBroker) Unbind(
...
+			return nil, err
+		}
+	}
 
-	err = apb.Unbind(serviceInstance, a.clusterConfig, a.log)
-	if err != nil {
-		return nil, err
+	if bindExtCreds != nil {
+		err = a.dao.DeleteExtractedCredentials(bindingUUID.String())
+		if err != nil {
```
jmrodri: is it possible we could leak bindings?

pkg/handler/handler.go
```diff
@@ -262,14 +262,11 @@ func (h handler) unbind(w http.ResponseWriter, r *http.Request, params map[strin
...
-
 	planID := r.FormValue("plan_id")
 	if planID == "" {
 		writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: "unbind request missing plan_id query parameter"})
 	}
-
 	resp, err := h.broker.Unbind(instanceUUID, bindingUUID, planID)
-
 	if errors.IsNotFound(err) {
 		writeResponse(w, http.StatusGone, resp)
```
jmrodri: any reason we deleted the blank lines? they were single lines which isn't egregious and I think it makes it more readable.

scripts/prep_local_devel_env.sh
```diff
@@ -113,7 +113,7 @@ echo -e "Wrote \n${BROKER_SVC_ACCT_TOKEN}\n to: ${SVC_ACCT_TOKEN_FILE}\n"
 ###
 # Fetch the tls.crt for the asb deployment
 ###
-TLS_CRT_DATA=`oc get secret asb-tls -o jsonpath='{ .data.tls\.crt }'`
+TLS_CRT_DATA=`oc get secret asb-tls -o jsonpath='{ .data.tls\.crt}'`
```
jmrodri: What this intentional? we should be consistent and we currently aren't. I would not make this change as part of this PR. Here are other ones in the same file:
```
[jesusr@speed3 ansible-service-broker{authimpl}]$ grep jsonpath scripts/prep_local_devel_env.sh 
BROKER_SVC_ACCT_SECRET_NAME=`oc get serviceaccount asb -n ansible-service-broker -o jsonpath='{.secrets[0].name}'`
SVC_ACCT_CA_CRT_DATA=`oc get secret ${BROKER_SVC_ACCT_SECRET_NAME} -n ${ASB_PROJECT} -o jsonpath='{ .data.service-ca\.crt }'`
BROKER_SVC_ACCT_TOKEN=`oc get secret ${BROKER_SVC_ACCT_SECRET_NAME} -n ${ASB_PROJECT} -o jsonpath='{ .data.token }'`
TLS_CRT_DATA=`oc get secret asb-tls -o jsonpath='{ .data.tls\.crt }'`
TLS_KEY_DATA=`oc get secret asb-tls -o jsonpath='{ .data.tls\.key }'`
etcd_route=`oc get route asb-etcd -n ${ASB_PROJECT} -o=jsonpath=\'\{.spec.host\}\'`
```

pkg/broker/broker.go
```diff
@@ -789,21 +769,47 @@ func (a AnsibleBroker) Unbind(
...
+			return nil, err
+		}
 	}
 
 	err = a.dao.DeleteBindInstance(bindingUUID.String())
 	if err != nil {
 		return nil, err
 	}
-
```
jmrodri: I would have left this blank line as well.

pkg/broker/broker.go
```diff
@@ -789,21 +769,47 @@ func (a AnsibleBroker) Unbind(
...
+	bindExtCreds, err := a.dao.GetExtractedCredentials(bindingUUID.String())
+	if err != nil && !client.IsKeyNotFound(err) {
+		return nil, err
+	}
+	// Add the credentials to the parameters so that an APB can choose what
+	// it would like to do.
+	params[provisionCredentialsKey] = provExtCreds.Credentials
+	if bindExtCreds != nil {
+		params[bindCredentialsKey] = bindExtCreds.Credentials
+	}
```
jmrodri: overall this looks good.

pkg/handler/handler.go
shawn-hurley: No good reason, I will add it back

scripts/prep_local_devel_env.sh
shawn-hurley: I had to fix this and they you submitted the patch, I must have chosen the wrong one on the merge. I will add the space back my fault.

pkg/broker/broker.go
shawn-hurley: I see 4 cases on [777](https://github.com/openshift/ansible-service-broker/pull/302/files/63fb106568574aae52036b0b4b219bdeee23bbf3#diff-55aca6fb74ac953881309c295ee9441eR777) we will attempt to get the credentials.
1. etcd/dao error that is something other than creds could not be found
  - we return the error and error out to service catalog which will re-try.
2. etcd/dao error that is creds not found
  - we don't need to attempt to delete anything that is not there
3. etcd/dao finds creds and delete succeeds 
  - happy path
4. etcd/dao finds creds and delete creds fails.
  - we will error out which will bubble up to service catalog which will re-try.


Can you think of another path?

docs/config.md
eriknelson: +1

pkg/broker/broker.go
```diff
@@ -759,23 +742,20 @@ func (a AnsibleBroker) Bind(instanceUUID uuid.UUID, bindingUUID uuid.UUID, req *
...
-) map[string]interface{} {
-	// TODO: Implement, need to handle case where either are empty
-	return provExtCreds.Credentials
+	if bindExtCreds != nil {
+		err = a.dao.SetExtractedCredentials(bindingUUID.String(), bindExtCreds)
+		if err != nil {
+			a.log.Errorf("Could not persist extracted credentials - %v", err)
+			return nil, err
+		}
+		return &BindResponse{Credentials: bindExtCreds.Credentials}, nil
```
eriknelson: This is better than our original thinking. Keeps the broker dumb and delegates the responsibility for what needs to be in the bind response entirely to the APBs, while still allowing them to include provision credentials since they're passed in as a param during the bind action. Nice job :+1:

pkg/broker/broker.go
```diff
@@ -789,14 +769,41 @@ func (a AnsibleBroker) Unbind(
 		return nil, errors.New(errMsg)
 	}
 
+	params := make(apb.Parameters)
+	provExtCreds, err := a.dao.GetExtractedCredentials(instanceUUID.String())
```
eriknelson: This reads like any APB that has been bound is *required* to have credentials stashed for both provision *and* bind. What if I have a bindable APB that only returns credentials during a bind operation? This will error out unless the provision object is some kind of empty object, which I also don't think we should be doing.

pkg/broker/broker.go
```diff
@@ -789,14 +769,41 @@ func (a AnsibleBroker) Unbind(
...
+	}
+	// Add the credentials to the parameters so that an APB can choose what
+	// it would like to do.
+	params[provisionCredentialsKey] = provExtCreds.Credentials
+	if bindExtCreds != nil {
+		params[bindCredentialsKey] = bindExtCreds.Credentials
+	}
 	serviceInstance, err := a.getServiceInstance(instanceUUID)
 	if err != nil {
 		a.log.Debugf("Service instance with id %s does not exist", instanceUUID.String())
```
eriknelson: Just noticing this; if this branch is taken it's a hard error and we need to return w/ error code. We'll potentially hit an NPE continuing with a nil `serviceInstance`.

This would also be quite bad; if someone is trying to unbind from a service that we don't know exists, we're in a real corrupt state.

pkg/broker/broker.go
```diff
@@ -789,14 +769,41 @@ func (a AnsibleBroker) Unbind(
...
+	if serviceInstance.Parameters != nil {
+		params["provision_params"] = *serviceInstance.Parameters
+	}
+	// only launch apb if we are always launching the APB.
+	if a.brokerConfig.LaunchApbOnBind {
+		err = apb.Unbind(serviceInstance, &params, a.clusterConfig, a.log)
+		if err != nil {
+			return nil, err
+		}
+	}
```
eriknelson: I would add a warning to indicate the APB was not run in the logs, making sure we are clear that nothing was launched in case somebody expects otherwise.

pkg/broker/broker.go
eriknelson: I think we're covered by service cat retries.

PR #309, opened 2017-07-27
Update local scripts to run etcd with a local broker

dymurray:


scripts/prep_local_devel_env.sh
```diff
@@ -159,9 +159,20 @@ while [ "$?" -ne 1 ]; do
   oc get pods -n ${ASB_PROJECT} | grep asb
 done
 
+oc scale deployment etcd --replicas 0 -n ${ASB_PROJECT}
```
rthallisey: Does this exist? Since etcd + asb are in a single deploymentconfig now?

dymurray: @rthallisey No after a catasb run. It only exists if you have run the prep_local_devel_env.sh script but it doesn't cause problems if it doesn't already exist. I put this in so that you can rerun the prep script over and over again with a completely reset ansible-service-broker project.

rthallisey: That's fine, but what about the loop below ```oc get pods -n ${ASB_PROJECT} | grep etcd```.  If you ran catasb first then there is not etcd here.

dymurray: Right so it just skips that loop. I only have that in case it needs to loop.

rthallisey: Gotcha, I misinterpreted the check in the loop.

PR #312, opened 2017-07-27
Create a spec template

rthallisey:
The spec template will be a guide on how to create a spec.


eriknelson:
This is awesome @rthallisey. ACK.

eriknelson:
We really need a CONTIBUTING.md that references this.

PR #316, opened 2017-07-28
get both tls.key AND tls.crt not two tls.keys

jmrodri:
I had an oversight when I reviewed PR 305. The change from using jq to jsonpath also introduced this bug where we stopped looking at the tls.crt which is actually quite important


jmrodri:
## TEST
* make really-clean
* make prepare-local-env
* make run

The broker should start. Without this fix you get error with tls at startup.

PR #326, opened 2017-08-01
Configurable refresh interval of Broker updating specs

rthallisey:
Refresh Interval will be the number of seconds between registry
queries.

Refresh Interval is shared for all registries.

 - New config option refresh interval


rthallisey:
[[13] Add a configurable refresh interval of Broker querying registries for updates](https://trello.com/c/ky2WJTSR/374-13-add-a-configurable-refresh-interval-of-broker-querying-registries-for-updates)

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -170,10 +170,10 @@ objects:
       broker:
         dev_broker: ${DEV_BROKER}
         bootstrap_on_startup: ${BOOTSTRAP_ON_STARTUP}
+        refresh_interval: ${REFRESH_INTERVAL}
         launch_apb_on_bind: ${LAUNCH_APB_ON_BIND}
         output_request: ${OUTPUT_REQUEST}
         recovery: ${RECOVERY}
-        bootstrap_on_startup: true
```
rthallisey: I don't need to, but I saw ```bootstrap_on_startup: ${BOOTSTRAP_ON_STARTUP}``` was defined above. And ```${BOOTSTRAP_ON_STARTUP}``` is defaulted to true below.

pkg/app/app.go
```diff
@@ -164,6 +165,29 @@ func (a *App) Start() {
 		a.log.Notice("Broker successfully bootstrapped on startup")
 	}
 
+	interval := a.config.Broker.RefreshInterval
+	if interval != 0 {
+		ticker := time.NewTicker(time.Duration(interval) * time.Second)
+		quit := make(chan struct{})
```
rthallisey: Agreed, will switch to context.

pkg/app/app.go
```diff
@@ -164,6 +165,29 @@ func (a *App) Start() {
 		a.log.Notice("Broker successfully bootstrapped on startup")
 	}
 
+	interval := a.config.Broker.RefreshInterval
+	if interval != 0 {
+		ticker := time.NewTicker(time.Duration(interval) * time.Second)
```
rthallisey: time.Seconds is type time.Duration. If I don't type cast it as time.Duration I get an error.

etc/example-config.yaml
```diff
@@ -37,3 +37,4 @@ broker:
   output_request: true
   ssl_cert_key: /path/to/key
   ssl_cert: /path/to/cert
+  refresh_interval: 600
```
rthallisey: 👍

pkg/app/app.go
```diff
@@ -164,6 +165,29 @@ func (a *App) Start() {
...
+		quit := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case <-ticker.C:
+					a.log.Info("Broker configured to refresh specs every %v seconds", interval)
```
rthallisey: Sure. I think printing the time is a good idea.

templates/deploy-ansible-service-broker.template.yaml
shawn-hurley: :+1: didn't notice the other one

pkg/app/app.go
shawn-hurley: yeah that is correct, my fault, you don't have to cast if you are hard coding 600. 

I actually have an idea and let me know if you hate it but we could offer a lot more flexibility by using [parseDuration](https://golang.org/pkg/time/#ParseDuration). change interval to a string and parse it on startup. If we don't understand it just do not set up the ticker and warn at that time.

rthallisey: Nice. I'll use parseDuration.

pkg/app/app.go
```diff
@@ -164,9 +165,37 @@ func (a *App) Start() {
...
+	} else {
+		ticker := time.NewTicker(interval)
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		defer cancelFunc()
+		go func() {
+			for {
+				select {
+				case <-ticker.C:
+					a.log.Info("Broker configured to refresh specs every %v seconds", interval)
+					v := <-ticker.C
```
shawn-hurley: I think that you want to do this in the case statement. 
```go
case v := <-ticker.C:
```
This will just block until the tick, I think at least am I wrong?

docs/config.md
```diff
@@ -109,3 +109,4 @@ recovery|Allow the broker to attempt to recover itself by dealing with pending j
 output_request|Allow the broker to output the requests to the log file as they come in for easier debugging|false|N
 ssl_cert_key|Tells the broker where to find the tls key file. Can be ignored if using `--inscure` option but otherwise is required|""|N
 ssl_cert|Tells the broker where to find the tls crt file. Can be ignored if using the `--insecure` option but otherwise is required|""|N
+refresh_interval|The interval to query registries for new image specs|"600s"|N
```
jmrodri: +1

pkg/app/app.go
```diff
@@ -164,9 +165,36 @@ func (a *App) Start() {
 		a.log.Notice("Broker successfully bootstrapped on startup")
 	}
 
+	interval, err := time.ParseDuration(a.config.Broker.RefreshInterval)
+	a.log.Debug("RefreshInterval: %v", interval.String())
+	if err != nil {
+		a.log.Error(err.Error())
+		a.log.Error("Not using a refresh interval")
```
jmrodri: We should print the value `a.config.Broker.RefreshInterval` in the error. If I get an error I want to know what value was entered.

pkg/broker/broker.go
```diff
@@ -52,6 +52,7 @@ type Config struct {
 	OutputRequest      bool   `yaml:"output_request"`
 	SSLCertKey         string `yaml:"ssl_cert_key"`
 	SSLCert            string `yaml:"ssl_cert"`
+	RefreshInterval    string `yaml:"refresh_interval"`
```
jmrodri: +1

scripts/deploy.sh
```diff
@@ -16,6 +16,7 @@ DEV_BROKER="true"
 LAUNCH_APB_ON_BIND="false"
 OUTPUT_REQUEST="true"
 RECOVERY="true"
+REFRESH_INTERVAL="600s"
```
jmrodri: +1 to deploy

templates/deploy-ansible-service-broker.template.yaml
jmrodri: +1 to removing the duplicate. Probably happened during a merge.

pkg/app/app.go
```diff
@@ -164,9 +165,36 @@ func (a *App) Start() {
...
+	} else {
+		ticker := time.NewTicker(interval)
```
jmrodri: do we have a minimum interval? I mean how often can we really bootstrap? We should probably figure out what our minimum is and verify that during reading of the config above.

pkg/app/app.go
```diff
@@ -164,9 +165,36 @@ func (a *App) Start() {
...
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		defer cancelFunc()
+		go func() {
+			for {
+				select {
+				case v := <-ticker.C:
+					a.log.Info("Broker configured to refresh specs every %v seconds", interval)
+					a.log.Info("Attempting bootstrap at %v", v.UTC())
+					if _, err := a.broker.Bootstrap(); err != nil {
```
jmrodri: Will we have a problem with constant bootstraps? So far bootstrap got called at startup which means no active provisions, etc. We would also run it when we called the endpoint which usually meant we probably didn't have any active provisions, etc. Will this auto bootstraps potentially cause problems?

pkg/app/app.go
rthallisey: You'll get the value in the message right before the error check ```a.log.Debug("RefreshInterval: %v", interval.String())```.  Also, I checked all the error messages returned from PraseDuration and they return the interval.  So even if someone doesn't have debug logging level on, they'll get the value in the error plus our message from the broker.

pkg/app/app.go
rthallisey: I have no clue what our minimal interval is. I did do testing as low as 2s, which caused craziness in the logs since the broker was being peppered with requests. I'm sure if someone sets it that low it will be clear what wrong since  ```a.log.Debug("RefreshInterval: %v", interval.String())``` will be constantly showing up in the logs.

I think as far as we should go is warning that the value is low. But even if we did that, it's still pretty obvious what's happening because the logs will be flooded with TONS of spec data plus the interval message.

jmrodri: Fair enough. Let's leave it for now, we can address it if it becomes an issue.

pkg/app/app.go
rthallisey: > Will we have a problem with constant bootstraps? 
>
Yes, Stdout gets overflowed with spec information if there are constant bootstraps. The default setting is 600s = 10m which is far out of the crazy zone (<10s).
> We would also run it when we called the endpoint which usually meant we probably didn't have any active provisions, etc. Will this auto bootstraps potentially cause problems?
>
 ```listen_and_serve``` will spawn go routines per request. The only thing we don't have async bootstraps, which doesn't make a whole lot of sense. So the bootstrap requests will be processed sequentially.  But, I don't think this will cause problems for provision because the last step of bootstrap does: ```_, err := d.kapi.Set(context.Background(), key, val /*opts*/, nil)```. This sets the key/value pairs based on all the spec data gathered. It doesn't delete anything or make the data unavailable, it only adds new data.

PR #327, opened 2017-08-01
updating handler to use FormValue call to retrieve data from query param

### LINKED ISSUE #303: Convert queryparam extractions to FormValue
@shawn-hurley welcomly (is that a word?) pointed out request.FormValue is a much better way of unwrapping request params. Let's update the handler layer to respect it.

shawn-hurley:
Move to a common way of retrieving data from query parameters.

 - remove uses of r.URL.Query() and the resulting value
 - use [request.FormValue](https://golang.org/pkg/net/http/#Request.FormValue)

depends-on <PR>
No 
fixes #303

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -107,12 +107,9 @@ func (h handler) provision(w http.ResponseWriter, r *http.Request, params map[st
 	}
 
 	var async bool
-	queryparams := r.URL.Query()
 
-	if val, ok := queryparams["accepts_incomplete"]; ok {
-		// ignore the error, if async can't be parsed it will be false
-		async, _ = strconv.ParseBool(val[0])
-	}
+	// ignore the error, if async can't be parsed it will be false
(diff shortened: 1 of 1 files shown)


pkg/handler/handler.go
```diff
@@ -290,23 +284,24 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
...
+	if serviceID != "" {
+		req.ServiceID = serviceID
 	}
 
 	// plan_id is optional
-	if val, ok := queryparams["plan_id"]; ok {
-		req.PlanID = val[0]
+	planID := r.FormValue("plan_id")
+	if planID != "" {
+		req.PlanID = planID
```
jmrodri: Would this work?

```
if planID := r.FormValue("plan_id"); planID != "" {
    req.PlanID = planID
}
```

shawn-hurley: Yes, that will work, especially because ```planID``` is not needed, ill update the others as well.

PR #328, opened 2017-08-01
fusor test will now print out details on the actual file

### LINKED ISSUE #320: ft.<assertion> reports the problem from within the ft util
We have a bunch of test assertions that, when they fail, only report the failure inside of the test function instead of the actual place things broke. Makes it almost useless for developer testing purposes. We need to do a better job reporting where the actual problem arose.

shawn-hurley:
Will determine the actual file that caused the issues and print that out during the test run.
Allows user's to more easily see the tests that were broken. and example out put is
```
--- FAIL: TestRegistryLoadSpecsNoError (0.00s)
        util.go:113:
                registry_test.go had error on line 165
                Message: 1 != 2
```

 - get the actual caller of the fusor test method
   - 3 callers up, 0 -> call to Callers, 1 -> call to unexported ```fail``` 2 -> call to  unexported ```assert```, 3 -> call to the public ```AssertXXXX``` method.

depends-on <PR>
No
fixes #320

### DIFF
--- a/pkg/fusortest/util.go
+++ b/pkg/fusortest/util.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"runtime"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


eriknelson:
Thanks @shawn-hurley !

Issue #233, opened 2017-06-26
Proposal: Support both Openshift and Kubernetes


rthallisey:
# Feature: Support other Clusters than OpenShift
The ansible-service-broker is currently OpenShift only because OpenShift has a few features that Kubernetes is still developing. The delta between OpenShift and Kubernetes is going to be a common occurrence so need a way to easily support both. I'm proposing that we can solve this by organizing our code paths into separate pieces and filling the gaps to meet the service-catalog specs:
- Kubernetes operations
- OpenShift operations
- Kubernetes hack operations that perform the actions done in 'OpenShift operations', but for Kubernetes

## Proposed work

1. ~~Split out all the clients into pkg/clients/~~
  1a. ~~https://github.com/openshift/ansible-service-broker/issues/213~~
  2a. ~~https://github.com/openshift/ansible-service-broker/pull/222~~
2. Split out all Cluster logic into a pkg/cluster/openshift.go and pkg/cluster/kubernetes.go
- Kubernetes is used for the majority of the broker actions while OpenShift a small number of actions. Let's break apart these code paths and call into a cluster pkg for cluster actions.
3. Create a pkg/cluster/kubernetes-hack.go that will do OpenShift specific functions the Kubernetes way
4. Use a Kubernetes solution to service accounts
  4a. https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/svc_acct.go
5. Change APBs so they support Kubernetes
  5a. https://github.com/fusor/apb-examples/issues/60
6. Remove all "oc" commands and replace them with Kubernetes and OpenShift client calls
  6a. https://github.com/openshift/ansible-service-broker/search?p=1&q=%22oc%22&type=&utf8=%E2%9C%93

rthallisey:
I think git might've closed this.  Reopening to track more PR's

ron1:
Was this closed prematurely?

rthallisey:
Once #556 merged, the broker was able on kubernetes. 

If you want to use ansible to setup the broker & catalog on kubernetes see: https://github.com/fusor/catasb#openshift-or-kubernetes

If you want to use the template for the broker:  https://github.com/openshift/ansible-service-broker/blob/master/templates/k8s-ansible-service-broker.yaml.j2 and https://github.com/openshift/ansible-service-broker/blob/master/templates/k8s-template.py

rthallisey:
[...]

PR #329, opened 2017-08-02
Spec: Kubernetes and COE agnostic support

rthallisey:
Spec outlining the work for supporting Kubernetes.

https://github.com/openshift/ansible-service-broker/issues/233

### DIFF
--- a/docs/specs/add-kubernetes-support.md
+++ b/docs/specs/add-kubernetes-support.md
@@ -4,7 +4,7 @@
 The ansible-service-broker and all of the tooling around it
 (apb-examples, catasb, and ansible-playbook-bundles) specifically target
 OpenShift as the underlying COE (container orchestration engine) when it
-should be cluster agnostic.
+should be runtime agnostic.
 
 ## Problem Description
 Issues have been coming up during testing because the broker does not target
@@ -19,17 +19,17 @@ The Broker's functionality won't change, but all the tooling around it will.
 
 ## Split Out All Cluster logic
 OpenShift is Kubernetes plus a few features, API name changes, and tools.
-In the core ansible-service-broker code, ```pkg/cluster/kubernetes.go```
+In the core ansible-service-broker code, ```pkg/runtime/kubernetes.go```
 should be able to account for about 80% of the cluster opertations required
-by the broker. The remaining ~20% will be held in ```pkg/cluster/openshift.go```.
+by the broker. The remaining ~20% will be held in ```pkg/runtime/openshift.go```.
 
 ### Code
-Start by creating a new pkg for cluster objects, ```pkg/cluster/...```. From
-there, the cluster pkg will hold ```pkg/cluster/cluster.go```,
-```pkg/cluster/kubernetes.go```, and ```pkg/cluster/openshift.go```.
+Start by creating a new pkg for cluster objects, ```pkg/runtime/...```. From
+there, the runtime pkg will hold ```pkg/runtime/cluster.go```,
+```pkg/cluster/kubernetes.go```, and ```pkg/runtime/openshift.go```.
 
-The file ```pkg/cluster/cluster.go``` will provide an object that will abstract
-cluster logic.
+The file ```pkg/runtime/cluster.go``` will provide an object that will abstract
+runtime logic.
 ```diff
 +type COE interface {
 +     CreateSandbox()
@@ -37,14 +37,14 @@ cluster logic.
 +     DestroySandbox()
 +}
 
-+struct Cluster {
++struct RunTime {
 +     ...
 +}
 
-+func (c *Cluster) ApbAction(...) {
-+     c.cluster.CreateSandbox()
-+     c.cluster.Run(APBActionProvision, req)
-+     c.cluster.DeleteSandbox()
++func (c *RunTime) ApbAction(...) {
++     c.runTime.CreateSandbox()
++     c.runTime.Run(APBActionProvision, req)
++     c.runTime.DeleteSandbox()
 +}
 
 +func (k *Kubernetes) Run(action, req) {
@@ -55,7 +55,7 @@ cluster logic.
 ```
 
 The ```COE``` interface will hold all the public functions whose details are
-hidden away inside the cluster files, kubernetes.go and openshift.go. Adding a
+hidden away inside the runtime files, kubernetes.go and openshift.go. Adding a
 new COE operation requires adding an abstraced function that describes what
 the interface does so it remains agnostic.
 
@@ -70,17 +70,17 @@ type COE interface {
 ```
 
 Using an abstration is an advantage because the broker code will only have a
-single path and have no knowlege of cluster resources. It will call the cluster
-object using broker native information, which will translate into cluster
+single path and have no knowlege of runtime resources. It will call the runtime
+object using broker native information, which will translate into runtime
 resource allocation.
 
 To provision an APB:
 ```diff
 - k8scli.CoreV1().Pods(ns).Create(pod)
-+ Cluster.ApbAction(APBActionProvision, req)
++ RunTime.ApbAction(APBActionProvision, req)
 ```
 
-Down inside the cluster delegates, kubernetes.go and openshift.go, there will be
+Down inside the runtime delegates, kubernetes.go and openshift.go, there will be
 functions that handle the resource specific work.
 ```diff
 + func (k *Kubernetes) CreatePod(name string, image spec.Image, extraVars string,
@@ -143,7 +143,7 @@ functions that handle the resource specific work.
 The diagram represents the code split into two files and the API objects
 in each of them
 
-**pkg/cluster/kubernetes.go**
+**pkg/runtime/kubernetes.go**
 
 | API Resource | Uses API | Uses CLI | Different between OpenShift and Kubernetes |
 |:---:|:---:|:---:|:---:|
@@ -163,45 +163,45 @@ in each of them
 | ClusterRoles |  | X | X |
 | Project |  | X | X |
 
-## Cluster Identification
-Since the plan is for the Broker to run on two clusters, there needs to be a way
-to identify which is being used.
+## Runtime Identification
+Since the plan is for the Broker to run on at least two clusters, there needs to
+be a way to identify which is being used.
 
 ### Broker Configuration and Validation
-The broker will have configuration settings for the cluster.
-```ClusterVersion``` will accept the special character '+'
+The broker will have configuration settings for the runtime.
+```RuntimeVersion``` will accept the special character '+'
 to account for multiple versions.
 
 ```diff
 broker:
-+ Cluster: OpenShift
-+ ClusterVersion: "1.6+"
++ Runtime: OpenShift
++ RuntimeVersion: "1.6+"
 ```
 or
 ```diff
 broker:
-+ Cluster: Kubernetes
-+ ClusterVersion: "1.7"
++ Runtime: Kubernetes
++ RuntimeVersion: "1.7"
 ```
 
 When the broker is started, there will be a validation test to make sure
-the cluster setting is correct.
+the Runtime setting is correct.
 
 ### APB Spec
-APBs will be written only for a single cluster and should identify which cluster
-they work with. If not cluster is specified in the APB, assume it's meant for
-OpenShift all the current examples work.
+APBs will be written only for a single runtime and should identify which runtime
+they work with. If runtime is not specified in the APB, assume it's meant for
+OpenShift so all the current examples work.
 
 ```diff
 bindable: false
-+ cluster: kubernetes
++ runtime: kubernetes
 async: optional
 ```
 
-## APB Cluster Identification
-One of the trickiest parts of having one APB per cluster is identification.
+## APB Runtime Identification
+One of the trickiest parts of having one APB per runtime is identification.
 It needs to be obvious for a user to know that an APB only works with OpenShift
-or Kubernetes. Currently, the only annotation outlined is the 'cluster' option
+or Kubernetes. Currently, the only annotation outlined is the 'runtime' option
 in apb.yml, but there needs to be further identification.
 
 ### New APB Level Directory
@@ -280,7 +280,7 @@ directory can use either the 'New APB Level Directory' strategy or the
 'New Top Level Directory' strategy.
 
 ## APB Containers
-In order to target multiple clusters the APB containers require two changes:
+In order to target multiple runtimes the APB containers require two changes:
  - Packages
  - oc-login.sh
 
@@ -323,7 +323,7 @@ broker and catalog on top of both Kubernetes and OpenShift.  Jenkins will do
 testing on OpenShift.
 
 ## Phased Plan
-Moving to a cluster agnostic architecure will require a large about of effort.
+Moving to a runtime agnostic architecure will require a large about of effort.
 We'll use a phased approach to accomplish the move so things don't get too
 chaotic all at once.
 
@@ -338,12 +338,12 @@ and begin transitioning broker code.
 
 **Ansible-service-broker**
 
- 3a. Create ```pkg/cluster/kubernetes-hack.go``` that will run client commands
+ 3a. Create ```pkg/runtime/kubernetes-hack.go``` that will run client commands
      via  ```kubectl```.
- 3b. Create ```pkg/cluster/openshift-hack.go``` that will run client commands
+ 3b. Create ```pkg/runtime/openshift-hack.go``` that will run client commands
      via  ```oc```.
- 4.  Split out the cluster logic into a ```pkg/cluster/openshift.go``` and
-     ```pkg/cluster/kubernetes.go```.
+ 4.  Split out the cluster logic into a ```pkg/runtime/openshift.go``` and
+     ```pkg/runtime/kubernetes.go```.
 
 ### Phase Two
 Tool Transition - Get all the environment tools working and the gate green.
@@ -354,8 +354,8 @@ Tool Transition - Get all the environment tools working and the gate green.
     Projects, using the clients.
  6. Convert any broker tools and scripts to target either OpenShift or
     Kubernetes.
- 7. Hook up the broker to the Cluster pkg.
- 8. Accept Cluster and ClusterVersion parameters and create validations.
+ 7. Hook up the broker to the RunTime pkg.
+ 8. Accept Runtime and RuntimeVersion parameters and create validations.
  9. Pass $CLUSTER as a variable to the APB.
 
 **apb-examples**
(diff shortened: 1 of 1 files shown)


docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Upstream and Downstream Code Hierarchy
+OpenShift is Kubernetes plus a few features, API name changes, and tools.
+In the core ansible-service-broker code, ```pkg/cluster/kubernetes.go```
+should be able to account for about 80% of the cluster opertations required
+by the broker. The remaining ~20% will be held in ```pkg/cluster/openshift.go```.
+
+### Hierarchy Diagram
+This diagram represents if the current state of code were split into two
+files.
```
jmrodri: This table represents if the code were split into two files.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,91 @@
...
+## APBs
+The APBs only target OpenShift as a cluster. The APBs will differ in two ways:
+ - Packages
+ - oc-login.sh
+
+### Packages
+The main issue with packaging is that folks using Kubernetes may only want the
+Kubernetes client. The package ```origin-clients``` installs both the OpenShift
+and Kubernetes clients.  If this the container size isn't an issue, then there
+may not be a problem using ```origin-clients```.
```
jmrodri: There seems to be two parts to this. 1) the code being compiled into the broker for both 2) the branches will require specific clients i.e. openshift.go would require oc client.  We could put in code to only require the specific oc client IF openshift.go is being invoked. 

There was a blog post Shawn shared with me about plugins in go programs. I haven't read it yet but if there was a way to make these pieces more "pluggable" that could help.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,91 @@
...
+ - Create ```pkg/cluster/kubernetes-hack.go``` that will run client commands
+   via  ```kubectl```.
+ - Create a Kubernetes solution for ClusterRoleBindings, ClusterRoles, and
+   Projects using the clients.
+ - Convert all the client commands to API [calls](https://github.com/openshift/ansible-service-broker/search?p=1&q=%22oc%22&type=&utf8=%E2%9C%93).
+ - Convert any broker tools to target either OpenShift or Kubernetes.
+
+### apb-examples
+ - Add support for Kubernetes to ```oc-login.sh``` script. Also, it should
+   be renamed.
```
jmrodri: ... it should be renamed to `login.sh`

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,91 @@
...
+## Work Items
+### Ansible-service-broker
+ - Split out all Cluster logic into a ```pkg/cluster/openshift.go``` and
+   ```pkg/cluster/kubernetes.go```.
+ - Create ```pkg/cluster/kubernetes-hack.go``` that will run client commands
+   via  ```kubectl```.
+ - Create a Kubernetes solution for ClusterRoleBindings, ClusterRoles, and
+   Projects using the clients.
+ - Convert all the client commands to API [calls](https://github.com/openshift/ansible-service-broker/search?p=1&q=%22oc%22&type=&utf8=%E2%9C%93).
+ - Convert any broker tools to target either OpenShift or Kubernetes.
```
jmrodri: Can we get a first list of the tools that need to change?

docs/specs/add-kubernetes-support.md
shawn-hurley: The [plugin package](https://golang.org/pkg/plugin/) is what @jmrodri is referring to.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,91 @@
...
+may not be a problem using ```origin-clients```.
+
+### Script
+The ```oc-login.sh``` script should be changed to accept an environment varible
+identifing the $COE.  Based on $COE, the login will user the ```kubectl``` or
+```oc``` client.
+
+## Work Items
+### Ansible-service-broker
+ - Split out all Cluster logic into a ```pkg/cluster/openshift.go``` and
```
shawn-hurley: Do we want two clients to share the same package namespace? If this is too in weeds, please just ignore this, but I am little confused on how separate the proposed clients are. Are they only separate for 1 or 2 functions.
Example:
we have a kubernetes client that does 80% of the stuff. then the openshift client has an embedded kubernetes type? Or are they going to be completely separate objects that each contain their own logic for each action?

docs/specs/add-kubernetes-support.md
rthallisey: ```Currently plugins only work on Linux.```

docs/specs/add-kubernetes-support.md
rthallisey: > Do we want two clients to share the same package namespace
>
Yes.
> are they going to be completely separate objects that each contain their own logic for each action?
>
What I'm thinking we do is create an abstraction where the broker has only once code path for things like serviceaccounts, rolebindings, ect... The broker makes a call to the client which handles whether the action needs run with either the openshift or kubernetes client/API.  Kubernetes and openshift will be in different files in the same pkg namespace with different logic for the ~20% of overlapping calls. The result is: If I create a rolebinding and I'm on openshift, the client object will know we need to use the openshift api. If I create a pod in openshift, the client object will know to use the kubernetes api.

Good questions @shawn-hurley. Will add more detail to the doc.

docs/specs/add-kubernetes-support.md
rthallisey: sure

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,209 @@
...
+single path. It will call the cluster object asking for an operation. For
+example, to create a new Pod:
+```diff
+- k8scli.CoreV1().Pods(ns).Create(pod)
++ Cluster.Create(pod, ns)
+```
+
+The Cluster object will recieve the request and call the correct API.
+```diff
++ func Create(c interface{}, ns string) {
```
shawn-hurley: Is there some method they both share that we could make an interface for? That would allow the compiler to make some guarantees.

rthallisey: I'm using ```c interface{}``` to accept any API type as a parameter so I don't think a normal interface will work.  Another way I could do this is a struct to hold variables for all the API types.

shawn-hurley: Why don't we create methods to create each resource?

I guess I am wondering why we are making it so generic, I think that in this case, the caller would probably just prefer to create a pod and not care about the pod object itself or the implementation of how that pod is created.

rthallisey: > I guess I am wondering why we are making it so generic, I think that in this case, the caller would 
> probably just prefer to create a pod and not care about the pod object itself or the implementation 
> of how that pod is created.
>
Create is very generic.  I can make it more specific with ```Create<resource>```.  I agree, being more specific will be easier. 

By being specific, we could do a few things. The caller can create a pod object and pass it to CreatePod or the caller can pass information to ```Create<resource>```.  I like the idea of abstracting the whole thing.

shawn-hurley: I am arguing that we take the values that we need to create a pod and create a function that will take those values and both create the pod struct as well as call the method to whichever client.

The client is responsible for dealing with
1. kube/oc specific errors making them client package errors
2. kube/oc specific structs making the callers unaware of anything specific about the cluster we are running in
3. Making a public(to our broker) API that will allow the broker to create the resources it needs. This is not a generic client that is being imported, it is very specific for just our broker IMO.

rthallisey: I updated it see the latest spec.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,247 @@
...
+The APBs only target OpenShift as a cluster. The APBs will differ in two ways:
+ - Packages
+ - oc-login.sh
+
+### Packages
+The main issue with packaging is that folks using Kubernetes may only want the
+Kubernetes client. The package ```origin-clients``` installs both the OpenShift
+and Kubernetes clients.  If this the container size isn't an issue, then there
+may not be a problem using ```origin-clients```.
+
```
jwmatthews: How do we annotate that an APB requires OpenShift and can not run on Kubernetes.  Say it creates a Route or uses a DeploymentConfig?

I'm assuming any APB that runs on Kubernetes will be able to run on OpenShift.
Yet if an APB truly requires OpenShift (Route, DeploymentConfig, etc) it must have OpenShift and cannot run on pure k8s.

Would we want to allow an APB to downgrade itself based on environment?
i.e., it's written to deploy to OpenShift if available and use a DeploymentConfig, but if on K8S it follows a different path in the APB?

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,247 @@
...
+the downstream CI testing on OpenShift.
+
+## Phased Plan
+Moving to a cluster agnostic architecure will require a large about of effort.
+We'll use a phased approach to accomplish the move so things don't get too
+chaotic all at once.
+
+### Phase One
+Infrastructure setup - Make sure there's a repeatable infrastracture setup
+and begin transitioning broker code.
```
jwmatthews: +1  Want to be sure team has a shared common way to run a script and get a working environment to develop/debug/test in.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,247 @@
...
+
+ - docs/bind-data-transmission.md
+ - docs/local_development.md
+ - README.md
+
+## CI
+The community has two tools for CI: Jenkins and Travis. When adding
+Kubernetes support to the broker, the upstream Travis CI job should test the
+broker and catalog on top of the upstream COE, Kubernetes.  Jenkins will become
+the downstream CI testing on OpenShift.
```
jwmatthews: Do we need to have Travis only test Kubernetes support?

If we could also do a sanity check with oc cluster up like we do now that'd be good feedback to have with a PR.  

Essentially have the gate checking both K8S and OCP behavior.

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,247 @@
...
++		},
++		"roleRef": map[string]string{
++			"name": ApbRole,
++		},
++      }
++
++      if OpenShift {
++         OCcli.CoreV1().ClusterRoleBinding(ns).Create(roleBindingM)
++      } elif Kubernetes {
++         k8scli.CoreV1().ClusterRoleBinding(ns).Create(roleBindingM)
```
jwmatthews: Is there a mechanism in Go where we can have the OCP Object reuse the K8S implementation and just override/implement the deltas?

Opposed to needing to redeclare, if OCP/if K8S workflows

Assuming OCP workflow can reuse k8scli except for those cases where there are deltas in behavior.

docs/specs/add-kubernetes-support.md
rthallisey: Good questions @jwmatthews.

> How do we annotate that an APB requires OpenShift and can not run on Kubernetes. Say it 
> creates a Route or uses a DeploymentConfig?
>
I think this is up to the user and we need to document 'Creating an APB' very well so it's clear where openshift and kubernetes differ. Looking at this as an APB writer, if I'm using Kubernetes, I'm going to write my apb with the resources I know Kubernetes uses. Ditto OpenShift. The case I'm concerned about is someone who didn't write the APB trying to use it with the wrong cluster. But I think we make it clear by having maybe a separate playbook directory, git repo, or top level directory for Kubernetes and OpenShift APBs.

> I'm assuming any APB that runs on Kubernetes will be able to run on OpenShift.
>
Not 100% sure. What I'm aware of that could be problematic is the ClusterRoleBinding resource. But, that resource is mapped to the OpenShift security policy where as in Kubernetes it's Rbac. So I think it will work, but function different.

Either way, I think we should discourage this behaviour and treat the playbooks and the clusters as completely independent entities. 
>Would we want to allow an APB to downgrade itself based on environment?
>
No. Let's keep them separate.  I think to have success here requires clean separation of the playbooks so we communicate to the user that didn't write the playbook what it's used for.

docs/specs/add-kubernetes-support.md
rthallisey: > Do we need to have Travis only test Kubernetes support?
>
We can have travis test both.

docs/specs/add-kubernetes-support.md
eriknelson: >But, that resource is mapped to the OpenShift security policy where as in Kubernetes it's Rbac. So I think it will work, but function different.

From my testing, RBAC entities are entirely ignored by OpenShift. It's very deceptive, because I was able to create RBAC *RoleBindings that appeared to have created absolutely fine (because they did), so my expectation was that they were, well...doing what role bindings do, but they were not from an openshift perspective. I am very hesitant to say this will work.

> Either way, I think we should discourage this behaviour and treat the playbooks and the clusters as completely independent entities.

+1. I think fundamentally we should encourage good design practice around this. Organize logic in such a way where shared logic can be passed around with things like roles, and encapsulate pieces that are specific to certain clusters.

WRT annotation, I would like to see something in the apb spec that tells the broker what COE a given APB supports. I think the broker should validate early, know what cluster it's *able* to deploy to and what APBs are *possible* to be deployed to that cluster. That way the broker can say `if !clusterSupports(apb) { handleUnsupported() } else { continue }`

docs/specs/add-kubernetes-support.md
rthallisey: > Is there a mechanism in Go where we can have the OCP Object reuse the K8s implementation 
> and just override/implement the deltas?
>
From the broker's perspective, that's essentially what I'm doing. The broker calls ```Cluster.CreatePod()``` and has no idea how the pod is created, but gets back the pod information.  The cluster pkg handles the abstraction and figures out what cli/api to call.
> Opposed to needing to redeclare, if OCP/if K8S workflows
> 
I agree that the way we did it before wasn't good, but these checks aren't 'branching' the broker code like before.  The extend of the 'branch' only encompasses one line, the client/API call.  And the broker won't know this is happening because it's completely abstracted.  The design is similar to a plugin, where the plugin code is abstracted away from the core broker code and does whatever it wants under the covers.  Also different from before, ```OpenShift``` will come from the config file instead broker core code and we'll verify it when we start the broker.

eriknelson: Yes, albeit a bit differently than traditional virtual methods. Go chooses to use composition over inheritance. On a high level, it's all about sharing behavior and delegation of specifics to components that know better (also providing encapsulation). We've organized our registry code and adapters in a similar way, so there is precedence, and I think we can leverage a similar pattern.

Spitballing, we can have a parent COE object that exposes the set of methods the broker needs to run APB actions. I would take the `RunPod` abstraction even further to `RunApbAction`. The fact that APBs are actually run as pods is an implementation detail; also, it provides optionality to run an APB action in a completely different way, (in an environment where the concept of pods don't even exist). Broker operates here on a higher level as well, calling into `coe.RunApbAction(ActionProvision, params)`.

Broker can build a coe runner via factory method, which is responsible for constructing a coe runner based on the current, concrete cluster the broker is configured to target: `runner := NewCOERunner(ClusterTypeKubernetes)`. Inside this, we can construct a parent COE with a `clusterDelegate ClusterDelegate` field, where ClusterDelegate is an interface sitting in front of cluster specifics. We can implement a `ClusterDelegate` for kubernetes and for openshift, and this is where the specialized behavior belongs.

Pseudocode:


```go
type ClusterDelegate interface {
    SetupPermissions()
}

func (r *COERunner) RunApbAction(...) {
    r.sharedBehavior(...)
    r.clusterDelegate.SetupPermissions()
}
```

In this case, we know we need to `SetupPermissions` for all cluster types, they just need to be done in a different way, which the delegates define themselves. For similar cases, think about what's actually getting achieved where differences arise, abstract and declare that behavior (`SetupPermissions`) on the `ClusterDelegate`, and then allow the specific implementations to define what it actually means to `SetupPermissions`.

More food for thought, this very much reminds me of PIMPL from the c++ world, or the Bridge Pattern:
http://en.cppreference.com/w/cpp/language/pimpl
https://en.wikipedia.org/wiki/Bridge_pattern

docs/specs/add-kubernetes-support.md
rthallisey: > WRT annotation, I would like to see something in the apb spec that tells the broker what COE a 
> given APB supports. I think the broker should validate early, know what cluster it's able to deploy 
> to and what APBs are possible to be deployed to that cluster. That way the broker can say if 
> !clusterSupports(apb) { handleUnsupported() } else { continue }
>
@eriknelson good idea.

What if we create a field in apb.yaml for ```Cluster: (OpenShift/Kubernetes)```.
From the broker, we read the specs and identify which specs will work on the
current cluster and ignore the specs that don't work.

jmrodri:
@rthallisey @shawn-hurley @eriknelson I setup a meeting for 11am tomorrow to discuss this. I found it worked well with registries and filtering.

rthallisey:
@jmrodri Sounds good. Can you post something on the ML about the discussion you scheduled?

eriknelson:
photo_2017-08-09_12-27-19

docs/specs/add-kubernetes-support.md
```diff
@@ -0,0 +1,363 @@
...
+When the broker is started, there will be a validation test to make sure
+the cluster setting is correct.
+
+### APB Spec
+APBs will be written only for a single cluster and should identify which cluster
+they work with.
+
+```diff
+bindable: false
++ cluster: kubernetes
```
jwmatthews: We will need to track the version of the cluster type as well in the APB
Assuming some APBs may only work with k8s 1.6, others 1.7+ etc

docs/specs/add-kubernetes-support.md
```diff
@@ -128,21 +128,127 @@ in each of them
...
+## APB Cluster Identification
+One of the trickiest parts of having one APB per cluster is identification.
+It needs to be obvious for a user to know that an APB only works with OpenShift
+or Kubernetes. Currently, the only annotation outlined is the 'cluster' option
+in apb.yml, but there needs to be further identification.
+
+### New APB Level Directory
+Add a new directory layer into each APB.
+
+mediawiki123-apb/
```
jmrodri: can you surround these directory listings in ```

eriknelson:
Example of the runner pattern, important types introduced are the `InputBundle`, `COERunner`, `COEDelegate`, and the concrete delegate types.

https://gist.github.com/eriknelson/4c2c29ea107e6e197fb03ea7abe19a8d

Runnable: https://play.golang.org/p/Rv1ktceNEj

rthallisey:
I think this is good to go since this covers the major design topics.  If anything is left to add I think it will be minor and we can deal with it on the fly.

docs/specs/add-kubernetes-support.md
```diff
@@ -28,20 +28,62 @@ Start by creating a new pkg for cluster objects, ```pkg/cluster/...```. From
...
+The file ```pkg/cluster/cluster.go``` will provide an object that will abstract
 cluster logic.
+```diff
++type COE interface {
++     CreateSandbox()
++     Run(APBActionProvision APBAction, req Context)
++     DestroySandbox()
++}
+
++struct Cluster {
```
eriknelson: One question since this was discussed as a `Delegate`: Do we care about runtimes that are **not** kubernetes or openshift? For example: vanilla docker? IMO, the answer is "No, but we should design for the possibility" since we've already had interest expressed from outside the team. With that in mind, are these other runtimes always going to be "Clusters"? I don't think it would be appropriate to call docker that.

My request would be to keep this as the runner's `Delegate`, curious how other folks feel about that. Otherwise, looks great. Thanks for writing this all up @rthallisey.

rthallisey: @eriknelson that's true, Docker would not be a cluster.  Something like Delegate wfm or maybe Runtime?

PR #332, opened 2017-08-04
remove trailing slash

jmrodri:
Fix the trailing slash error during auto builds of RPM containers. This problem was introduced by the auth PR.


eriknelson:
ACK

PR #336, opened 2017-08-04
Update AddApb to use FQNames

eriknelson:
* Move to FQNames broke apb push. Need to make sure we ID them
correctly before they're inserted to etcd
* Some extra logging for visibility in the future


pkg/broker/broker.go
```diff
@@ -872,6 +872,10 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
 
 //AddSpec - adding the spec to the catalog for local development
 func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
+	a.log.Debug("broker::AddSpec")
```
jmrodri: what is this C++?

PR #337, opened 2017-08-04
Accept ints from exported credentials

### LINKED ISSUE #335: Non-quoted numeric bind credentials cause other bind credentials to disappear from secrets.
> ### Bug: Non-quoted numeric bind credentials cause other bind credentials to disappear from secrets.

**What happened**:

1. Ran a MySQL APB containing an `asb_encode_binding` command including a port number.
```
- name: Encode bind credentials
  asb_encode_binding:
    fields:
      DB_TYPE: "mysql"
      DB_PASSWORD: "{{ MasterUserPassword }}"
      DB_NAME: "{{ DBName }}"
      DB_HOST: "{{ rds_endpoint_address }}"
      DB_PORT: "{{ PortNumber }}"
      DB_USER: "{{ MasterUsername }}"
```
 2. Created a hello-world APB to bind to the database 
 3. Binded the hello-world APB to database
 4. Inspected the secret created as a result of bind
 5. Noticed that secret created only contained the first 3/6 encoded fields

 6. Added escaped quotes around port number encode line
```
      DB_PORT: "\"{{ PortNumber }}\""
```
 7. Re-deployed MySQL APB, hello-world APB
 8. All bind credentials showed up as expected

 
Same as above but expected all fields to show up in secret without needing to put quotes around DB_PORT value. Alternatively, would have been useful to receive an error message explaining the issue and workaround.

eriknelson:
- Fixes issue preventing broker from handling integer credentials

fixes #335

### DIFF
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -83,7 +83,7 @@ func buildExtractedCredentials(output []byte) (*ExtractedCredentials, error) {
 	return &ExtractedCredentials{Credentials: creds}, nil
 }
 
-func decodeOutput(output []byte) (map[string]string, error) {
+func decodeOutput(output []byte) (map[string]interface{}, error) {
 	str := string(output)
 
 	decodedjson, err := base64.StdEncoding.DecodeString(str)
@@ -91,7 +91,7 @@ func decodeOutput(output []byte) (map[string]string, error) {
 		return nil, err
 	}
 
-	decoded := make(map[string]string)
+	decoded := make(map[string]interface{})
 	json.Unmarshal(decodedjson, &decoded)
 	return decoded, nil
 }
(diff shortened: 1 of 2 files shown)


shawn-hurley:
ACK

jmrodri:
In master the unit test will result in a failure (expected).
```
=== RUN   TestInt
--- FAIL: TestInt (0.00s)
        util.go:113: 
                ext_creds_test.go had error on line 37
                Message: port does not match

```

jmrodri:
With your fix it passes:
```
=== RUN   TestInt
--- PASS: TestInt (0.00s)
```

jmrodri:
What's interesting is that it parses it as a float64 not an int.

jmrodri:
What's also weird is that `func decodeOutput(output []byte) (map[string]string, error)` has returned `map[string]string` since the beginning, why are things breaking now?

PR #340, opened 2017-08-07
Updated deployment template to use string substitution when applicable

dymurray:
This PR fixes yaml errors when using characters in a password
 - Use strings for config map

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -178,7 +178,7 @@ objects:
         color: true
       openshift:
         host: "${CLUSTER_AUTH_HOST}"
-        ca_file: "{CA_FILE}"
+        ca_file: "${CA_FILE}"
         bearer_token_file: "${BEARER_TOKEN_FILE}"
         image_pull_policy: "${IMAGE_PULL_POLICY}"
       broker:
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -162,12 +162,12 @@ objects:
   data:
     broker-config: |
       registry:
-        - type: ${REGISTRY_TYPE}
-          name: ${REGISTRY_NAME}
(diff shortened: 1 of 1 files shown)


templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -177,14 +177,14 @@ objects:
         level: debug
         color: true
       openshift:
-        host: ${CLUSTER_AUTH_HOST}
-        ca_file: ${CA_FILE}
-        bearer_token_file: ${BEARER_TOKEN_FILE}
-        image_pull_policy: ${IMAGE_PULL_POLICY}
+        host: "${CLUSTER_AUTH_HOST}"
+        ca_file: "{CA_FILE}"
```
jmrodri: You removed the $ from {CA_FILE} this causes the ca_file setting to be "{CA_FILE}" which is not correct.

jmrodri:
CI failed

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -177,14 +177,14 @@ objects:
...
+        host: "${CLUSTER_AUTH_HOST}"
+        ca_file: "${CA_FILE}"
+        bearer_token_file: "${BEARER_TOKEN_FILE}"
+        image_pull_policy: "${IMAGE_PULL_POLICY}"
       broker:
         dev_broker: ${DEV_BROKER}
         bootstrap_on_startup: ${BOOTSTRAP_ON_STARTUP}
-        refresh_interval: ${REFRESH_INTERVAL}
+        refresh_interval: "${REFRESH_INTERVAL}"
         launch_apb_on_bind: ${LAUNCH_APB_ON_BIND}
```
rthallisey: Any reason why we shouldn't quotes all the variables in the configmap?

dymurray: Yes booleans fail if you convert them into strings when umarshalling the yaml file. So everything is treated as a string other than bools.

PR #345, opened 2017-08-08
Add secret support to the Broker

fabianvf:
This PR allows a broker admin to create associations between apbs and secrets, so that they can be securely added to an APB.

 - Add `secrets` section to broker config. Will associate certain secrets with certain apbs
 - Read in the keys from specified secrets 
   - Mask parameters provided by a secret on a `/catalog` request
- Add volumeMounts and volumes to apb pod to mount secrets into apb container
- if an an apb needs to consume a secret, run it in the broker's namespace with permissions to execute against other namespaces.
- Refactor return values of ExecuteApb into a wrapper struct that contains relevant context (pod name, namespace, serviceaccount)
- Move service account deletion out of the broker and into the apb actions

This is a super initial version of it, right now it's just a list of direct secret-apb associations. WIP is allowing patterns rather than exact matches.

I have a built version of this broker for testing at docker.io/fabianvf/ansible-service-broker:latest

https://trello.com/c/oPvl3Rdy

## Testing

1. Set up an openshift environment with the broker image `fabianvf/origin-ansible-service-broker:latest` and the dockerhub org `fabianvf`. There is a single APB in that org (`rhscl-postgresql-apb`). 

1. Once your cluster is up and my broker is running, run (from the ansible-service-broker project)
    ```bash
    ./scripts/create_broker_secret.py test ansible-service-broker docker.io/fabianvf/rhscl-postgresql-apb  postgresql_user=fabian postgresql_password=testpassword postgresql_version="9.5" 
    ```
1. This will create a secret named test containing the specified keys, and update the broker config to associate it with my APB. 

1. Now, either wait for the catalog to make a new catalog request (this will update the parameters in the UI) (or trigger it if you know how). 

1. Once the catalog request has been made, go to the openshift UI, select the postgresql-apb app, click Next, select whatever namespace you want.
 
1. Note that the parameters `postgresql_user`, `postgresql_password`, `postgresql_version` should not be visible. If you can still see them, either the catalog has not yet updated, or I'm terrible at my job.  

1. Click next and then create.

1. There should be an APB spawned inside the broker namespace. Postgresql should be deployed into whatever namespace you specified. Verify that postgres is up and running in your target namespace.

1. Now create a bind

1. Open up the secret created by the bind, verify that the contents of the secret match the values you passed into the create_secret script.


fabianvf:
[(13) Ability to read secrets from file/env variables in the broker](https://trello.com/c/oPvl3Rdy/377-13-ability-to-read-secrets-from-file-env-variables-in-the-broker)

pkg/app/app.go
```diff
@@ -12,6 +12,7 @@ import (
 	kubeversiontypes "k8s.io/apimachinery/pkg/version"
 
 	logging "github.com/op/go-logging"
+	"github.com/openshift/ansible-service-broker/pkg/apb"
```
jmrodri: I don't think app should import apb ever.

fabianvf:
@eriknelson @jwmatthews this PR is getting unwieldy, how do you feel about reviewing + merging it in its current semi-complete state (doesn't support patterns, just direct apb - secret connections), and then I can implement the next piece without being so diverged from master.

jmrodri:
@fabianvf fix linter and tests before submitting it for review. `make lint` and `make test` should work.

fabianvf:
@jmrodri sorry, all those are fixed.

pkg/apb/types.go
```diff
@@ -200,3 +201,10 @@ type RecoverStatus struct {
 	InstanceID uuid.UUID `json:"id"`
 	State      JobState  `json:"state"`
 }
+
+// ExecutionContext - Contains the information necessary to track and clean up an APB run
+type ExecutionContext struct {
```
jmrodri: Why couldn't we reuse the `Context` struct which already has a `Namespace` and `Platform`?

pkg/apb/provision.go
```diff
@@ -45,26 +45,36 @@ func Provision(
...
 	if err != nil {
-		log.Errorf("Problem executing apb [%s]", podName)
+		log.Errorf("Problem executing apb [%s]", executionContext.PodName)
 		log.Error(err.Error())
-		return podName, nil, err
+		return executionContext.PodName, nil, err
 	}
 
-	creds, err := ExtractCredentials(podName, instance.Context.Namespace, log)
+	creds, err := ExtractCredentials(executionContext.PodName, executionContext.Namespace, log)
```
jmrodri: It's weird to have an instance.Context and this executionContext mixed in the same method.

pkg/broker/provision_job.go
```diff
@@ -59,8 +56,8 @@ func (p *ProvisionJob) Run(token string, msgBuffer chan<- WorkMsg) {
 		return
 	}
 
-	p.log.Info("Destroying APB sandbox...")
-	sm.DestroyApbSandbox(podName, p.serviceInstance.Context.Namespace)
+	msgBuffer <- ProvisionMsg{InstanceUUID: p.serviceInstance.ID.String(),
+		JobToken: token, SpecID: p.serviceInstance.Spec.ID, PodName: podName, Msg: "", Error: ""}
```
jmrodri: why are we sending the `ProvisionMsg` twice?

pkg/broker/provision_job.go
```diff
@@ -59,8 +56,8 @@ func (p *ProvisionJob) Run(token string, msgBuffer chan<- WorkMsg) {
 		return
 	}
 
-	p.log.Info("Destroying APB sandbox...")
-	sm.DestroyApbSandbox(podName, p.serviceInstance.Context.Namespace)
```
jmrodri: I'm not 100% sure why we moved the serviceaccountmanager into the calls out from the job. Is it simply to avoid duplication?

pkg/app/testdata/test-config.yaml
```diff
@@ -42,3 +43,7 @@ broker:
       enabled: true
     - type: oauth
       enabled: false
+secrets:
+  - title: Best secret
+    apb_name: test
+    secret: test_secret
```
jmrodri: :+1: config looks good

pkg/app/config.go
```diff
@@ -46,6 +47,13 @@ func CreateConfig(configFile string) (Config, error) {
 		return Config{}, err
 	}
 
+	if config.Openshift.Namespace == "" {
+		if dat, err = ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err != nil {
+			return Config{}, err
+		}
+		config.Openshift.Namespace = string(dat)
+	}
```
jmrodri: Why do we need the namespace on the service account

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
+## Motivation
```
rthallisey: This should be two separate patches. One with the proposal and the other with the implementation.

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+##### AWS Environment specifics
+- Region
+- VPC_ID
+- Subnets
+- T.B.D. network related variables for RDS, etc.
+
+## Proposed Solution
+
+The Cluster Administrator will create a Secret ahead of time with the data they want to hide from end user.  The Secret will exist in the namespace of the Broker which the end user likely does not have access to.
+The Broker will have a section in it’s configuration file to list Secrets and associate to APBs.  When the Broker runs an APB that has an associated Secret, the Secret is mounted as a file to the running Pod and the APB is run in the Broker’s namespace.
```
rthallisey: Why does the broker have a list of secrets? That would mean if you want to add a new secret you have to restart the broker.  Maybe the secret should be in apb.yml?

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Details
+
+#### Broker Perspective:
+- Modify the Broker’s configuration file and add a new section for Secrets.
+- The Secrets section will contain a list of Secrets referred to by name which must exist in the Broker namespace.
+- Each referenced Secret will have a list of regexes for whitelisted APBs to apply the Secret to.
+    - For first iteration it will just be a 1-1 mapping of secret to apb name
+- It’s possible for an APB to be referenced in multiple secrets
+- If Broker is running an APB that matches an entry in the Secret config:
```
rthallisey: ```If the Broker is...```

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+    - Its pod is assigned ALL of the Secrets which match the configuration.
+    - The Secrets are mounted as files to a known location on the APB Pod.
+    - Secrets are mounted for all methods on the APB
+    - The APB runs in the namespace of the Secret.  (For first implementation this will be Broker namespace)
+    - The generated artifacts will continue to be populated in the namespace the user selects at time of provision, it’s only the ABP pod itself which will run in the namespace of the secret.
+    - During the /catalog operation the Broker will programmatically update the parameters associated with a ServiceClass/Plan.
+    - If the APB has an associated Secret AND the Secret exists, the Broker will read the ‘keys’ from the Secret and remove the matching parameters from being associated with the ServiceClass.
+    - This is done to remove the WebUI from prompting for this values which will obtained from Secret, yet it allows the APB author to continue to declare the parameters as they normally would, and in if no Secret is available the APB is still usable with those parameters entered.
```
rthallisey: s/this values/these values/

rthallisey: s/from Secret/from the Secret/

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+    - It’s possible for a Secret to not yet exist, the APB workflow will continue as is until the Secret is created.  The next invocation of the /catalog operation will update the parameters to reflect what Secrets exist at that point of time.
```
rthallisey: is the /catalog operation a bootstrap?

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+#### APB Perspective
+Desire is to _not_ impact an APB author with this workflow.  APB author will declare the parameters required for the application to function as they normally would.  Details for obtaining the Secret info is handled in the base image.
+
+If the Secret exists for the APB, then the parameters will come from the Secret and the end user will not be prompted for them,  otherwise the APB will function as normal and it will be expected end user provides values for each parameter declared.
+
+#### APB Base Image
+Secrets will be mounted to a known location on the Pod
```
rthallisey: What location?

docs/proposals/secrets.md
```diff
@@ -0,0 +1,80 @@
...
+Base Image will pass secret data in through the ansible mechanism for passing in extra variables in a file:
+“ -e @${PATH_TO_SECRETS}/my_secret_vars.yml”
+Ensuring no leakage of parameters contained in a secret through debug/log statements
+We will write a layer between the playbook and the output logs that removes any secrets from the log, guaranteeing that they will never show up in plain text
```
rthallisey: Where do we output passwords in the logs? We changed the broker bind output stuff so it no longer does that and the broker side no longer logs passwords.  If there is an issue with logging sensitive info in the playbook logs, then I think it should be handled in ansible.

pkg/apb/executor.go
```diff
@@ -76,20 +88,49 @@ func ExecuteApb(
...
-	log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, ns))
+	log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, executionContext.Namespace))
 	k8scli, err := clients.Kubernetes(log)
 	if err != nil {
-		return apbID, err
+		return executionContext, err
+	}
+	_, err = k8scli.CoreV1().Pods(executionContext.Namespace).Create(pod)
+
+	return executionContext, err
```
fabianvf: Yeah that was my instinct as well, the only reason I'm returning the partial execution context is because the function was returning some values and an error already, and I didn't want to break anything that may have been using it.

docs/proposals/secrets.md
fabianvf: yeah, it's definitely a little weird, but the reasons that we did that were to 
a) avoid requiring changes to apbs 
b) avoid requiring apb authors to determine the specific secrets they need and where they are located 
c) allow admins deploying the broker to have full control over what secrets are used, what they contain, and who gets to use them

But yes, if you want to update associations between secrets + apbs, you will need to modify the broker config and restart the broker. You can update secrets that are already associated without broker restarts though.

fabianvf: @jwmatthews if you want to weigh in ^

docs/proposals/secrets.md
fabianvf: No, `/catalog` is when the catalog is requesting the apb specs from the broker.

docs/proposals/secrets.md
fabianvf: yes, this will be a change to the base image, ideally a pure ansible solution, possibly a callback plugin or something similar. Broker shouldn't have to do anything for this.

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+	apbName string
+	secret  string
+}
+
+type secretsCache struct {
+	mapping map[string][]string
+	rwSync  sync.RWMutex
+	rules   []AssociationRule
+	config  []SecretsConfig
+	log     *logging.Logger
```
fabianvf: Yeah, it's very strange and gross. Talked with @shawn-hurley about it and we also concluded that this was just a symptom of the larger logging anti-pattern we have going.

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+var secrets secretsCache
+
+// GetSecrets - Returns a list of secrets to be attached to a specified spec
+func GetSecrets(spec *Spec) []string {
+	secrets.rwSync.RLock()
+	defer secrets.rwSync.RUnlock()
+	return secrets.mapping[spec.FQName]
+}
```
fabianvf: It does impact testing annoyingly, but I made it take the entire spec so that no other modules need to care about how we key the specs in the cache. If we change it to be keyed on something other than FQName in the next iteration (which I could see happening), we won't be changing the API to this module.

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+
+}
+
+func addSecret(spec *Spec, rule AssociationRule) {
+	secrets.mapping[spec.FQName] = append(secrets.mapping[spec.FQName], rule.secret)
+}
+
+func match(spec *Spec, rule AssociationRule) bool {
+	return spec.FQName == rule.apbName
+}
```
fabianvf: Yeah, I separated a lot of these functions out pre-emptively because I anticipate that they will become significantly more complex with the second phase of these changes.

pkg/apb/types.go
fabianvf: I could do that (add PodName adn ServiceAccount to the context struct), though we would have 2 unrelated Context in the same scope, which may be confusing. Nothing against it really though, just was trying not to change too many existing structures.

pkg/app/config.go
fabianvf: Don't necessarily, that's just where I knew I could find the broker namespace. Is there a better place I should look?

rthallisey:
Any links for the issues we want to track with this?
1) avoid having to restart the broker when adding a secret
2) see if there is a way to using existing resources like namespaces and role bindings to safely provide APBs with secrets

pkg/broker/provision_job.go
fabianvf: It seemed unnecessary to have it here (the sandboxes are created in the apb module but destroyed in the broker module), especially since the logic for it was duplicated in `provision_job.go`, `deprovision_job.go` and `broker.go` (implemented for all 4 actions in there too, so provision and deprovision were actually implemented twice). 

Since I required the extra context to delete the sandboxes (`serviceInstance.Context.Namespace` is not always where the apb is running anymore), I figured I'd move this logic to the end of the action execution, where I think it makes more sense and where I can more closely track the effects of my changes.

pkg/broker/provision_job.go
fabianvf: This looks like a rebase mistake.

pkg/apb/executor.go
```diff
@@ -42,28 +43,39 @@ func ExecuteApb(
...
+		executionContext.Namespace = clusterConfig.Namespace
+		roleScope = "ClusterRole"
+	} else {
+		executionContext.Namespace = context.Namespace
+		roleScope = "Role"
+	}
+	executionContext.PodName = fmt.Sprintf("apb-%s", uuid.New())
 
 	sam := NewServiceAccountManager(log)
-	serviceAccountName, err := sam.CreateApbSandbox(ns, apbID)
```
fabianvf: I could see that, just seemed like a vector for mistake without much benefit

docs/proposals/secrets.md
fabianvf: This is a good question, right now it's going to `/etc/apb-secrets/`, but that doesn't really feel right to me. Any suggestions?

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+
+// NewSecrets - Generates AssociationRules from config and initializes the global secrets cache
+func NewSecrets(config []SecretsConfig, log *logging.Logger) {
```
eriknelson: +1, we use `NewT` effectively as a constructor for types.

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+		if match(spec, rule) {
+			addSecret(spec, rule)
+		}
+	}
+
+}
+
+func addSecret(spec *Spec, rule AssociationRule) {
+	secrets.mapping[spec.FQName] = append(secrets.mapping[spec.FQName], rule.secret)
+}
```
fabianvf: I anticipate this function getting more logic in future iterations, so I pre-emptively broke it out to make refactoring/updating it easier.

pkg/apb/secrets.go
fabianvf: Yeah, this was a refactoring artifact, originally we were returning a secrets object instead of making the global secrets cache. Good catch.

pkg/apb/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+var secrets secretsCache
+
+// GetSecrets - Returns a list of secrets to be attached to a specified spec
+func GetSecrets(spec *Spec) []string {
+	secrets.rwSync.RLock()
+	defer secrets.rwSync.RUnlock()
+	return secrets.mapping[spec.FQName]
+}
+
+// AddSecrets - Uses the AssociationRules generated from config to link specs to secrets and add them to the global secrets cache
```
fabianvf: Should I wrap code to 80 as well?

pkg/apb/types.go
fabianvf: Not sure if it needs to be at a type level, but there's probably a better way I can do this that makes more clear the difference between the context of the service instance and the context of the apb. Open to suggestions if you have any, I'm going to think on it for a bit.

pkg/apb/executor.go
fabianvf: Actually, some context does need to be returned if it has been created (such as namespace or service account name), so that we can run cleanup on it correctly. 

I could move the cleanup code into the executor module, in which case it would also have be responsible for doing stuff like gathering creds etc. That might actually be better and a little more sensical, since all the code for interacting with apbs would be contained in the executor module, the actions wouldn't need to know how things are done or how to clean up what the executor module has done. It might also make it easier to switch to different cluster types etc in the longer term.

jmrodri:
@fabianvf rebase please. Also thank you for addressing the review comments.

fabianvf:
I saw a failure on my last run on Friday, still investigating

shawn-hurley:
ACK

jmrodri:
ACK

PR #348, opened 2017-08-09
Prevent CI failures when building the broker

rthallisey:
Travis occasionally fails when building the broker image.
Add a retry around the build and add error checking around
all the setup operations.

 - Retry when building images
 - Add a bunch of new logging around the cluster setup
 - Remove another sleep


jmontleon:
ACK

PR #350, opened 2017-08-09
Update broker makefile

djzager:
The purpose of this update is to start bringing some of the automation
for CI into one place. In the current state, I test locally through
several executions of the travis shell script. What I propose as an
alternative is to bring in as much of that shell script into a Makefile
as possible.

The intention is to make it so a contributor can simply run `make ci` to
run a local version of what travis will accomplish when a PR is
created and add a help target that prints helpful imformation.

- Removed all but the install portion of travis.sh and made that a
function to be called
- Simplified the Makefile and travis.yaml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,11 @@ sudo: required
 group: edge
 
 before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y python-apt autoconf pkg-config e2fslibs-dev libblkid-dev zlib1g-dev liblzo2-dev asciidoc
+  - ./scripts/travis.sh before_install
 install:
-  - sudo pip install ansible==2.3.1
   - ./scripts/travis.sh install
 before_script:
-  - sudo ufw disable
-  - tmp=`mktemp`
-  - echo '{"insecure-registries":["172.30.0.0/16"]}' > ${tmp}
-  - sudo mv ${tmp} /etc/docker/daemon.json
-  - sudo mount --make-shared /
-  - sudo service docker restart
-  - export ERROR=false
+  - ./scripts/travis.sh before_script
 script:
   - ./scripts/travis.sh format
   - ./scripts/travis.sh vet
@@ -26,4 +18,3 @@ script:
   - ./scripts/travis.sh build
   - ./scripts/travis.sh test
   - ./scripts/travis.sh ci
-  - make ci LOCAL_CI=false
--- a/Makefile
+++ b/Makefile
@@ -5,33 +5,49 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
 BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/build"
 PREFIX           ?= /usr/local
 BROKER_CONFIG    ?= $(PWD)/etc/generated_local_development.yaml
+SOURCE_DIRS      = cmd pkg
 SOURCES          := $(shell find . -name '*.go' -not -path "*/vendor/*")
 SVC_ACCT_DIR     := /var/run/secrets/kubernetes.io/serviceaccount
 KUBERNETES_FILES := $(addprefix $(SVC_ACCT_DIR)/,ca.crt token tls.crt tls.key)
 .DEFAULT_GOAL    := build
 
-vendor:
+vendor: ## Install project dependencies
 	@glide install -v
 
-broker: $(SOURCES)
+broker: $(SOURCES) ## Build the broker
 	go build -i -ldflags="-s -w" ./cmd/broker
 
-build: broker
+build: broker ## Build the project
 	@echo > /dev/null
 
-run: broker | $(KUBERNETES_FILES)
+lint: ## Run golint
+	@golint -set_exit_status $(addsuffix /... , $(SOURCE_DIRS))
+
+fmt: ## Run go fmt
+	@gofmt -d $(SOURCES)
+
+fmtcheck: ## Check go formatting
+	@gofmt -l $(SOURCES) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
+
+test: ## Run unit tests
+	@go test ./pkg/...
+
+vet: ## Run go vet
+	@go tool vet ./cmd ./pkg
+
+pre-flight: fmtcheck vet lint build test ## Pre-flight checks before creating PR
+
+run: broker | $(KUBERNETES_FILES) ## Run the broker executable locally
 	@./scripts/run_local.sh ${BROKER_CONFIG}
 
 $(KUBERNETES_FILES):
 	@./scripts/prep_local_devel_env.sh
 
-prepare-local-env: $(KUBERNETES_FILES)
+prepare-local-env: $(KUBERNETES_FILES) ## Prepare the local environment for running the broker locally
 	@echo > /dev/null
 
-prepare-build-image: build
+build-image: ## Build a docker image with the broker binary
 	cp broker build/broker
-
-build-image: prepare-build-image
 	docker build -f ${BUILD_DIR}/Dockerfile-src -t ${BROKER_IMAGE}:${TAG} ${BUILD_DIR}
 	@echo
 	@echo "Remember you need to push your image before calling make deploy"
@@ -47,24 +63,30 @@ release: release-image
 
 push:
 	docker push ${BROKER_IMAGE}:${TAG}
-clean:
+
+clean: ## Clean up your working environment
 	@rm -f broker
 	@rm -f build/broker
 
-really-clean: clean
+really-clean: clean ## Really clean up the working environment
 	@rm -f $(KUBERNETES_FILES)
 	cleanup-ci
 
-deploy:
+deploy: ## Deploy a built broker docker image to a running cluster
 	@./scripts/deploy.sh ${BROKER_IMAGE}:${TAG} ${REGISTRY} ${ORG}
 
-test:
-	go test ./pkg/...
-
-cleanup-ci:
+cleanup-ci: ## Cleanup after ci run
 	./scripts/broker-ci/cleanup-ci.sh
 
-ci:
+ci: ## Run CI (fmtcheck, vet, lint, build, test)
 	./scripts/broker-ci/local-ci.sh
 
-.PHONY: vendor build run prepare-build-image build-image release-image release push clean deploy test ci cleanup-ci
+help: ## Show this help screen
+	@echo 'Usage: make <OPTIONS> ... <TARGETS>'
+	@echo ''
+	@echo 'Available targets are:'
+	@echo ''
+	@grep -E '^[ a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | \
+		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: run build-image release-image release push clean deploy ci cleanup-ci lint build vendor fmt fmtcheck test vet help
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -8,9 +8,20 @@
 action=$1
 
 export GLIDE_TARBALL="https://github.com/Masterminds/glide/releases/download/v0.12.3/glide-v0.12.3-linux-amd64.tar.gz"
-export PROJECT_ROOT=$GOPATH/src/github.com/openshift/ansible-service-broker
 
-if [[ "$action" == "install" ]]; then
+if [[ "$action" == "before_install" ]]; then
+  echo "================================="
+  echo "        Before Install           "
+  echo "================================="
+  sudo apt-get -qq update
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -58,13 +75,21 @@ really-clean: clean
...
 	@./scripts/deploy.sh ${BROKER_IMAGE}:${TAG} ${REGISTRY} ${ORG}
 
-test:
-	go test ./pkg/...
-
 cleanup-ci:
 	./scripts/broker-ci/cleanup-ci.sh
 
-ci:
+ci: fmtcheck vet lint build test ## Run CI (fmtcheck, vet, lint, build, test)
```
rthallisey: ```make ci``` is meant to be both a command run by travis and locally. When run locally, it will do a quick sanity check to test your patch works.  I like the idea about moving these into the makefile, but I think they shouldn't be apart of ```make ci```. How about ```make travis```?

djzager: I admit that I wasn't super keen on adding all of those dependencies to `make ci` and I think that creating a `make travis` is a good alternative. I'll make that change once I get travis to function with my changes.

Makefile
```diff
@@ -5,20 +5,37 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
...
+vendor: ## Install project dependencies
 	@glide install -v
 
-broker: $(SOURCES)
-	go build -i -ldflags="-s -w" ./cmd/broker
+broker: $(SOURCES) ## Build the broker
+	@go build -i -ldflags="-s -w" ./cmd/broker
 
-build: broker
+build: vendor broker ## Build the project
```
jmrodri: Don't add vendor here. We don't need to update vendor since (sadly) it is checked into the repo.

Makefile
jmrodri: remove these dependencies from ci target.

Makefile
```diff
@@ -5,20 +5,37 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
 BUILD_DIR        = "${GOPATH}/src/github.com/openshift/ansible-service-broker/build"
 PREFIX           ?= /usr/local
 BROKER_CONFIG    ?= $(PWD)/etc/generated_local_development.yaml
+SOURCE_DIRS      = cmd pkg
```
jmrodri: that's fine.

.travis.yml
```diff
@@ -10,7 +10,24 @@ before_install:
   - sudo apt-get install -y python-apt autoconf pkg-config e2fslibs-dev libblkid-dev zlib1g-dev liblzo2-dev asciidoc
 install:
   - sudo pip install ansible==2.3.1
-  - ./scripts/travis.sh install
+  # dash? wtf is dash? UGH! use a real shell
+  - sudo rm /bin/sh
```
jmrodri: NACK the intent was not to remove the scripts/travis.sh script. That should remain. It should simply call the make targets for each of its steps.

.travis.yml
```diff
@@ -10,7 +10,24 @@ before_install:
   - sudo apt-get install -y python-apt autoconf pkg-config e2fslibs-dev libblkid-dev zlib1g-dev liblzo2-dev asciidoc
 install:
   - sudo pip install ansible==2.3.1
-  - ./scripts/travis.sh install
```
jmrodri: keep the script.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-#
-# travis.sh
-#
-# This script is used by travis to test the broker
-#
-action=$1
-
-export GLIDE_TARBALL="https://github.com/Masterminds/glide/releases/download/v0.12.3/glide-v0.12.3-linux-amd64.tar.gz"
-export PROJECT_ROOT=$GOPATH/src/github.com/openshift/ansible-service-broker
-
```
jmrodri: bring this script back

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-  sudo make install
-  cd $PROJECT_ROOT
-
-  # now install deps
-  wget -O /tmp/glide.tar.gz $GLIDE_TARBALL
-  tar xfv /tmp/glide.tar.gz -C /tmp
-  sudo mv $(find /tmp -name "glide") /usr/bin
-
-  # install golint
-  go get -u github.com/golang/lint/golint
```
jmrodri: Keep install section as is. I don't like these instructions being in travis.yml. Writing bash is hard enough, and couple that with yaml it's a pain in the ass. So let's keep all that here.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-  # to have lint fail build add -set_exit_status option
-  # to each golint command
-  if [ -n "$(golint $PROJECT_ROOT/cmd/...)" ]; then
-        golint $PROJECT_ROOT/cmd/...
-        exit 1
-  fi
-  if [ -n "$(golint $PROJECT_ROOT/pkg/...)" ]; then
-        golint $PROJECT_ROOT/pkg/...
-        exit 1
-  fi
```
jmrodri: Keep goling section EXCEPT instead of calling golint, just call `make lint`.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-elif [[ "$action" == "format" ]]; then
-  echo "================================="
-  echo "             Format              "
-  echo "================================="
-  if [ -n "$(gofmt -l $PROJECT_ROOT/cmd $PROJECT_ROOT/pkg)" ]; then
-      gofmt -d $PROJECT_ROOT/cmd $PROJECT_ROOT/pkg
-      exit 1
```
jmrodri: Keep format section EXCEPT instead of calling gofmt, just call `make format`.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-  fi
-elif [[ "$action" == "vet" ]]; then
-  echo "================================="
-  echo "              Vet                "
-  echo "================================="
-  go tool vet $PROJECT_ROOT/cmd $PROJECT_ROOT/pkg
-  exit $?
```
jmrodri: Keep vet section EXCEPT instead of calling go vet, just call `make vet`.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-elif [[ "$action" == "build" ]]; then
-  echo "================================="
-  echo "             Build               "
-  echo "================================="
-  make vendor
-  make build
```
jmrodri: Keep build section EXCEPT instead of calling make vendor and make build, just call `make build`.

scripts/travis.sh
```diff
@@ -1,87 +0,0 @@
...
-  echo "================================="
-  echo "              Test               "
-  echo "================================="
-  make test
-elif [[ "$action" == "ci" ]]; then
-  echo "================================="
-  echo "            Broker CI            "
-  echo "================================="
-  ./scripts/broker-ci/setup.sh
-fi
```
jmrodri: leave the above test and ci as is.

Makefile
jmrodri: @djzager skip the `make travis` more than likely I'm going to run one of the targets to see if it failed travis. i.e. lint or format may have failed, I'll just run make lint or make format fix the problem and repush.

scripts/travis.sh
djzager: After a few failed attempts at writing bash in yaml...I now appreciate what `scripts/travis.sh` was doing.

scripts/travis.sh
djzager: This should be addressed with latest push.

djzager:
Once I have positive feedback on the proposed changes. I'll make the appropriate changes to the supporting documentation. Although, with respect to make targets we should now be able to point users to `make help` and simply document all of our targets in the Makefile.

Makefile
```diff
@@ -5,20 +5,36 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
...
-	go build -i -ldflags="-s -w" ./cmd/broker
+broker: $(SOURCES) ## Build the broker
+	@go build -i -ldflags="-s -w" ./cmd/broker
 
-build: broker
+build: broker ## Build the project
 	@echo > /dev/null
 
+lint: ## Run golint
+	@golint -set_exit_status $(SOURCE_DIRS)
```
jmrodri: This doesn't work.

```
[jesusr@speed3 ansible-service-broker{djzager-simple-make}]$ make lint
read cmd: is a directory
read pkg: is a directory
[jesusr@speed3 ansible-service-broker{djzager-simple-make}]$ 
```

If you looked at the travis.sh before golint was run separately:

```
-  # to have lint fail build add -set_exit_status option
-  # to each golint command
-  if [ -n "$(golint $PROJECT_ROOT/cmd/...)" ]; then
-        golint $PROJECT_ROOT/cmd/...
-        exit 1
-  fi
-  if [ -n "$(golint $PROJECT_ROOT/pkg/...)" ]; then
-        golint $PROJECT_ROOT/pkg/...
-        exit 1
-  fi
```

you have to call golint twice once for each package and it MUST contain the `...` after the package name.

Makefile
```diff
@@ -5,20 +5,36 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
...
+	@gofmt -l -w $(SOURCE_DIRS)
+
+fmtcheck: ## Check go formatting
+	@gofmt -l $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
+
+test: ## Run unit tests
+	@go test ./pkg/...
+
+vet: ## Run go vet
+	@go tool vet ./cmd ./pkg
```
jmrodri: 👍 go vet works as expected.

Makefile
```diff
@@ -5,20 +5,36 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
...
+	@golint -set_exit_status $(SOURCE_DIRS)
+
+fmt: ## Run go fmt
+	@gofmt -l -w $(SOURCE_DIRS)
+
+fmtcheck: ## Check go formatting
+	@gofmt -l $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
+
+test: ## Run unit tests
+	@go test ./pkg/...
```
jmrodri: go test still works as well.

Makefile
```diff
@@ -5,20 +5,36 @@ BROKER_IMAGE     ?= $(REGISTRY)/$(ORG)/ansible-service-broker
...
 	@echo > /dev/null
 
+lint: ## Run golint
+	@golint -set_exit_status $(SOURCE_DIRS)
+
+fmt: ## Run go fmt
+	@gofmt -l -w $(SOURCE_DIRS)
+
+fmtcheck: ## Check go formatting
+	@gofmt -l $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
```
jmrodri: I'm not sure how I feel about the fmtcheck + fmt change. The old fmt would simply dump the diff with the error giving you a chance to fix it. But I guess we just want to fix it. It still would be nice to see the diff.  I'll let everyone else chime in:

@shawn-hurley @eriknelson @rthallisey any opinion on make fmt actually fixing the errors automatically? If you guys are fine with it, we can leave it alone.

Makefile
```diff
@@ -58,13 +74,18 @@ really-clean: clean
...
 	./scripts/broker-ci/cleanup-ci.sh
 
-ci:
+ci: ## Run CI (fmtcheck, vet, lint, build, test)
 	./scripts/broker-ci/local-ci.sh
 
-.PHONY: vendor build run prepare-build-image build-image release-image release push clean deploy test ci cleanup-ci
+help: ## Show this help screen
+	@echo 'Usage: make <OPTIONS> ... <TARGETS>'
+	@echo ''
```
jmrodri: I ❤️ the help target.

Makefile
djzager: I agree this needs to be fixed...I want to know though why I can't reproduce it. I'll find you when I make it into the office...and then fix it.

Makefile
eriknelson: IMO seeing the diff would be nice. To be honest, I almost never see this or use this because my editor is configured to do it for me, so I don't have a strong opinion.

rthallisey: I never use this since I have my IDE handles all the formatting for me. If you like it the way it is, then keep it. Either way works for me.

jmrodri: @djzager sounds like it won't impact most folks. So it will only be relevant to anyone that happens to cause a failure in travis. So this is ok.

shawn-hurley: I am fine with it automatically fixing format as well

djzager: I actually agree with the feedback. If you want to specifically run `go fmt` on a file and have it modify the file for you, which most are doing through vim plugins or IDEs, then you are more than welcome to do that.

I made it so `make fmt` will simply report the errors of your ways and `make fmtcheck` will simply report the file causing the failure and `exit(1)`.

djzager:
[Documentation Updates](https://trello.com/c/LIxva4xo/434-documentation-updates)

PR #354, opened 2017-08-10
Add local etcd support for local env

eriknelson:
Most broker developers don't use the in cluster etcd instance and instead prefer an instance running locally so etcdctl and other tools can be used to interact with raw etcd. This PR introduces a proper configuration param so we can stop hacking up the generated config.

 - Set LOCAL_ETCD="true" in `my_local_dev_vars` to have `prep_local` script configure for a local etcd instance running on the developer's host.

### DIFF
--- a/scripts/prep_local_devel_env.sh
+++ b/scripts/prep_local_devel_env.sh
@@ -180,8 +180,16 @@ echo "Sleeping for a few seconds to avoid issues with broker not being able to t
 echo "Appears like there is a delay of when we create the asb-etcd route and when it is available for use"
 sleep 5
 
-etcd_route=`oc get route asb-etcd -n ${ASB_PROJECT} -o=jsonpath=\'\{.spec.host\}\'`
-echo "etcd route is at: ${etcd_route}"
+if [ "$LOCAL_ETCD" == "true" ]; then
+  ETCD_ROUTE="localhost"
+  ETCD_PORT=2379
+else
+  ETCD_ROUTE=`oc get route asb-etcd -n ${ASB_PROJECT} -o=jsonpath=\'\{.spec.host\}\'`
+  ETCD_PORT=80
+fi
+
+echo "etcd route is at: ${ETCD_ROUTE}"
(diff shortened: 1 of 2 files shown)


scripts/prep_local_devel_env.sh
```diff
@@ -180,8 +180,16 @@ echo "Sleeping for a few seconds to avoid issues with broker not being able to t
 echo "Appears like there is a delay of when we create the asb-etcd route and when it is available for use"
 sleep 5
 
-etcd_route=`oc get route asb-etcd -n ${ASB_PROJECT} -o=jsonpath=\'\{.spec.host\}\'`
-echo "etcd route is at: ${etcd_route}"
+if [ -z "$LOCAL_ETCD" ]; then
```
jmrodri: This doesn't do what you want.
```
      -z string
              True if the length of string is zero.
```

```
LOCAL_ETCD ="false"
LOCAL_ETCD="true"
```
will both result in using localhost.

The only way to use the `oc get route` version is if LOCAL_ETCD is unset entirely.


### Empty
```
$ ./z.sh          
LOCAL_ETCD =                                  
using oc get route and port 80                
```
### set to true
```
$ ./z.sh true     
LOCAL_ETCD = true                             
using localhost etcd on 2379                  

### set to false
```
$ ./z.sh false    
LOCAL_ETCD = false                            
using localhost etcd on 2379 
```

Comment (jmrodri) [inline scripts/prep_local_devel_env.sh]: here is my z.sh script I used for testing the above:
```
LOCAL_ETCD="$1"
  
echo "LOCAL_ETCD = $LOCAL_ETCD"
if [ -z "$LOCAL_ETCD" ]; then
    echo "using oc get route and port 80"
else
    echo "using localhost etcd on 2379"
fi
```

jmrodri: I think you want `if [ "$LOCAL_ETCD" == "true" ]; ` and flip the conditions only run the local one if it is set to true. else do the `oc get`

PR #357, opened 2017-08-10
Allow specifying a tag for apbs

jmontleon:
Specify alternate tags for APB's. In order to more easily test canary/nightly apb images.

 - Add a 'tag' config option for registries in the config.yaml. The template will default it to latest.

depends-on <PR>
There's a half way done PR for catasb that will make it easier to override the tag.

fixes <issue_number>
N/A

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -240,8 +240,11 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 	for _, spec := range specs {
 		//need to make / a hyphen to allow for global uniqueness but still match spec.
-		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, spec.Image),
+
+		imageName := strings.Replace(spec.Image, ":", "-", -1)
+		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, imageName),
 			"/", "-", -1)
+		spec.FQName = fmt.Sprintf("%.51v", spec.FQName)
 
 		// ID Will be a md5 hash of the fully qualified spec name.
 		hasher := md5.New()
--- a/pkg/registries/adapters/adapter.go
+++ b/pkg/registries/adapters/adapter.go
@@ -3,6 +3,7 @@ package adapters
 import (
 	b64 "encoding/base64"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 
@@ -34,13 +35,13 @@ type Configuration struct {
 	Pass   string
 	Org    string
 	Images []string
+	Tag    string
 }
 
 // Retrieve the spec from a registry manifest request
-func imageToSpec(log *logging.Logger, req *http.Request) (*apb.Spec, error) {
+func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Spec, error) {
 	log.Debug("Registry::imageToSpec")
 	spec := &apb.Spec{}
-
 	req.Header.Add("Accept", "application/json")
 
 	resp, err := http.DefaultClient.Do(req)
@@ -101,6 +102,8 @@ func imageToSpec(log *logging.Logger, req *http.Request) (*apb.Spec, error) {
 		log.Errorf("Something went wrong loading decoded spec yaml, %s", err)
 		return nil, err
 	}
+
+	spec.Image = fmt.Sprintf("%s:%s", spec.Image, apbtag)
 
 	log.Debugf("adapter::imageToSpec -> Got plans %+v", spec.Plans)
 	log.Debugf("Successfully converted Image %s into Spec", spec.Image)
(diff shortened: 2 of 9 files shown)


jmontleon:
Name is in the correct format as is Image
```
[2017-08-10T18:53:48.637Z] [NOTICE] Spec.Name: dh-ansibleplaybookbundle-rhscl-postgresql-apb-nightly
[2017-08-10T18:53:48.637Z] [NOTICE] Spec.Image: ansibleplaybookbundle/rhscl-postgresql-apb:nightly
```
Container deployed successfully using the nightly tag.
```
oc describe pod apb-096134ee-c36e-4e5d-a20c-073ae75bd440 | grep -w Image:
    Image:  ansibleplaybookbundle/rhscl-postgresql-apb:nightly
```
Also displayed fine in the catalog UI.

pkg/broker/broker.go
```diff
@@ -240,8 +240,8 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 	for _, spec := range specs {
 		//need to make / a hyphen to allow for global uniqueness but still match spec.
-		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, spec.Image),
-			"/", "-", -1)
+		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName,
+			strings.Replace(spec.Image, ":", "-", -1)), "/", "-", -1)
```
jmrodri: Normally I'm ok with the embedded calls but it took me quite a bit to read this line with the 2 string replaces and the sprintf. Let's pull out the inner replacement:

```golang
imageName := strings.Replace(spec.Image, ":", "-", -1)
spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, imageName), "/", "-", -1)
```

pkg/registries/adapters/adapter.go
```diff
@@ -102,6 +102,8 @@ func imageToSpec(log *logging.Logger, req *http.Request) (*apb.Spec, error) {
 		return nil, err
 	}
 
+	spec.Image = spec.Image + ":" + apbtag
```
jmrodri: I know we probably do the concat somewhere but most of the time I see the Sprintf format:

```golang
spec.Image = fmt.Sprintf("%s:%s", spec.Image, apbtag)
```

jmrodri: Also something that popped in my head, what if `apbtag` is empty? this will create an image like:
```
mediawiki-apb:
```
I don't think that's valid.

pkg/registries/adapters/rhcc_adapter.go
```diff
@@ -56,11 +56,11 @@ func (r RHCCAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, error) {
 	specs := []*apb.Spec{}
 	for _, imageName := range imageNames {
 		req, err := http.NewRequest("GET",
-			fmt.Sprintf("%v/v2/%v/manifests/latest", r.Config.URL.String(), imageName), nil)
+			fmt.Sprintf("%v/v2/%v/manifests/%v", r.Config.URL.String(), imageName, r.Config.Tag), nil)
```
jmrodri: I think there's a `Validate` on the Config to make sure that a value is entered for Tag. If no value is present default it to "latest". Otherwise, I can change the value in the deployment template to empty and we'll be appending `:` to the end of the image names with no tags. Not sure what will happen.

jmontleon: I had to go out of my way, by modifying the template to remove a default value for tag in order to see what happens. This looks to cover it in a case where either nothing or a empty string are passed in.

https://github.com/openshift/ansible-service-broker/pull/357/commits/044e566bcfa162f7ca5ee0763d84a541a7644f93#diff-54e88cebcb04cc49b8796b2f774cb6c7R249

Once I set the default there to "" it did break, and off into a crash loop it went..
```
[2017-08-11T13:34:37.589Z] [ERROR] Error grabbing JSON body from response: json: cannot unmarshal number into Go value of type struct { History []map[string]string "json:\"history\"" }
```

I made a change to the code to use latest when Tag comes back as "", which seemed to get the broker off and rolling.

jmontleon:
Now also makes sure if no other tag is specified in the config that latest is used.

I also found that on bind the name was being drawn out longer than 63 characters, so I added a line to cut it back to 51 if it exceeded that. The catalog looks to add a -xxxxx-xxxxx piece which will extended it to the maximum of 63 characters if we should reach 51.

jmontleon:
@jmrodri I made the changes you requested.

jmrodri:
@eriknelson can you take a look over this so we can get ack #2?

scripts/broker-ci/postgresql.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: postgresql
   namespace: default
 spec:
-  serviceClassName: dh-ansibleplaybookbundle-rhscl-postgresql-apb
+  serviceClassName: dh-ansibleplaybookbundle-rhscl-postgresql-apb-lates
```
shawn-hurley: TYPO: ```-lastest```

jmontleon: It's not a typo :)

The name can at most be 51 characters if the catalog is not to complain. When it is created ithrough the normal process it will be `dh-ansibleplaybookbundle-rhscl-postgresql-apb-lates` not `dh-ansibleplaybookbundle-rhscl-postgresql-apb-latest`.

If this says latest it will actually fail.
```echo -ne "dh-ansibleplaybookbundle-rhscl-postgresql-apb-lates" | wc -c
51
```

PR #360, opened 2017-08-11
Fix rebase mistake

rthallisey:
I wanted to duplicate ```if ! ${ERROR};``` logic where we only run
```make ci``` if the setup worked.. Instead, ```make ci```
only runs when the setup errors.

 - Fixes a false positive introduced from #348


jmontleon:
ACK

PR #368, opened 2017-08-15
Updates first-pass proposal

eriknelson:
Think this is a decent starting point, there is a lot to this that I think we'll discover as we move through implementation. I'll try to continue to update this document for anything major. IMO we'll probably be iterating on this a bunch.

### DIFF
--- a/docs/proposals/updates-first-pass.md
+++ b/docs/proposals/updates-first-pass.md
@@ -180,7 +180,7 @@ plans:
     parameters: *_p # param anchor
     updates_to:
       - gold
-  - name: silver
+  - name: gold
     description: A single DB server with persistent storage
     free: true
     metadata: # ...snip
--- a/docs/proposals/updates-first-pass.md
+++ b/docs/proposals/updates-first-pass.md
@@ -1,0 +1,205 @@
+# Initial APB Update support
+
+[OBS Update Spec](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#updating-a-service-instance)
+
+## Introduction
+
+Updates allow APB authors to mutate existing service instances.
+
+> By implementing this endpoint, service broker authors can enable users
+to modify two attributes of an existing service instance: the service plan
+and parameters. By changing the service plan, users can upgrade or downgrade
+their service instance to other plans. By modifying properties, users can change
+configuration options that are specific to a service or plan
+
+Therefore, the cases we're focused on are:
+
+1) Changing the argument used for a given parameter. Ex: change a "name"
+
+2) Moving from one plan to another.  Ex: Move from non-persistent plan to
+a persisted plan. Another example of this could be moving from a multi-tenant
+environment to a dedicated environment for a service's database.
+
+NOTE: The Service Catalog currently does **not** implement update, so triggering
+these from the UI is not possible. Updates should be considered very "alpha".
+To demonstrate , plan is to enhance `asbcli` for update support.
+
+Updates are a potentially very complex subject considering the full breadth
+of what that can mean for any given APB, so the desire here is to control scope
+for an initial first-pass of what this could be.
+
+**First-pass constraints**
+
+* Attempting to update a service that has outstanding bindings is invalid and
+will result in a 400. Concern here is that a param update or plan transition
+could have breaking/destructive consequences for a binding consumer. We think
+it's possible in the future to support this, but it deserves an independent
+proposal and implementation plan separate from the first-pass.
+
+## Target Demos
+
+### Basic 
+
+* Etherpad - demonstrate an update to initial pad title, foo -> bar
+* Etherpad - demonstrate an update from non-persistent plan to persistent plan
+* Etherpad - demonstrate an invalid plan update request fails with bad request
+
+### Stretch
+
+* Demo update of credentials, moving a postgresql
+database from plan "silver" to "gold" and issuing new admin credentials as part of
+the process
+* Demo migration of data during a plan transition.
+Ex: postgresql "silver" -> "gold" plan, migrating data from "silver" -> "gold"
+dbs
+* Demo example of a schema migration. Ex: web app moving
+from 1.0 -> 2.0 that brings its own DB, and needs to migrate the schema as
+part of that version update
+
+
+## Implementation Details
+
+### Broker
+
+As outlined in the spec, broker must support both async and sync update variants.
+Async is requested via `accepts_incomplete` qparam.
+
+Broker should prefer to delegate any details of what an "update" actually means
+to the APBs. That means passing all the `previous_values` it receives as part
+of the update request, along with the set of `next_values` (reading the upstream
+spec, the `next_values` are not wrapped in that key specifically). It allows
+for the APBs to define "this is how we will move from our current state, to the
+next". Example request body taken from upstream spec, let's refer to this as
+the `UpdatePayload`:
+
+```json
+{
+  "context": {
+    "platform": "cloudfoundry",
+    "some_field": "some-contextual-data"
+  },
+  "service_id": "service-guid-here",
+  "plan_id": "plan-guid-here",
+  "parameters": {
+    "parameter1": 1,
+    "parameter2": "foo"
+  },
+  "previous_values": {
+    "plan_id": "old-plan-guid-here",
+    "service_id": "service-guid-here",
+    "organization_id": "org-guid-here",
+    "space_id": "space-guid-here"
+  }
+}
+```
+
+After validating a given request, the broker will pass this payload through to
+an APB, triggering a new "update" action that should execute a new `update.yml`
+playbook. See APB section for more.
+
+#### New Credentials
+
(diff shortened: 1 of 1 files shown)


eriknelson:
[(13) Service Instance Update](https://trello.com/c/1YVfRFAZ/422-13-service-instance-update)

docs/proposals/updates-first-pass.md
```diff
@@ -0,0 +1,205 @@
...
+
+1) Changing the argument used for a given parameter. Ex: change a "name"
+
+2) Moving from one plan to another.  Ex: Move from non-persistent plan to
+a persisted plan. Another example of this could be moving from a multi-tenant
+environment to a dedicated environment for a service's database.
+
+NOTE: The Service Catalog currently does **not** implement update, so triggering
+these from the UI is not possible. Updates should be considered very "alpha".
+To demonstrate , plan is to enhance `asbcli` for update support.
```
eriknelson: Thank you, apparently I can't speak english.

eriknelson:
@jmrodri I was unsure of the field name on APBs to describe the plan graphs. I originally wanted to call it `transitions`, but I think `updates_to` is better from an author perspective. CC @cfchase @dymurray

docs/proposals/updates-first-pass.md
eriknelson: @jmrodri @rthallisey @shawn-hurley RE: `asbcli`

It's the butt of our jokes because it was a load of crap python I hammered out on a Friday afternoon for a Monday demo expecting it to die on Tuesday, but it's proven to be a very useful tool for broker authors, and I'm finding myself now wanting to implement update support. @rthallisey implemented bind support because he needed it. It's clear ASB authors want this kind of tool, we should work together to define what we're looking for out of it and make that happen. IMO, python remains perfect for this. I'm going to log an issue so we have a distinct place to air desires for what we want this tool to be, and let's support it given the collective interest.

eriknelson:
@jmrodri Addressed your comments :+1:.  I'm going to be moving off update to look into some upstream catalog things, so this should get merged with the expectation that an implementation may not happen for a sprint or two. This represents the initial design work and requirements for a first pass.

docs/proposals/updates-first-pass.md
```diff
@@ -0,0 +1,205 @@
...
+      - silver
+      - gold
+  - name: silver
+    description: Silver DB plan with persistence
+    free: true
+    metadata: # ...snip
+    parameters: *_p # param anchor
+    updates_to:
+      - gold
+  - name: silver
```
dymurray: Assuming this is supposed to say Gold?

dymurray:
@eriknelson +1 on `updates_to:`. 

This looks good to me. My only question is: Does the APB author need to write two paths for the Gold plan? How does the Gold plan know which plan he is transitioning from? I guess I'm a little uncertain how an APB author would control the logic for upgrading different versions.
e.g.
```
Dev->Gold goes from version 1.0 to 2.0 where there's a massive schema change / db migration
Silver->Gold has no schema change / db migration
```

I only bring this up in an instance where knowing the previous plan (state of the app) is necessary for the APB. I only see that information being passed around in the broker to manage state. 

Overall this looks great, ACK, just bringing up a concern. Sorry if its been worked through :)

docs/proposals/updates-first-pass.md
cfchase: `name: gold`

docs/proposals/updates-first-pass.md
```diff
@@ -0,0 +1,205 @@
...
+from the broker.
+
+**plan transitions**
+
+The OSB spec explicitly states the broker must reject plan transitions that are
+invalid. Effectively, this is a directed graph with transitions as edges and
+plans as the nodes. APBs will need to define this graph in their `apb.yml`
+so the broker is able to validate a requested transition.
+
+Proposal is to add an optional `updates_to` list on `Plan` objects that indicates
```
cfchase: What do you think about using `updates_from` (or `updates` or `can_update`) on the destination plan instead of the source plan.  It might give some additional flexibility and the logic for updating seems like it would reside in the destination plan.  Putting an `updates_to` on an old version requires that the old version knows all future versions that it would be able to update to.  That's workable on a single apb that only upgrades from one plan to another, but in future use cases, that might not be the case.  For example, by using `updates_from` you can use the same syntax to update previous APBs.

```yaml
# Postgres 9.5 APB
name: rhscl-postgresql-9.5-apb
# ...snip
plans:
  - name: dev
    description: A single DB server with no storage
    free: true
    metadata: # ...snip
    parameters: *_p
    updates_from:
      - version: 9.4
        plan: dev
  - name: silver
    description: Silver DB plan with persistence
    free: false
    metadata: # ...snip
    parameters: *_p # param anchor
    updates_from:
      - dev # defaults to same version and changes plan
      - version: 9.4
        plan: dev
      - version: 9.4
        plan: silver
  - name: gold
    description: A single DB server with persistent storage
    free: false
    metadata: # ...snip
    parameters: *_p # param anchor
    updates_from:
      - dev # defaults to same version and changes plan
      - silver # defaults to same version and changes plan
      - version: 9.4
        plan: dev
      - version: 9.4
        plan: silver
      - version: 9.4
        plan: gold
```

In this case, the gold plan knows what it can update from and is responsible for the logic for doing so.  We may not use this, but it'd be nice to have the option.  Also, it feels better putting the option on the plan that will know what to do with the info.

eriknelson:
@dymurray 

> How does the Gold plan know which plan he is transitioning from?

The `UpdatePayload` that gets passed into the APB with a `previous_values` section that contains the previous `plan_id`: 

```
{
  "context": {
    "platform": "cloudfoundry",
    "some_field": "some-contextual-data"
  },
  "service_id": "service-guid-here",
  "plan_id": "plan-guid-here",
  "parameters": {
    "parameter1": 1,
    "parameter2": "foo"
  },
  "previous_values": {
    "plan_id": "old-plan-guid-here",
    "service_id": "service-guid-here",
    "organization_id": "org-guid-here",
    "space_id": "space-guid-here"
  }
}
```

> Dev->Gold goes from version 1.0 to 2.0 where there's a massive schema change / db migration
Silver->Gold has no schema change / db migration

This is a much bigger question. The spec reads to me like plans and plan updates were originally designed for tiers of service (multi-tenant DB vs dedicated instance) rather than versions and their updates. I don't know that it makes sense to couple versions to plans, because then you start talking about having a `plan` for every permutation of `<tier>+<version>+...+<N-deployment-variant>`, and that starts to get ugly quickly. Version is probably more appropriate passed as a parameter, but that opens up some interesting (and potentially complex) scenarios:

I have an instance of `MyFooApp-v1.2.1` that holds an active binding to an instance of `PostgreSQL-v9.5`. Its database has schema `v1` with a bunch of important data in it. What happens when:

1) I update `MyFooApp` instance to `v2.0.0`, which requires a schema+data migration. More importantly, what happens when that fails?

2) I update `PostgreSQL-v9.5`to `v9.6`, does this impact `MyFooApp` at all? Again, what if it fails?

These are both fair game for the next pass at this that takes active bindings into account, IMO.

docs/proposals/updates-first-pass.md
```diff
@@ -0,0 +1,205 @@
...
+
+1) Changing the argument used for a given parameter. Ex: change a "name"
+
+2) Moving from one plan to another.  Ex: Move from non-persistent plan to
+a persisted plan. Another example of this could be moving from a multi-tenant
+environment to a dedicated environment for a service's database.
+
+NOTE: The Service Catalog currently does **not** implement update, so triggering
+these from the UI is not possible. Updates should be considered very "alpha".
+To demonstrate , the plan is to enhance `asbcli` to support update.
```
eriknelson: Update is a [distinct operation/request](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#updating-a-service-instance) the platform (service-catalog in this case) is expected to implement. It is not a re-provision. @jmrodri any idea on a timeline? I was not present for the meeting where this came up.

docs/proposals/updates-first-pass.md
```diff
@@ -0,0 +1,205 @@
...
+```yaml
+name: rhscl-postgresql-apb
+# ...snip
+plans:
+  - name: dev
+    description: A single DB server with no storage
+    free: true
+    metadata: # ...snip
+    parameters: *_p
+    updates_to:
```
eriknelson: We need *some* way for the APBs to express a valid plan transition graph. In the absence of any of these, that implicitly says to me "I don't support plan updates".

docs/proposals/updates-first-pass.md
eriknelson: ACK, like the flexibility `update_from` supports. It feels a little backwards in my mind but your arguments are solid.

WRT versions, I think we need to think through and possibly decouple versions from plans [for these reasons](https://github.com/openshift/ansible-service-broker/pull/368#issuecomment-323755251). At the very least, that sounds like a problem discussion in-scope for the next pass proposal.

PR #373, opened 2017-08-15
bearer token proposal

jmrodri:
Add bearer token auth to the service broker.

### DIFF
--- a/docs/proposals/bearer-token.md
+++ b/docs/proposals/bearer-token.md
@@ -1,0 +1,123 @@
+# Bearer token authentication proposal
+
+## Introduction
+
+The service catalog currently supports basic auth as well as bearer token
+authentication. The broker will need to be enhanced to support bearer token
+authentication.
+
+## Problem Description
+The broker will need to support bearer token authentication.
+
+## <Implementation Details>
+Bearer token authentication uses the `Authorization` header followed by `Bearer`
++ 1 space + a base64 encoded token, see [RFC 6750 Section 2.1] [1]
+
+The broker will have a new `BearerAuth` struct that implements the [Provider] [2]
+interface.
+
+A new `TokenService` interface that will validate tokens either via Bearer
+token, shared key, or against a external service, etc. Basically used to deal with
+non username/password authentication.
+
+This new adapter will handle Bearer token and possibly SSL certificate
+authentication. The original UserService would handle BasicAuth, OAuth and any
+other model where a `UserPrincipal` will be returned.
+
+
+```golang
+type TokenService interface {
+   Validate(token string) bool
+}
+```
+
+There will be specific implementations of the TokenService which will validate
+the token. One simple one will simply read the token from the filesystem.
+Another one could connect to a service to validate the token.
+
+
+```golang
+// FileTokenService - reads a token from a static file
+type FileTokenService struct {
+   tokenFile string
+}
+
+func (f FileTokenService) Validate(token string) bool {
+    // read the file from tokenFile
+    // compare the token from the file to the one given
+    savedToken, err := ioutil.ReadFile(tokenFile)
+    if err != nil {
+        return false
+    }
+
+    return token == savedToken
+}
+```
+
+If we have an external service to handle validation something like this would be
+used:
+
+```golang
+type SomeExternalService struct {
+    service SomeAuthService
+}
+
+func NewSomeExternalService(service SomeAuthService) SomeExternalService {
+    return SomeExternalService{service: service}
+}
+
+func (s SomeExternalService) Validate(token string) bool {
+    return s.service.Validate(token)
+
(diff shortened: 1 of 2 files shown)


docs/proposals/bearer-token.md
```diff
@@ -0,0 +1,123 @@
...
+ - Implement a `FileTokenService` in `file_token_service.go`
+ - Investigate hooking up with OpenShift auth server
+   - new implementation of the `TokenService`
+   - new configuration item in broker config
+   - probably new configuration to setup (maybe) the auth server on OpenShift
+
+Other items to consider for consistency sake, but not directly required for this
+proposal:
+
+ - Rename `UserServiceAdapter` to `UserService`
```
shawn-hurley: :+1:

docs/proposals/bearer-token.md
```diff
@@ -0,0 +1,123 @@
...
+    // Basically the idea is the AuthProviders calling Validate
+    // don't really care.
+}
+```
+
+Basic questions about the impact to Broker and APBs:
+
+ - How will the broker's behavior change?
+
+   The broker will now have 2 ways for authentication: basic auth and bearer
```
jmrodri: @rthallisey yep, I can expand on it. You can have all the auths enabled if you want. Basically, the middleware handler will pick the first one that accepts the right authentication header. So if you have basic auth and bearer token enabled. And send in a Bearer token header, the basic auth provider will ignore it since there is no basic auth header. The bearer token provider will pick it up. If we had an SSL auth provider as well, it will have gotten skipped because the bearer token provider already processed the authentication.

docs/proposals/bearer-token.md
```diff
@@ -0,0 +1,123 @@
...
+   token. This can continue to expand as we see the need.
+
+ - Will this change APBs?
+
+   The bearer token will have no affect on the APBs.
+
```
jmrodri: I'll update the proposal to answer your question, but both basic and bearer will be enabled in catasb but for `make run` all auth is currently disabled.

docs/proposals/bearer-token.md
```diff
@@ -0,0 +1,123 @@
...
+#### Issues
+
+ - What's the best way to configure different auths?
+
+   For example, I want to have a `BasicAuth` that uses the `FileUserService` and a
+   `BasicAuth` that uses `DBUserService`, a fictitious service that loads users from a
+   database. Today's configuration does not support specifying a service backend
+   to a particular `AuthProvider`. Thoughts?
```
jmrodri: @rthallisey Yes, `BasicAuth` is an `AuthProvider`. And `AuthProvider` can be associated with some backend service adapter. Today `BasicAuth` is configured to use `FileUserServiceAdapter`. But conceivably I'd like to configure `BasicAuth` to use a database of users instead something called `DBUserService`. I'm not sure how to represent that in our configuration.

```golang
func createProvider(providerType string, log *logging.Logger) (Provider, error) {
	switch strings.ToLower(providerType) {
	case "basic":
		log.Info("Configured for basic auth")
		usa, err := GetUserServiceAdapter(log)
		if err != nil {
			return nil, err
		}
		return NewBasicAuth(usa, log), nil
	// add case "oauth":
	default:
		panic("Unknown auth provider")
	}
}

// GetUserServiceAdapter returns the configured UserServiceAdapter
func GetUserServiceAdapter(log *logging.Logger) (UserServiceAdapter, error) {
	// TODO: really need to figure out a better way to define what
	// should be returned.
	return NewFileUserServiceAdapter("/var/run/asb-auth", log)
}
```

I could extend the configuration to also include a key for the service backend to use:

```yaml
  auth:
    - type: basic
      backend: file
      enabled: false
    - type: basic
      backend: db
      enabled: true
```

Or if we want to support multiple basic auths something like this:
```yaml
  auth:
    - type: basicfile
      enabled: true
    - type: basicdb
      enabled: true
```
each would be distinct keys in the list of supported values.

jmrodri:
@rthallisey I updated the docs to clarify the questions section, and I tried to answer your question about AuthProvider.

docs/proposals/bearer-token.md
```diff
@@ -0,0 +1,153 @@
...
+
+#### Issues
+
+ - What's the best way to configure different auths?
+
+For example, I want to have a `BasicAuth` that uses the `FileUserService` and a
+`BasicAuth` that uses `DBUserService`, a fictitious service that loads users from a
+database. Today's configuration does not support specifying a service backend
+to a particular `AuthProvider`. Thoughts?
+
```
rthallisey: Do we have a solution for this?

jmrodri: @rthallisey not yet, right now basicauth is tied to FileUserService and bearer auth would be tied to FileTokenService. Just need to decide how we want this in the config file.

Issue #181, opened 2017-06-14
Proposal: Block out any passwords outputed to the logs


rthallisey:
It would be nice to print the parameters from apbs to stdout. But, we can't do that unless we blockout passwords.

eriknelson:
+1, was just going to log this issue WRT to @jmrodri's feedback. I'm not sure how we can know which parameters are sensitive from the broker since that's really the domain of an APB. Could make this configurable maybe to only output params in debug mode?

Just also of note, the extra-vars are available if you describe the apb pod.

eriknelson:
#151 merged.

eriknelson:
Gah, sorry. Thought this was tracking that PR. This should remain open.

shawn-hurley:
## Proposal

I think that the easiest way to solve this is to push the logic of determining the security of a parameter to the client. I think this means adding to the APB Spec a parameter that is either ```password```  or ```secure```(I think this makes more sense in case it is an API key or another secret).  

#### ASB Changes

Add to the ParameterDescriptor a field.
When printing out Parameters, will need to match parameter name with the spec name and determine if it is secret or not. 

#### APB Spec Changes

Add an optional parameter described above, that will default to not private. 

Just a thought that will not have us trying to infer if a parameter should be private or not

PR #376, opened 2017-08-16
UI form metadata proposal

cfchase:
Proposal to support order and groups in the front end forms based on apb information.  Display type is free along for the ride.

### DIFF
--- a/docs/proposals/form-metadata.md
+++ b/docs/proposals/form-metadata.md
@@ -1,0 +1,129 @@
+# Add UI metadata to catalog requests
+
+## Introduction
+Support better provision user experience for APBs by giving additional metadata about APB parameters.
+
+## Problem Description
+Currently, APB parameters are shown in random order and without any kind of grouping.  the OpenServiceBroker API utilizes json-schema data format and does not include any UI metadata.  As such, the UI cannot customize the experience and users are confronted with a bad experience which gets worse as the number of parameters increases. 
+
+## Implementation Details
+
+### APB changes
+APBs will allow an author to implement parameter order, grouping, and form field type.
+* Ordering will honor the order of parameters used in the apb.yml
+* Grouping will require the addition of a parameter field `display_group`.  This is an arbitrary string value but can be left empty to signify that the form input is at the top level and not grouped.  If it matches other `display_group` string from a parameter immediately before or after the parameter in question, they would be grouped together.
+* UI field types will require the addition of a parameter field `display_type`.  Default (empty) will use the default form input based on the json-schema.
+
+```yaml
+name: hello-world-db-apb
+# ...snip...
+plans:
+  - name: default
+    # ...snip...
+    parameters:                           # respects order of parameters
+      - name: postgresql_database         # first parameter on form
+        title: PostgreSQL Database Name
+        type: string
+        required: True
+        default: admin
+      - name: postgresql_user             # second parameter
+        title: PostgreSQL User
+        type: string
+        required: True
+        default: admin
+        display_group: User Information   # same group as password       
+      - name: postgresql_password         # third parameter
+        title: PostgreSQL Password
+        type: string
+        required: True
+        default: admin
+        display_type: password            # password type displayed in UI as ****
+        display_group: User Information   # same group as user
+```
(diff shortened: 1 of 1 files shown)


docs/proposals/form-metadata.md
```diff
@@ -0,0 +1,127 @@
...
+        title: PostgreSQL Database Name
+        type: string
+        required: True
+        default: admin
+      - name: postgresql_user             # second parameter
+        title: PostgreSQL User
+        type: string
+        required: True
+        default: admin
+        display_group: User Information   # same group as password
```
rthallisey: What's the default value of display_group? Can display group be an arbitrary string or are there categories like User Information, Bind Information, Database Information, ect...?

docs/proposals/form-metadata.md
```diff
@@ -0,0 +1,127 @@
...
+      }
+    ]
+  }
+]
+
+```
+
+## Notes
+* This will require changes to [origin-web-catalog](https://github.com/openshift/origin-web-catalog) to read, validate, and use the form definition.  Currently we are only planning on supporting only fieldset/sections for grouping, types, and honoring the order from the form definition.
+
```
rthallisey: So far, I think the options in apb.yaml have been straightforward.  I think these options warrant us to create a document explaining all the possible options that can go in apb.yaml if it doesn't already exist.

Action-item: docs

docs/proposals/form-metadata.md
cfchase: Default value of display group would be empty, that would signify that the item is at the top level and not grouped.

docs/proposals/form-metadata.md
cfchase: I added a note to the APB changes section about adding documentation

docs/proposals/form-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+        type: string
+        required: True
+        default: admin
+        display_type: password            # password type displayed in UI as ****
+        display_group: User Information   # same group as user
+```
+Documentation for the additional fields should be added to the appropriate documentation for APBs in the [ansible-playbook-bundle](https://github.com/fusor/ansible-playbook-bundle) project.
+
+### API changes
+The agreed upon request format with the OpenShift UI team matches the [form definition](https://github.com/json-schema-form/angular-schema-form/blob/development/docs/index.md#form-definitions) used on the front end UI.  This will be specific to OpenShift but here is no standard agreed upon and this would cause the least changes in the UI code with maximum flexibility.  In order to minimize impact and not make changes to the OpenServiceBroker API, the data will be contained in the plan metadata.  Since parameters are tied to the method, the json structure in metadata containing the form definition will mirror the json structure containing the schema.
```
shawn-hurley: >This will be specific to OpenShift but here is no standard agreed upon 

Typo:here -> there

docs/proposals/form-metadata.md
```diff
@@ -0,0 +1,128 @@
+# Add UI metadata to catalog requests
+
+## Introduction
+Support better provision user experience for APBs by giving additional metadata about APB parameters.
+
+## Problem Description
+Currently, APB parameters are shown in random order and without any kind of grouping.  the OpenServiceBroker API utilizes json-schema data format and does not include any UI metadata.  As such, the UI cannot customize the experience and users are confronted with a bad experience which gets worse as the number of parameters increases. 
+
+## Implementation Details
```
shawn-hurley: Is this WIP? 

I don't think this change will impact a lot of the broker, but a quick rundown of where changes may need to be made might be nice.

docs/proposals/form-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+      }
+    ]
+  }
+]
+
+```
+
+## Notes
+* This will require changes to [origin-web-catalog](https://github.com/openshift/origin-web-catalog) to read, validate, and use the form definition.  Currently we are only planning on supporting only fieldset/sections for grouping, types, and honoring the order from the form definition.
+
```
shawn-hurley: I think that this can also take care of issue #181 which might be a nice additional feature to adding these params.

docs/proposals/form-metadata.md
cfchase: done... should only be in one place

PR #386, opened 2017-08-21
Added UI form information to metadata fields for parsing by OpenShift

cfchase:
Added UI form information to metadata fields for parsing by OpenShift UI.

https://github.com/openshift/ansible-service-broker/blob/master/docs/proposals/form-metadata.md

 - Add json regarding parameter order, groups, and display type to metadata in the json-schema-form definition.
 - Unit tests

### DIFF
--- a/pkg/broker/util.go
+++ b/pkg/broker/util.go
@@ -33,7 +33,7 @@ func toBrokerPlans(apbPlans []apb.Plan) []Plan {
 			ID:          plan.Name,
 			Name:        plan.Name,
 			Description: plan.Description,
-			Metadata:    plan.Metadata,
+			Metadata:    updateMetadata(plan.Metadata, plan.Parameters),
 			Free:        plan.Free,
 			Bindable:    plan.Bindable,
 			Schemas:     parametersToSchema(plan.Parameters),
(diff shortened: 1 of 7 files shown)


cfchase:
[(8) Parameter Ordering and Grouping](https://trello.com/c/Orp4rGpI/441-8-parameter-ordering-and-grouping)

cfchase:
https://github.com/fusor/ansible-playbook-bundle/pull/99

pkg/broker/util.go
```diff
@@ -43,6 +43,82 @@ func toBrokerPlans(apbPlans []apb.Plan) []Plan {
...
+		return metadata
+	}
+
+	if metadata == nil {
+		metadata = make(map[string]interface{})
+	}
+
+	formDefinition := make([]interface{}, 0)
+
+	for paramIdx := 0; paramIdx < len(params); {
```
shawn-hurley: Typo: paramIdx -> paramIDx. 

Also this I think it should look like
```go
for paramIDx, pd := range params {
...
}
```

pkg/broker/util.go
```diff
@@ -43,6 +43,82 @@ func toBrokerPlans(apbPlans []apb.Plan) []Plan {
...
+		var formItem interface{}
+		pd := params[paramIdx]
+		if pd.DisplayGroup == "" {
+			formItem, paramIdx = createUIFormItem(pd, paramIdx)
```
shawn-hurley: Because the loop above change will hit every item I think we can remove the side effect of increasing the `paramIdx`.

pkg/broker/util.go
```diff
@@ -43,6 +43,82 @@ func toBrokerPlans(apbPlans []apb.Plan) []Plan {
...
+		pd := params[paramIndex]
+		if pd.DisplayGroup != groupName {
+			break
+		}
+
+		formItem, paramIndex = createUIFormItem(pd, paramIndex)
+		items = append(items, formItem)
+	}
+
+	group := make(map[string]interface{})
```
shawn-hurley: Do we ever see adding to the group? if we do, I would prefer this becomes a struct. If it is always going to be two values like this, then I don't see a problem with it.

pkg/registries/adapters/openshift_adapter.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	b64 "encoding/base64"
+
```
shawn-hurley: 👍 on fixing the imports

pkg/broker/util.go
cfchase: Parameter Index -> paramIdx, not sure how capitalizing the X makes sense....

cfchase: Also, doing it as a range, because I had to alternate between the top level list and the inner list in the group was super ugly.

pkg/broker/util.go
cfchase: The reason I'm modifying the paramIdx, is that the two functions consume a differing amount of items from the list.  If it was a formItem, it consumed 1 item.  If it was a group, it was an arbitrary number of items that matched.

pkg/broker/util.go
cfchase: I'll create a FormItem type to make this clearer.

pkg/broker/util.go
shawn-hurley: I dont like the indirection of the method having the side effect of increasing the param for the loop. I also dont like that the group will increase the loop. This seems like it could be pretty tricky to debug IMO. 
Could we do this instead?
```go 
       group := map[string]interface{}{}
       groupName := ""
	for paramIDx, pd := range params {  
               	formItem, paramIdx = createUIFormItem(pd, paramIdx)
                if pd.DisplayGroup == "" && len(group) > 0 {
                     //save group to list and reset the group. 
                    // save new item to list
                } else if pd.DisplayGroup != groupName {
                    //Save group to list
                   // reset group
                   // add form item
               } else {
                  // add item to list
              }
	}
```
This just seems more clear to me, and I think (please correct me if I am wrong) will have the same behavior. @jmrodri @eriknelson @rthallisey Do you guys have strong opinions?

cfchase: As an alternative to returning a new paramIdx, those functions could return the number of items consumed and the loop could add it to the paramIdx
```go
if pd.DisplayGroup == "" {
  formItem, numItems = createUIFormItem(pd, paramIdx)
} else {
  formItem, numItems = createUIFormGroup(params, pd.DisplayGroup, paramIdx)
}
paramIdx = paramIdx + numItems
```

shawn-hurley: Yeah, that does make it more clear to me what is going on to me, thanks.

PR #390, opened 2017-08-23
Label APBs with their FQNames

eriknelson:
Adds some additional identifying metadata in the form of a label to help users identify what services apb pods are deploying.


shawn-hurley:
ACK

PR #397, opened 2017-08-24
Fix spelling in logs

djzager:

I noticed when running the local-ci script that there were spelling
mistakes in both the local-ci script and the logs being dumped for the
broker. As going through all files would be a long process, this change
addresses all the issues found in the two files I noticed had issues
from the logs.

N/A

N/A


shawn-hurley:
ACK

PR #404, opened 2017-08-28
ignore the broker only at the root

jmrodri:
The ag search tool respects the .gitignore file but having broker alone
was too aggressive for the tool. So by adding the / it tells git to
ignore broker only at the project root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,9 +31,9 @@ etc/prod.config.yaml
 etc/mock.config.yaml
 
 #Compiled binaries
-broker
+/broker
 build/broker
-!pkg/broker
+#!pkg/broker
 
 # python artifacts
 .python-version


jmrodri:
Before:

```
$ /usr/bin/ag --ignore-dir vendor "^\/\/[A-Za-z]"
pkg/handler/handler.go
419://printRequest - will print the request with the body.

pkg/app/app.go
36://CreateApp - Creates the application

pkg/apb/types.go
13://SpecManifest - Spec ID to Spec manifest
```
After

```
$ /usr/bin/ag --ignore-dir vendor "^\/\/[A-Za-z]"
pkg/handler/handler.go
419://printRequest - will print the request with the body.

pkg/app/app.go
36://CreateApp - Creates the application

pkg/apb/types.go
13://SpecManifest - Spec ID to Spec manifest

pkg/broker/types.go
183://BindResponse - Response for a bind

pkg/broker/broker.go
120://Login - Will login the openshift user.
870://AddSpec - adding the spec to the catalog for local development
```

With the previous .gitignore ag would never look in the broker packages for anything. Now it does. Also, you can add files to broker directory still and they are not ignored which was the reason for the !pkg entry before.

.gitignore
```diff
@@ -31,9 +31,9 @@ etc/prod.config.yaml
 etc/mock.config.yaml
 
 #Compiled binaries
-broker
+/broker
 build/broker
-!pkg/broker
+#!pkg/broker
```
shawn-hurley: Should we just delete this line?

jmrodri: shit um yeah. I'll do it in another PR since I just merged this one before seeing your comment.

PR #407, opened 2017-08-28
Proposals to make configuration easier to use.

shawn-hurley:
Proposals to determine how to we will change configuration values to make them easier to use.

### DIFF
--- a/docs/proposals/configuration_changes.md
+++ b/docs/proposals/configuration_changes.md
@@ -3,6 +3,21 @@
 ### Goal
 Create a package for creating and managing configuration values as a value store. This will be used to remove the necessity of each package to be loaded just for configuration values. The other benefit is that this package could be a more robust solution to the configuration, Allowing us to do a better job of watching for updates to the configmap and managing subscriptions to those changes. This value store should also be usable by other parts of the application and therefore will require less usage of hard coded configuration structures and will allow the addition or removal of configuration values to not create such an issue. 
 
+### Potential Library to user: [Viper](https://github.com/spf13/viper)
+Seems to be the most like what we were thinking and is used already in some large projects.
+#### Pros
+* Interface seems to be the same idea of what we wanted.
+* Sub maps that can be retrieved
+* Can read from a YAML
+* Optionally can set file watchers, if we eventually want to. 
+* Under active development.
+* No new package needed.
+
+#### Cons
+* More permissive interface.
+* Another library that we will need to the vendor and keep up to date.
+* May be a little overkill for our use case.
+
 ### Work Items
 * Create a specific package for managing configuration 
 * Create a package mutex map to be used to store configuration values.
@@ -13,8 +28,8 @@ m["registry"] = [map["name": "dh"...], map["name": "play"...]]
 ```
 * Create a syntax, and document, for retrieving a single configuration value from the list. I.E `config.MustGetBoolConfig("broker.dev_broker")`
 * Create a public API that will give the caller multiple ways to retrieve data.
-    - `MustGet` -> returns a single value and will panic if not able to retrieve the data
-    - `MustGet/Get<Type>` -> Allows for typed returns so the caller does not have to worry about type casting.
+    - `Get<Type>(key)` -> Allows for typed returns so the caller does not have to worry about type casting. The default for the type will be returned if the value is not found.
+    - `Put<Type>(key, value)` - Allows for programmatic setting of configuration values.
     - Ability to specify a `map[string]interface{}` as a return value so one could say `config.MustGetMapConfig("broker")` and would retrieve the map of values for the broker configuration section. 
 * Update the all of the initialization code to take a map[string]interface{} to create the structures with the correct values. At this point, we can use the "kind" argument for the adapters in the registry addressing [this issue](https://github.com/openshift/ansible-service-broker/issues/49).
 * Allows for the initialization code to print warnings if configuration values are not present, taking care of [this issue](https://github.com/openshift/ansible-service-broker/issues/270).
--- a/docs/proposals/configuration_changes.md
+++ b/docs/proposals/configuration_changes.md
@@ -18,6 +18,11 @@ Seems to be the most like what we were thinking and is used already in some larg
 * Another library that we will need to the vendor and keep up to date.
 * May be a little overkill for our use case.
 
+**We have decided that the we will create our own package. The reasons are:** 
+1. **It is not that much code to write**
+2. **We will have more control**
+3. **Not another dependency with its dependencies to manage and will therefore make things cleaner**
+
 ### Work Items
 * Create a specific package for managing configuration 
 * Create a package mutex map to be used to store configuration values.
--- a/docs/proposals/configuration_changes.md
+++ b/docs/proposals/configuration_changes.md
@@ -1,0 +1,71 @@
+## Configuration Changes
+
+### Goal
+Create a package for creating and managing configuration values as a value store. This will be used to remove the necessity of each package to be loaded just for configuration values. The other benefit is that this package could be a more robust solution to the configuration, Allowing us to do a better job of watching for updates to the configmap and managing subscriptions to those changes. This value store should also be usable by other parts of the application and therefore will require less usage of hard coded configuration structures and will allow the addition or removal of configuration values to not create such an issue. 
+
+### Work Items
+* Create a specific package for managing configuration 
+* Create a package mutex map to be used to store configuration values.
+* Take structured yaml and create a map of map of map.... saving the key as the name of the config value
+```go
+m["broker"] = map["dev_broker": true...]
+m["registry"] = [map["name": "dh"...], map["name": "play"...]]
+```
+* Create a syntax, and document, for retrieving a single configuration value from the list. I.E `config.MustGetBoolConfig("broker.dev_broker")`
+* Create a public API that will give the caller multiple ways to retrieve data.
+    - `MustGet` -> returns a single value and will panic if not able to retrieve the data
+    - `MustGet/Get<Type>` -> Allows for typed returns so the caller does not have to worry about type casting.
+    - Ability to specify a `map[string]interface{}` as a return value so one could say `config.MustGetMapConfig("broker")` and would retrieve the map of values for the broker configuration section. 
+* Update the all of the initialization code to take a map[string]interface{} to create the structures with the correct values. At this point, we can use the "kind" argument for the adapters in the registry addressing [this issue](https://github.com/openshift/ansible-service-broker/issues/49).
+* Allows for the initialization code to print warnings if configuration values are not present, taking care of [this issue](https://github.com/openshift/ansible-service-broker/issues/270).
+* This will also allow us to use the per package logger that we have decided on.
+
+Go Map of generated config
+```go
+map[
+    registry:[
+        map[
+            type:dockerhub 
+            name:dh 
+            url:https://registry.hub.docker.com 
+            user:shurley 
+            pass:testingboom 
+            org:shurley
+        ], 
+        map[
+            pass:testingboom 
+            org:ansibleplaybookbundle 
+            type:dockerhub 
+            name:play 
+            url:https://registry.hub.docker.com 
+            user:shurley
+        ]
+    ] 
+    dao:map[
+        etcd_host:asb-etcd-ansible-service-broker.172.17.0.1.nip.io 
+        etcd_port:80
+    ] 
+    log:map[
+        logfile:/tmp/ansible-service-broker-asb.log 
+        stdout:true 
+        level:debug 
+        color:true
+    ] 
+    openshift:map[
+        host:172.17.0.1 
+        bearer_token_file:<nil> 
+        ca_file:<nil> 
+        image_pull_policy:<nil>
+    ] 
+    broker:map[
+        launch_apb_on_bind:false 
+        bootstrap_on_startup:true 
+        recovery:true 
+        output_request:true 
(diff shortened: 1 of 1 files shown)


docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
+## Configuration Changes
+
+### Goal
+Create a package for creating and managing configuration values as a value store. This will be used to remove the necessity of each package to be loaded just for configuration values. The other benefit is that this package could be a more robust solution to the configuration, Allowing us to do a better job of watching for updates to the configmap and managing subscriptions to those changes. This value store should also be usable by other parts of the application and therefore will require less usage of hard coded configuration structures and will allow the addition or removal of configuration values to not create such an issue. 
+
+### Work Items
+* Create a specific package for managing configuration
```
jmrodri: I'm assuming the configuration file stays the same, since you didn't specify any changes in the document.

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
...
+        map[
+            pass:testingboom 
+            org:ansibleplaybookbundle 
+            type:dockerhub 
+            name:play 
+            url:https://registry.hub.docker.com 
+            user:shurley
+        ]
+    ] 
+    dao:map[
```
jmrodri: This structure seems way more complicated since it's like the specific structs except as maps. Maybe I'm being myopic or simple minded but I always envisioned just a simple flat map.

```yaml
---
registry.docker.type: dockerhub
registry.docker.name: docker
registry.docker.url: https://registry.hub.docker.com
registry.docker.user: DOCKERHUB_USER
registry.docker.pass: DOCKERHUB_PASS
registry.docker.org: DOCKERHUB_ORG
registry.docker.fail_on_error: false
registry.rhcc.type: rhcc
registry.rhcc.name: rhcc
registry.rhcc.url: registry.access.stage.redhat.com
registry.rhcc.fail_on_error: true
registry.rhcc.white_list.1: "^legitimate.*-apb$"
registry.rhcc.black_list.2: "malicious.*-apb$"
registry.rhcc.black_list.3: "^specific-blacklist-apb$"
dao.etcd_host: localhost
dao.etcd_port: 2379
log.logfile: /var/log/ansible-service-broker/asb.log
log.stdout: true
log.level: debug
log.color: true
openshift.host: ""
openshift.ca_file: ""
openshift.bearer_token_file: ""
openshift.apb_sandbox_role: "admin"
openshift.image_pull_policy: IfNotPresent
broker.bootstrap_on_startup: true
broker.dev_broker: true
broker.launch_apb_on_bind: false
broker.recovery: true
broker.output_request: true
broker.ssl_cert_key: /path/to/key
broker.ssl_cert: /path/to/cert
broker.refresh_interval: "600s"
broker.auth.basic.type: basic
broker.auth.basic.enabled: true
```

Then the `Config` struct would look like this:

```golang
type Config interface {
    ToMap() map[string]string
    ContainsKey(key string) bool
    GetBoolean(string, bool) bool
    GetString(key string) string
    GetList(key string) []string
    SubsetMap(prefix string) map[string]string
     ... // many more utility methods
}

type MapConfiguration struct {
    configMap map[string]string
}

func NewConfigurationFromMap(configmap map[string]string) Config {
    return MapConfiguration{configMap: configmap}
}

func NewConfiguration() Config {
    return MapConfiguration{}
}

func (c MapConfiguration) SubsetMap(prefix string) map[string]string {
   // return a new map with all the entries that match the prefix from the embedded configMap
}
```

If I call `config.SubsetMap("registry.docker")` I will get a map with only the following keys:

```
registry.docker.type: dockerhub
registry.docker.name: docker
registry.docker.url: https://registry.hub.docker.com
registry.docker.user: DOCKERHUB_USER
registry.docker.pass: DOCKERHUB_PASS
registry.docker.org: DOCKERHUB_ORG
registry.docker.fail_on_error: false
```

If you called `config.GetBoolean("registry.docker.fail_on_error")` it will return false. Store everything as simple strings.

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
...
+* Take structured yaml and create a map of map of map.... saving the key as the name of the config value
+```go
+m["broker"] = map["dev_broker": true...]
+m["registry"] = [map["name": "dh"...], map["name": "play"...]]
+```
+* Create a syntax, and document, for retrieving a single configuration value from the list. I.E `config.MustGetBoolConfig("broker.dev_broker")`
+* Create a public API that will give the caller multiple ways to retrieve data.
+    - `MustGet` -> returns a single value and will panic if not able to retrieve the data
+    - `MustGet/Get<Type>` -> Allows for typed returns so the caller does not have to worry about type casting.
+    - Ability to specify a `map[string]interface{}` as a return value so one could say `config.MustGetMapConfig("broker")` and would retrieve the map of values for the broker configuration section.
```
jmrodri: I'd like to see this a little more. I don't really care for the `Must`. +1 to typed getters:

```
GetInt(key) int
GetString(key) string
```

Instead of the `Must` something that returns the default value if not present:
`GetSafeInt(key string, defValue int) int`

docs/proposals/configuration_changes.md
shawn-hurley: Yes, that was my hope. I was hoping that we could keep yaml because I think it looks nice and fits with the config map stuff that is already in place.

docs/proposals/configuration_changes.md
shawn-hurley: :+1: to the interface 
I think that this would totally work, it is just that we would
1. have to transform the yaml map to this structure
2. For calls like this, we would need to loop through every key to build the new map and return that, while the map approach would be constant time to retrieve each part.
3. returning map[string]string will make the caller use the [strconv](https://golang.org/pkg/strconv/) package, where if it is of type interface, they can just cast the value. I am not opposed to it, just making sure we are all in agreement. 

I am not opposed to doing the extra work in the config package, I just don't know if it will be easier to work with.

jmrodri: So after thinking about this overnight and looking at the yaml I proposed, I'm not sure I like it. The original one we had seems better. So -1 to my proposed flat file approach :)

As for map[string]string vs map[string]interface{}. I always struggle with configuration since they're usually just values in a file so treating them all as strings seems logical to me. I guess I fear if we use interface{} you could put in some custom struct in there is that really what we want?

Regarding the caller having to use strconv, I don't understand why? If we offer them conversion methods like `GetInt(key string) int`. For example, if I want the port number for etcd as an int I'd call `config.GetInt("dao.etcd_port")` and get back `2379`. The implementation of GetInt would have to do the conversions.  But if we used `interface{}` no one has to convert. hrm.

docs/proposals/configuration_changes.md
jmrodri: Like I said below, I think you're right, the current yaml would look nicer than my proposed yaml + properties file mixture.

jmrodri:
Okay here's what I'd like to do:

- create a Config interface with a bunch of utility methods
- create a Config implementation of that interface `MapConfiguration` or whatever is appropriate.
- keep the same format we have, as you suggested
- I'd get rid of the `Must` stuff just use `Get`s and `Put`s if we want to be able to programmatically update the config at runtime (in memory of course)
-stick with your map of map (of map of map sorry hahaha) implementation. Since it is hidden behind the interface and struct we really won't see it.

Some of the requirements I'd like are:

- given a top level key get a submap from the config. i.e. given "broker" give me the broker. submap
- be able to get a key as an int, string or string slice.

I think that's pretty much what you were proposing, just a few requirements.

jmrodri:
@eriknelson any thoughts on this?

shawn-hurley:
@jmrodri Just for my sanity: 
> be able to get a key as an int, string or string slice.

you mean a config value from a key correct?

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
+## Configuration Changes
+
+### Goal
+Create a package for creating and managing configuration values as a value store. This will be used to remove the necessity of each package to be loaded just for configuration values. The other benefit is that this package could be a more robust solution to the configuration, Allowing us to do a better job of watching for updates to the configmap and managing subscriptions to those changes. This value store should also be usable by other parts of the application and therefore will require less usage of hard coded configuration structures and will allow the addition or removal of configuration values to not create such an issue.
```
eriknelson: Is there a use case for config changes changing while a broker is running? Right now we just update the broker-config and bounce the pod.

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
+## Configuration Changes
+
+### Goal
+Create a package for creating and managing configuration values as a value store. This will be used to remove the necessity of each package to be loaded just for configuration values. The other benefit is that this package could be a more robust solution to the configuration, Allowing us to do a better job of watching for updates to the configmap and managing subscriptions to those changes. This value store should also be usable by other parts of the application and therefore will require less usage of hard coded configuration structures and will allow the addition or removal of configuration values to not create such an issue. 
+
```
rthallisey: +1

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
...
+### Work Items
+* Create a specific package for managing configuration 
+* Create a package mutex map to be used to store configuration values.
+* Take structured yaml and create a map of map of map.... saving the key as the name of the config value
+```go
+m["broker"] = map["dev_broker": true...]
+m["registry"] = [map["name": "dh"...], map["name": "play"...]]
+```
+* Create a syntax, and document, for retrieving a single configuration value from the list. I.E `config.MustGetBoolConfig("broker.dev_broker")`
```
rthallisey: +1 to an API.

docs/proposals/configuration_changes.md
eriknelson: Is there an alternative flavor of getter, like a `MayGet`?

If not, I'd just name them `Get`. Instead of the caller providing the default, should that be behind the `Getter` so it's configured in one place (the config package)? Trying to avoid having gets spread out across packages and possibly have each of those calls potentially use different defaults. Someone could easily come in and change it, and not know about the other calls.

docs/proposals/configuration_changes.md
```diff
@@ -0,0 +1,71 @@
...
+* Create a public API that will give the caller multiple ways to retrieve data.
+    - `MustGet` -> returns a single value and will panic if not able to retrieve the data
```
rthallisey: I'm not a fan of ```MustGet```.  I think a single API call makes more sense with the caller processing the possible return values and failures.

docs/proposals/configuration_changes.md
rthallisey: I like that we can retrieve sections of the config info. Do you have an example of how we can use this? If you don't it's not a big deal because this can come in handy down the road.  My thought is this could be useful when we handle upgrading the broker.

docs/proposals/configuration_changes.md
shawn-hurley: I think that we should eventually do this. I think it is a nice to have feature down the road, but that we should eventually have it. I don't see why, when I update the config map I then have to bounce the pod as the ideal end state IMO. I am not saying that as part of the initial PR to move to the new configuration that needs to work, just mentioning that moving to this new way will enable this feature in the future.

Does that make any sense?

eriknelson: So you want to edit the config file and have that trigger something in the broker that reconfigures it? I guess I just don't see this being used often, unless I'm misunderstanding.

shawn-hurley:
@jmrodri @rthallisey 
I have added a comparison to a library that already exists, as @eriknelson has asked for. 

I think that we should have a discussion if the weight of another vendor dependence is worth the benefit of not having to write the config package. 

I am leaning towards saying it is worth the dependence because this is something that we have to do, but not something I want to spend time on. I think that the interface is almost exactly what we were looking for and is used by some big projects so we can make some assumption that it is not completely broken. 

Also as a side note,
If we still go with creating our own config package, I did update the interface, to only have `Get` and `Put`methods and removed the concept of a `MustGet` method.

jmrodri:
@shawn-hurley @eriknelson @rthallisey I don't mind writing our own Config struct. Every project I've been on seems to write one and they've all been different. On RHN/Satellite 5 our configs were a series of .conf files with a hierarchy, on Candlepin we had a properties file (name=value) which was just string=string. 

So unless we need all the features of `viper` I'd vote just writing our own `Config`.

To @rthallisey I believe shawn does have an example of being able to get an entire section. So I should be able to get a map of the entire "broker" config, etc.

eriknelson:
> I am leaning towards saying it is worth the dependence because this is something that we have to do, but not something I want to spend time on. I think that the interface is almost exactly what we were looking for and is used by some big projects so we can make some assumption that it is not completely broken.

+1, this was mostly my thinking. I also don't see the addition of a vendor lib like this as an appreciable drawback; is the complaint the size of extra files? Doesn't seem like a big deal IMO.

jmrodri:
@shawn-hurley @eriknelson it isn't really about the size of the vendor (though viper does have its own set of deps that will pull in). It's just that config isn't that hard and wouldn't be much code. And I've always had to modify it to suit our particular needs. But at this point, we've spent more time debating than it would've taken to write a first cut at the config. So do whatever is easier.

rthallisey:
I'd rather we make our own pkg. We control it, it's not much code, and it cleans things up.

shawn-hurley:
I have updated the proposal to include the decision around building our own. 

Once Travis is completed can someone merge?
@jmrodri @eriknelson @rthallisey one last check that we are on the same page

PR #411, opened 2017-08-29
Default for no filter mode is to not contain a single APB.

shawn-hurley:
The goal of the PR is to make the default behavior for the time when there is no white list or black list to be the most restrictive, and not include any APBs.

 - Default for `NoFilterMode` is to not include any APBs
- Template updates to make catasb behave as it does today. 
- Updates to the `prepare_local_devel_env.sh` so that it will behave as it does today.


rthallisey:
What would be the use case when having no blacklist and no whitelist in this scenario?  You would end up with no Apps in your broker. I interpret having not blacklist and no whitelist as 'I don't have any filters' not 'I don't want anything' because I don't think there's any use for a broker without any Apps to serve. Can you go into more detail about what this solves?

shawn-hurley:
@rthallisey This is something that the security folks asked about. The reasoning is that if a cluster admin turns on the service broker the APBs have a lot of power, and they want the admin to make a very conscious choice to include the APBs that they would want to include.

rthallisey:
@shawn-hurley ok. Can you add a note based on your last comment to ```docs/filtering_apbs.md```?

dymurray:
ACK, LGTM.

PR #412, opened 2017-08-29
Kube template

rthallisey:
Resurrect the Kubernetes template. The template will be consumed by catasb.

 - Jinja2 k8s template
 - Python script to render the template using variables in k8s-variables.yaml

https://github.com/fusor/catasb/pull/120

### DIFF
--- a/templates/k8s-ansible-service-broker.yaml.j2
+++ b/templates/k8s-ansible-service-broker.yaml.j2
@@ -1,0 +1,197 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ansible-service-broker
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: asb
+  namespace: ansible-service-broker
+
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: asb
+subjects:
+- kind: ServiceAccount
+  name: asb
+  namespace: ansible-service-broker
+roleRef:
+  kind: ClusterRole
+  name: admin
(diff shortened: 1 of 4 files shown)


rthallisey:
[(20) Agnostic Runtime phase 1](https://trello.com/c/P1PfmxER/437-20-agnostic-runtime-phase-1)

rthallisey:
This works for me now. Could use some help testing from folks with different environments.

templates/k8s-variables.yaml
```diff
@@ -0,0 +1,81 @@
...
+# Path of the etcd binary
+etcd_path: /usr/local/bin/etcd
+
+# Configuration filepath for Ansible Service Broker
+broker_config: /etc/ansible-service-broker/config.yaml
+
+# SSL enabled: insecure="false", termination=Reencrypt
+# Insecure enabled: insecure="true", termination=edge
+# Run in insecure mode
+insecure: "true"
```
shawn-hurley: I thought that you said above that SSL was a limitation of the k8s? should we either comment or remove the variables that we don't need for k8s?

rthallisey: This var gets ignored. I'll add a comment since I would like to add it in the future.

templates/k8s-ansible-service-broker.yaml.j2
```diff
@@ -0,0 +1,241 @@
+#
+# Limitations With the Kubernetes template
+# ----------------------------------------
+# - ssl
+#
```
djzager: I would add that the broker is not included in this template.

rthallisey: the broker is down on line 193

PR #413, opened 2017-08-30
Proposal: CI Framework

rthallisey:
New CI framework proposal.

### DIFF
--- a/docs/proposals/ci-framework.md
+++ b/docs/proposals/ci-framework.md
@@ -1,0 +1,90 @@
+# Continuous Integration Framework
+
+## Introduction
+The current CI infrastructure uses the services: MediaWiki and Postgresql to
+test provision, bind, unbind, and deprovision on every PR to the
+ansible-service-broker repo. This infrastructure needs to scale so that it can
+test provision, bind, unbind, and deprovision on any set of APB(s).
+
+## Problem Description
+The four actions that make up the APB API are constant across all APBs. In
+order to scale the CI infrastructure, there needs to be a tool that gathers
+user input and translates that into one or more APB actions instead of a bash
+script that only works for MediaWiki and Postgresql.
+
+## Framework Overview
+The new framework could be written in: bash, go, or python. Although bash worked
+great for the initial gate work, go or python will be nicer for building an API.
+
+The goal is *not* to build something complex, but to provide an out of the box
+framework that will allow an APB developer to easily standup CI in their APB git
+repos using Travis or Jenkins.
+
+### Input
+Each APB repo will have a config file for running a CI job.
+
+Example input file for the mediawkik apb.
+```yaml
+ - provision: ansibleplaybookbundle/mediawiki123-apb
+ - provision: ansibleplaybookbundle/postgresql-apb
+ # Verify provision was successful
+
+ - bind: ansibleplaybookbundle/postgresql-apb
+ # Verify bind was successful
+
+ - verify: verify-app.sh
+ # Run as a shell command
+
+ - unbind: ansibleplaybookbundle/postgresql-apb
+ # Automatic verification occurs after the unbind
+
+ - deprovision: ansibleplaybookbundle/mediawiki123-apb
+ - deprovision: ansibleplaybookbundle/postgresql-apb
+ # Automatic verification occurs after the deprovision
+```
+
+### Output
+The execution of CI depends on the input order.
+
+## Code
+The functions from ```scripts/broker-ci/local-ci.sh``` will be translated into
+API objects with the config file as inputs.
+
+```bash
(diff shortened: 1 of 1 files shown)


jmontleon:
I like the idea. Will there be a way to input values when necessary as well? For instance, for mediawiki you need to specify a password and there is no default. In this case it seems the there would need to be a way to address this in input.

rthallisey:
@jmontleon I forgot to address config. Thanks for pointing that out.

In order to keep the CI yaml as simple as possible, I was thinking we could have a default location to drop and APB config file.  Since each APB will have it's own git repo, we can curl the location of any config we need.  I think this is a good way to start since it's simple. As things progress, we could always add config override or custom config file locations.

There's also @shawn-hurley's work with the test apb playbook and that playbook will have a set of defaults. Maybe we can share those defaults?

shawn-hurley:
I have a general question here, what are we attempting to verify with this CI? 
Are trying to make sure that the APB's are good to be published or are we trying to verify that the ASB can deploy and manage a configurable setup of APBs? Alternatively, verifying a third option that I am not aware of.

rthallisey:
This is to replace the existing CI we have, which does a provision and bind test between mediawiki and postgresql. 

The goal is to cover: ```ASB can deploy and manage a configurable setup of APBs```, but it can also tell us: ```APB's are good to be published```.

shawn-hurley:
> ASB can deploy and manage a configurable setup of APBs

Makes complete sense for this. 
> APB's are good to be published

This should be covered by the `test` action that we have defined. I think that makes more sense. These are going to be doing the same basic things, running a configurable number of actions on an APB. The benefit of this PR is that we can have two APBs interact while running the configurable number of actions, which IMO should be a test of the broker and should tell us nothing of the APBs status. 

> Each APB repo will have a config file for running a CI job.

This to me still feels wrong. I think the APB should be able to be tested independently of any service catalog/service broker for it's CI. The requirement that we make an APB test with specific other APBs, IMO, takes away from their generic nature (until we get to have dependencies). Also, in your example you're requiring and APB that has set bindable to false, to test the binding of a database APB. Which would we mark as failed if binding failed?  I think this can be easily remedied by moving the configuration out of the APB repo.

>There's also @shawn-hurley's work with the test apb playbook and that playbook will have a set of defaults. Maybe we can share those defaults?

I agree, we will need to add to the test documentation, that there is a specific default location, and will need to add some sort of mechanism for passing those in on certain actions, but yes this makes total sense to me.

rthallisey:
>>  APB's are good to be published
>>
> This should be covered by the test action that we have defined. I think that makes more sense. These 
> are going to be doing the same basic things, running a configurable number of actions on an APB. .
>
I agree with what your saying.  The goal of this framework is to enable complex testing.  But, basic testing is unavoidable when doing complex testing.  In the gate, the mediawiki and postgresql CI completely covers all of the smoke tests for mediawiki and postgresql. It's unavoidable.  The difference between this in the test playbook is that the playbook is portable and this CI sits in the repo. They are different use cases.
> The benefit of this PR is that we can have two APBs interact while running the configurable number of 
> actions, which IMO should be a test of the broker and should tell us nothing of the APBs status
>
I slightly disagree. It does tell us the APBs status, but only in the master branch. It does not tell us the status of any published APB image you pull from dockerhub. That's where the test playbook will give you a status because it's shipped with the APB code while the framework is not.

rthallisey:
>   Each APB repo will have a config file for running a CI job.
>
>> This to me still feels wrong. I think the APB should be able to be tested independently of any
>> service catalog/service broker for it's CI. 
>>
I think it's great to have APBs be independently tested from the catalog and broker, but it will improve our coverage if we also supplement that by testing with the broker and catalog.  I would like test bind, multi APB apps, and whether one APB works with another.
>> The requirement that we make an APB test with specific
>> other APBs, IMO, takes away from their generic nature (until we get to have dependencies). 
>>
The framework is designed to make it really easy to test with other APBs, but it's not a requirement.  The point of making it easy is to encourage APB developers to do it so we improve our CI testing.
>> Also, in your example you're requiring and APB that has set bindable to false, to test the binding of
>> a database APB. Which would we mark as failed if binding failed? I think this can be easily
>> remedied by moving the configuration out of the APB repo.
>>
The are no requirement that the APB does not itself impose. So if the APB can't bind to anything and tries to it will fail as expected.

rthallisey:
Spoke with @shawn-hurley.  To summarize our conversation:
 1) We want to try and reuse test.yml with the CI framework if possible.
 2) There's overlap with the test.yml work, but the CI framework will test APB using the broker and catalog while the test.yml is a portable smoketest.
 3) We want apb devs to use both the smoke test and CI framework when building new APBs.

PR #414, opened 2017-08-30
Update build to also work with Fedora 27

jmontleon:
Changes the build command in the RPM spec. We have been getting an error due to using -tags, but it has not been fatal until Fedora 27.

 - On the related bug the suggestion if using any other flag besides -o is to use the go build command directly rather than %gobuild, so I am doing that.
 - https://bugzilla.redhat.com/show_bug.cgi?id=1409931

depends-on <PR>
N/A

fixes <issue_number>
N/A


shawn-hurley:
ACK

jmrodri:
ACK

PR #416, opened 2017-09-01
Move variable assignment for clarity in script

djzager:
Based on discussion in the #asbroker, it seemed possible that moving the
variable assignments to just before the `oc process` would make it
easier to see our recommended values.


eriknelson:
Is this travis failure legit?

PR #419, opened 2017-09-05
updating default values for configuration values needed.

### LINKED ISSUE #424: Local dev environment fails to run without /var/run/secrets/kubernetes.io/serviceaccount/namespace
$make run
Running /home/rhallisey/src/github.com/openshift/ansible-service-broker/broker --config /home/rhallisey/src/github.com/openshift/ansible-service-broker/etc/generated_local_development.yaml --insecure
////////////////////////////////////////////////////////////////////////////////////////
==           Starting Ansible Service Broker...           ==
///////////////////////////////////////////////////////////////////////////////////////
ERROR: Failed to read config file
open /var/run/secrets/kubernetes.io/serviceaccount/namespace: no such file or directory
make: *** [run] Error 1

**What happened**:
Local dev environment fails to run.

Fix this by adding the expected file /var/run/secrets/kubernetes.io/serviceaccount/namespace

shawn-hurley:
This will set default values during the local run if they are not set in my vars. 

fixes #424

### DIFF
--- a/scripts/prep_local_devel_env.sh
+++ b/scripts/prep_local_devel_env.sh
@@ -215,7 +215,7 @@ openshift:
   ca_file:${CA_FILE}
   image_pull_policy: ${IMAGE_PULL_POLICY}
   sandbox_role: ${SANDBOX_ROLE:-edit}
-  namespace: ${NAMSPACE:-ansible-service-broker}
+  namespace: ${NAMESPACE:-ansible-service-broker}
 broker:
   dev_broker: true
   launch_apb_on_bind: false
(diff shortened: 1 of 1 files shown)


scripts/prep_local_devel_env.sh
```diff
@@ -214,7 +214,8 @@ openshift:
   bearer_token_file:${BEARER_TOKEN_FILE}
   ca_file:${CA_FILE}
   image_pull_policy: ${IMAGE_PULL_POLICY}
-  sandbox_role: ${SANDBOX_ROLE}
+  sandbox_role: ${SANDBOX_ROLE:-edit}
+  namespace: ${NAMSPACE:-ansible-service-broker}
```
cfchase: NAMSPACE -> NAMESPACE ?

rthallisey:
fixes #424

PR #420, opened 2017-09-05
Force delete the mediawiki pod

rthallisey:
Sometimes the mediawiki pod complains about a secrent not being
    mounted from the service account APB. I'm not sure why yet,
    but this will at least be a work around for now.


rthallisey:
@shawn-hurley May just go with the work around for now. I'm not sure what's causing it.

PR #423, opened 2017-09-06
Proposal to host static assets for APBs

cfchase:


docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+COPY assets /assets
+CMD ["noop"]
+```
+
+This docker container will be built with the same image name, but the tag will be prefixed with `_assets_`.  APB tooling will perform `docker build -t hello-world-apb:_assets_latest .`  and keep track of the image ID.
+
+After building the assets image, the tooling will also add a label to the APB image with the assets image ID.  This will ensure that if the APB is updated,  the matching assets will be associated and we can detect changes in the assets even if the APB spec has not changed.
+
+```bash
+docker build . -t docker.io/cfchase/hello-world-apb:latest --label "com.redhat.apb.assetsImageId=<ImageId>"
```
cfchase: I can use this method to eliminate the `LABEL "com.redhat.apb.spec"` from the Dockerfile as well and just have it as part of the build process so it's not checked in as changes.  That way our apb.yml and Dockerfile would not be out of sync anymore.  Let me know if that would be desired or if there are any reasons we want to keep it in the Dockerfile.

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
+# Portable Service Assets
+
+## Introduction
+Service classes in the Service Catalog require a number of UI graphical assets (i.e. service icon, company logo, etc).  For APBs that are sourced from the RHCC, we do not want to reference these externally on a remote (non-RH managed endpoint).  Additionally, we want to be able to handle off-network / disconnected clusters thus making this a truly portable solution.
```
jmrodri: first use of APBs in the document, I'd spell it out Ansible Playbook Bundles (APBs). then use APB in the rest of the document.

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+Assets will be stored under the APB folder in a folder named `assets`. As part of the `apb build` process a container image will be created copying all files from the `assets` folder into the container.
+
+```Dockerfile
+# Assets Dockerfile
+FROM scratch
+COPY assets /assets
+CMD ["noop"]
+```
+
+This docker container will be built with the same image name, but the tag will be prefixed with `_assets_`.  APB tooling will perform `docker build -t hello-world-apb:_assets_latest .`  and keep track of the image ID.
```
jmrodri: I like this approach.

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+
+After building the assets image, the tooling will also add a label to the APB image with the assets image ID.  This will ensure that if the APB is updated,  the matching assets will be associated and we can detect changes in the assets even if the APB spec has not changed.
```
jmrodri: interesting approach as well.

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+
+```bash
+docker build . -t docker.io/cfchase/hello-world-apb:latest --label "com.redhat.apb.assetsImageId=<ImageId>"
+```
+
+## Specifying Assets
+In a given apb.yml the user can specify `{{ assets }}` inside any metadata string values. Currently, only metadata is planned for support, but we could parse more if the use case is made.
```
jmrodri: add tick marks around apb.yml i.e `apb.yml`

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+#apb.yml
+name: hello-world-apb
+...
+metadata:
+  imageUrl: "{{ assets }}/my-pic.svg"
+  documentationUrl: "{{ assets }}/docs.html"
+```
+
+## Downloading Assets
+After every bootstrap of apb specs, ASB will begin downloading assets in an asynchronous work job.  For each spec, ASB will check to see if there is an associated assets image ID.  If there is none, no assets are downloaded.  If there is an assets image ID and the assets are missing or the ID has changed from the last download, it will download them from the built image assets (e.g. `hello-world-apb:_assets_latest`) using commands mimicking the following:
```
jmrodri: first use of ASB, change it to Ansible Service Broker (ASB), then use ASB in the rest of the document.

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+mkdir -p <www-docroot>/assets/apbs/dh-ansibleplaybookbundle-hello-world-apb-latest/
+docker create docker.io/ansibleplaybookbundle/hello-world-apb:_assets_latest
+docker cp $DOWNLOADED_IMAGE_ID:/assets/. <www-docroot>/assets/apbs/dh-ansibleplaybookbundle-hello-world-apb-latest
+```
+To avoid using the docker daemon, will attempt to use [buildah](https://github.com/projectatomic/buildah) or [skopeo](https://github.com/projectatomic/skopeo) to perform those commands.
```
jmrodri: change 'will attempt' to 'we will attempt'

docs/proposals/portable-service-assets.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Work Items
+* APB tooling to build assets
+* Serving files
+* Downloading assets
+* Modify APB processing for requests
+* Update documentation for building APBs
+
+## Future Considerations
+For running multiple instances of the broker, we'll have to eventually use a persistent volume and consider multiple pods downloading assets.
```
jmrodri: It would be interesting if creating a pod with a httpd server on it dedicated to serve content would be useful or not.

PR #428, opened 2017-09-08
User Impersonation Implementation 

shawn-hurley:
**Describe what this PR does and why we need it**:
This PR will allow us to check for users ability to run the APB, will run APB in transient name space that will hide the escalated permissions. 
Changes proposed in this pull request
 - Add open shift client and create/delete project, subject rules review ability 
 - Add copied security/validation code from origin to check if rules cover the cluster role's rules . 
 - Add ability to pull out originating user header info.

**Does this PR depend on another PR (Use this to track when PRs should be merged)**
depends-on <PR>
https://github.com/kubernetes-incubator/service-catalog/pull/1162

### Note:
This PR will depend on using openshift 3.7

PR #430, opened 2017-09-12
first pass at administration documentation

shawn-hurley:
Adding documentation for how to deploy the ansible service broker. 

depends-on <PR>
https://github.com/openshift/ansible-service-broker/pull/428

### DIFF
--- a/docs/administration.md
+++ b/docs/administration.md
@@ -1,0 +1,70 @@
+# Administration Examples
+
+The Ansible service broker has three main use cases from an administration perspective. Below we will describe the main use cases and discuss the configurations for the broker that would make sense of each one. 
+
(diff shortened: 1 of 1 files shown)


docs/administration.md
```diff
@@ -0,0 +1,70 @@
...
+...
+openshift:
+  ...
+  sandbox_role: "edit"
+broker:
+  ...
+  auto_escalate: true # will allow all users to deploy the approved APB. Could be false if the administrator would still like the broker to check the permissions.
+```
+
+### Very Limited Tenant
```
rthallisey: Is the difference between very limited and light that the admin will approve apbs instead of letting any apb through?

shawn-hurley: No, as far as configuration goes they are basically the same thing. But the deployment and what someone would want are very different and will make it more clear by separating them.

PR #431, opened 2017-09-13
Remove image field from APB spec

djzager:
A proposal to remove the image field from the APB spec, including some
implementation details.

### DIFF
--- a/docs/proposals/apb-remove-image-field.md
+++ b/docs/proposals/apb-remove-image-field.md
@@ -65,11 +65,75 @@ example in [pkg/registries/adapters/dockerhub_adapter.go](../../pkg/registries/a
 
 ```diff
 -  return imageToSpec(r.Log, req, r.Config.Tag)
-+  return imageToSpec(r.Log, req, fmt.Sprintf("%s:%s", imageName, r.Config.Tag))
++  return imageToSpec(r.Log, req, fmt.Sprintf("%s/%s:%s", r.RegistryName(), imageName, r.Config.Tag))
+```
+
+### Update FQName
+
+In order for these changes to work, we will also have to update the `FQName` of
+the APB Spec. Since we know that image name must be unique to a registry
+(each entry under `registry` in the broker config is a specific registry)
+**and** each registries name must be unique, then
+`${registry_name}-${apb_name}` is a sufficiently unique name for `FQName`. We
+will update the [pkg/broker/broker.go](../../pkg/broker/broker.go):
+
+```diff
+-  imageName := strings.Replace(spec.Image, ":", "-", -1)
+-  spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, imageName),
++  spec.FQName = strings.Replace(
++      fmt.Sprintf("%v-%v", registryName, spec.FQName),
+       "/", "-", -1)
+```
+
+## Examples
+
+With all of these changes and a broker configured with the following registries:
+
+```yaml
+---
+registry:
+  - type: dockerhub
+    name: dh
+    url: https://registry.hub.docker.com
+    user: changeme
+    pass: changeme
+    org: ansibleplaybookbundle
+  - type: dockerhub
+    name: dy
+    url: https://registry.hub.docker.com
+    user: changeme
+    pass: changeme
+    org: dymurray
+```
+
+These registries were chosen because there are name collisions among the 2
+registries. One such collision is on `mediawiki123-apb`.
+
+From the `dh` registry:
+
+```
+{
+    "id":"268dbc13f56297fdd3737b7d30104eb4",
+    "name":"dh-mediawiki123-apb",
+    "image":"docker.io/ansibleplaybookbundle/mediawiki123-apb:latest",
+    ...
+}
+```
+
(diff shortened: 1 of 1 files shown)


djzager:
[(2) Draft proposal to remove the image field from the APB Spec](https://trello.com/c/Gi6fVc18/457-2-draft-proposal-to-remove-the-image-field-from-the-apb-spec)

fabianvf:
Will this proposal interact at all with the way whitelisting/blacklisting apbs works? @eriknelson

docs/proposals/apb-remove-image-field.md
```diff
@@ -0,0 +1,75 @@
+# Remove Image Field from APB Spec Proposal
+
+## Introduction
+
+The broker currently relies on the image field of the
+[apb.yaml](https://github.com/fusor/ansible-playbook-bundle/blob/master/src/apb/dat/apb.yml.j2)
+to determine the location of the docker image to deploy instead of relying on the
+registry adapters to formulate the image path based on the location of the spec
```
rthallisey: How will this handle having multiple adaptors with duplicate images?

djzager: I'm not certain that I follow you. I feel like what you may be suggesting is the case where the broker is configured like so (using @dymurray since he actually has apb's in his org):

```
registry:
  - type: dockerhub
    name: dh
    url: https://registry.hub.docker.com
    user: changeme
    pass: changeme
    org: ansibleplaybookbundle
  - type: dockerhub
    name: dy
    url: https://registry.hub.docker.com
    user: changeme
    pass: changeme
    org: dymurray
```

And the images are stored in etcd, looking at MediaWiki since there is one in each.

For ansibleplaybookbundle organization:
```
{ ... "image":"ansibleplaybookbundle/mediawiki123-apb:latest" ...}
```

For dymurray organization:
```
{... "image":"dymurray/mediawiki123-apb:latest" ...}
```

Is this what you are referring to?

eriknelson: `FQName` was introduced to uniquely identify APBs across adapters; it can be considered our "primary key", unless I'm misunderstanding this.

eriknelson:
@fabianvf The filter works on image names reported by the adapters, so it's ultimately up to the adapters what they report: https://github.com/openshift/ansible-service-broker/blob/master/pkg/registries/registry.go#L74

djzager:
@eriknelson 

https://paste.fedoraproject.org/paste/8O4bB4aKzhdeszxhK25xeg shows 2 entries (in etcd) for similar apb's grabbed from different orgs.

and provisioning them gives (relevant portion of respective apb pods):
```
  labels:
    apb-fqname: dh-ansibleplaybookbundle-mediawiki123-apb-latest
```

and 

```
  labels:
    apb-fqname: dy-dymurray-mediawiki123-apb-latest
```

eriknelson:
LGTM, @djzager can we get that added to the proposal in some kind of `Examples` section for future reference?

docs/proposals/apb-remove-image-field.md
```diff
@@ -0,0 +1,75 @@
...
+-  spec.Image = fmt.Sprintf("%s:%s", spec.Image, apbtag)
++  spec.Image = image
+```
+
+Then update the adapters to include the image when calling `imageToSpec`, for
+example in [pkg/registries/adapters/dockerhub_adapter.go](../../pkg/registries/adapters/dockerhub_adapter.go):
+
+```diff
+-  return imageToSpec(r.Log, req, r.Config.Tag)
++  return imageToSpec(r.Log, req, fmt.Sprintf("%s:%s", imageName, r.Config.Tag))
```
shawn-hurley: I think here we could also add the adapter prefix for the image. for example docker hub should be `docker.io/<imageName>:APBTag`. We have been adding this value to the adapters as we go.

https://github.com/openshift/ansible-service-broker/blob/dec34bfcc4b7b0ba1579fecd438243673e67deba/pkg/registries/adapters/dockerhub_adapter.go#L60


https://github.com/openshift/ansible-service-broker/blob/dec34bfcc4b7b0ba1579fecd438243673e67deba/pkg/registries/adapters/adapter.go#L39

djzager: After talking with @shawn-hurley and seeing @eriknelson's request for examples, here is what I'm going to work on for the next commit (and what I'm not).

1. Will add the `RegistryName` to the `image` field that is stored in etcd (ie. `docker.io/ansibleplaybookbundle/mediawiki123-apb:latest`).
1. Look at what updates need to be made to `FQName` in `pkg/broker/broker.go:addNameAndIdForSpec` in order for 1) this to not break because of `docker.io` 2) see if we can shorten it.
1. ~~The `ORG` discussion as it relates to the adapter parsing the spec and how it is stored~~, I think this will at least in part be solved by having a closer look at the `FQName` ... so leaving this alone for now.

eriknelson:
ACK!

dymurray:
@djzager LGTM. I agree with @shawn-hurley that we should be removing the organization from the APB spec completely. The impetus for removing the 'image' field was that we don't want APBs tied to organizations anymore.

PR #432, opened 2017-09-13
Split the deploy.sh script to work with both kube & openshift

rthallisey:
These two scripts are very different because one will use OpenShift
templates and the other will use jinja2. I added an intermediate script
that will direct to the proper location.

 - Run ```make deploy``` for both openshift and kubernetes envs

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,67 +1,11 @@
 #!/bin/bash
-source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
-# from makefile
-BROKER_IMAGE=$1
-REGISTRY=$2
-DOCKERHUB_ORG=$3
+BROKER_ROOT=$(dirname "${BASH_SOURCE}")/..
 
-# can be overridden via my_local_dev_vars
-PROJECT=${ASB_PROJECT}
-ROUTING_SUFFIX="172.17.0.1.nip.io"
-OPENSHIFT_TARGET="https://kubernetes.default"
-REGISTRY_TYPE="dockerhub"
-DEV_BROKER="true"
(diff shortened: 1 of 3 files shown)


rthallisey:

rthallisey:
Testing
-----
The gate uses ```make deploy``` for with an OpenShift cluster. Since the gate is green, the other test here is with Kubernetes:
```
make deploy
```

scripts/deploy.sh
```diff
@@ -1,60 +1,11 @@
...
-oc delete clusterrolebindings --ignore-not-found=true asb
-
-# deploy
-oc new-project ${PROJECT}
-oc process -f ${BROKER_TEMPLATE} \
-  -n ${PROJECT} \
-  ${VARS} | oc create -f -
+if [[ "${KUBERNETES}" ]]; then
+    echo "Using Kubernetes"
+    "${BROKER_ROOT}/scripts/kubernetes/deploy.sh"
```
djzager: My recommendation is to bring in the `scripts/kubernetes/deploy.sh` and `scripts/openshift/deploy.sh` into this file.

rthallisey: What I didn't want to do is turn these scripts into if k8s then do 100 things else openshift do these 100 things. With two separate files there's a clean separation because these are two separate scripts. So anything kubernetes related I know exists in ```scripts/kubernetes/...```.  It's not the best situation, but Bash isn't exactly the best for creating abstraction.

jmrodri: @rthallisey +1 keep them separate. They will be easier to manage. Unless they are 100% the same, otherwise it will turn into a bunch of `if kube else` etc.

jmrodri:
@djzager -1 to merging the files into one.

PR #433, opened 2017-09-14
1497819 - Broker should not rely on image field of APB yaml

### LINKED ISSUE #288: Remove organization from the image field in the apb.yml
Current behavior:
#apb.yml
image: ansibleplaybookbundle/my-apb
This means that the broker downloads the spec from an organization, but may actually download the image from an entirely different organization, leading to confusion.

Proposed:
#apb.yml
image: my-apb
ASB will assume the apb image is in the same repository and org where the spec was found.

Will require changes in:
* ansible-service-broker - no longer use the org in a spec and use where it's hosted instead.
* ansible-playbook-bundle - remove org from apb-init and build

Reasoning:
1. Stable - Allow a user to download an image, retag it, and then push it to a different org for either testing or production use without modifying and rebuilding.  We want to test the built image as is without modifying it and introducing changes.
2. Less confusion - since a user may forget to change the org, they'd be testing the original image on a different org and may be confused why changes aren't reflected.
3. Reusable - we would want users and developers to be sharing images without forcing them to rebuild the image with changes before pushing to their own organization for use.

djzager:
The purpose of this change is to make it so that the full path to each
image is used when executing APBs instead of relying on what is stated
in the APB yaml.

- Update the Spec to make Image optional (for when we remove it from APB
  yaml)
- Update the registry adapters to store the full path to the image when
  unmarshalling the APB yaml -> APB spec before we store it into `etcd`
- Update the broker to have simpler format for `FQName`

Implements #431
Fixes #288
Bug 1497819

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -62,7 +62,7 @@ type Plan struct {
 type Spec struct {
 	ID          string                 `json:"id"`
 	FQName      string                 `json:"name" yaml:"name"`
-	Image       string                 `json:"image"`
+	Image       string                 `json:"image,omitempty"`
 	Tags        []string               `json:"tags"`
 	Bindable    bool                   `json:"bindable"`
 	Description string                 `json:"description"`
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -62,7 +62,7 @@ type Plan struct {
 type Spec struct {
 	ID          string                 `json:"id"`
 	FQName      string                 `json:"name" yaml:"name"`
-	Image       string                 `json:"image,omitempty"`
+	Image       string                 `json:"image" yaml:"-"`
 	Tags        []string               `json:"tags"`
 	Bindable    bool                   `json:"bindable"`
 	Description string                 `json:"description"`
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -268,8 +268,8 @@ func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 		// need to make / a hyphen to allow for global uniqueness
 		// but still match spec.
 
-		imageName := strings.Replace(spec.Image, ":", "-", -1)
-		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, imageName),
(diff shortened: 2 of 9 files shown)


pkg/broker/broker.go
```diff
@@ -264,8 +264,8 @@ func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 		// need to make / a hyphen to allow for global uniqueness
 		// but still match spec.
 
-		imageName := strings.Replace(spec.Image, ":", "-", -1)
-		spec.FQName = strings.Replace(fmt.Sprintf("%v-%v", registryName, imageName),
+		spec.FQName = strings.Replace(
+			fmt.Sprintf("%v-%v", registryName, spec.FQName),
```
shawn-hurley: I worry that the FQName could have `.` in it. If it does, this has caused issues with the service catalog in the past and might be worth replacing here.

pkg/apb/types.go
```diff
@@ -62,7 +62,7 @@ type Plan struct {
 type Spec struct {
 	ID          string                 `json:"id"`
 	FQName      string                 `json:"name" yaml:"name"`
-	Image       string                 `json:"image"`
+	Image       string                 `json:"image,omitempty"`
```
shawn-hurley: I don't understand the need for this change. I don't think it will hurt but don't know why we need it. 

I could see you doing `yaml:"-"` so that we never take anything for the APB spec.

djzager:
[Remove image from APB Spec.](https://trello.com/c/siyUDbM0/456-remove-image-from-apb-spec)

pkg/apb/types.go
djzager: you are correct. I don't believe I fully understand yet what the rules and side-effects are for struct tags. I couldn't find anything that detailed `yaml:"-"` or the json/yaml struct tags in general, if there is anything you could point me to that would be helpful.

rthallisey:
Since this is a feature, do we want to wait on merging this until we branch?

jmrodri:
@djzager commit should be "bug # - commit message". Renamed the PR.

djzager:
I believe this is ready now.

PR #434, opened 2017-09-14
Make the prep_local_devel_env script work for Kubernetes & Openshift

rthallisey:
Most of the prep_local_devel_env.sh script was shared between the
two clusters. The main differences were:
  - etcd_port: 2379
  - kubernetes_port: 6443
  - routes
  - deployments
  - ca.crt vs service-ca.crt
  - oc to kubectl command


rthallisey:
Worked with Kubernetes. Testing with OpenShift.

rthallisey:

rthallisey:
Working for both openshift and k8s

rthallisey:
Testing
-----------
Make a new my_local_dev_vars
```bash
cp scripts/my_local_dev_vars.example  scripts/my_local_dev_vars
```
Run prep_local_devel_env script.
```bash
./scripts/prep_local_devel_env.sh
```
Run a local development env.
```
make run
```

rthallisey:
This is working with insecure. I'll need to add ssl shorly since inscure mode is being removed.

scripts/prep_local_devel_env.sh
```diff
@@ -130,37 +136,38 @@ fi
 echo "TLS Cert: tls.crt"
 echo -e "Wrote \n${TLS_KEY_DATA}\n to: ${TLS_KEY}\n"
 # Kill any running broker pods
-oc scale deploymentconfig asb --replicas 0 -n ${ASB_PROJECT}
+deployments scale asb --replicas 0 -n ${ASB_PROJECT}
```
djzager: Should it not be `kubectl deployments ...` ?

scripts/kubernetes/resources.sh
```diff
@@ -0,0 +1,60 @@
...
+    port: 1338
+    protocol: TCP
+EOF
+}
+
+function get-ca {
+    kubectl get secret ${BROKER_SVC_ACCT_SECRET_NAME} -n ${ASB_PROJECT} -o jsonpath='{ .data.ca\.crt }'
+}
+
+function ectd-port {
```
djzager: Should this be `etcd-port`?

scripts/openshift/resources.sh
```diff
@@ -0,0 +1,30 @@
...
+    asb_project="$2"
+    args="${@:3}"
+    oc process -f $template -n $asb_project $args | oc create -n $asb_project -f -
+}
+
+function get-ca {
+    kubectl get secret ${BROKER_SVC_ACCT_SECRET_NAME} -n ${ASB_PROJECT} -o jsonpath='{ .data.service-ca\.crt }'
+}
+
+function ectd-port {
```
djzager: Here too.

scripts/cluster-redirect.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
djzager: I think that this file + `scripts/kubernetes/resources.sh` and `scripts/openshift/resources.sh` should be combined into a single file at `scripts/lib/cluster.sh`. Looking something like:

```
if [ -n "${KUBERNETES}" ]; then
    echo "Kubernetes Cluster"
    function asb::cluster::deployments {
        ...
    }
    ...
else
    echo "OpenShift Cluster"
    function asb::cluster::deployments {
        ...
    }
    ...
fi
```

scripts/prep_local_devel_env.sh
djzager: Now that I've tracked it down, I see that this is a function call that just so happens to look like the only thing that is missing is the `kubectl` in front. I think if you were to put those functions in `scripts/lib/cluster.sh`, conditionally defined for OpenShift vs Kubernetes, then `asb::cluster::deployments` would clearly read as a library function call.

scripts/cluster-redirect.sh
rthallisey: I never really liked this cluster redirect script. What I'm going to do is add it to the top of the prep_local_dev_env.sh.

Similar to deploy.sh, I like having the abstraction of a directory for $CLUSTER files. I think it makes it easy to separate $CLUSTER scripts without overloading a single file.

jmrodri: @djzager -1 to the combining the files. while it does duplicate somethings it is way easier to see. Does this change kubernetes? open scripts/kubernetes files. Does it affect openshift? open scripts/openshift files. if they are combined you have to be more cognizant not to break the other worrying about the conditionals.

djzager: @jmrodri fair point.

My general feeling is that we are turning a single file into 4 to prepare the local environment, when there are only subtle differences between how to prepare the environment in k8s vs openshift and that strikes me as overkill.

Having said that, @rthallisey, delete `scripts/openshift/resources.shs` and I'll change my vote.

rthallisey: @djzager we do end up with 4, but prep_local_devel_env.sh is already hard to read and I think this does minimal damage to it's readability.  I do think there is a bunch of overlap here and a lot of potential for improvements.  This patch is mainly for functionality without changing how the script reads.  If you want to improve it, there is definitely an opportunity to and I would encourage it! But, it wasn't my intention with this patch. 

I'll remove the extra file.

PR #436, opened 2017-09-15
pass in BROKER_KIND

jmrodri:
thomasmckay in irc was trying to run the latest service broker with a 3.7 server. The template uses `Broker` kind by default instead of the required `ServiceBroker`. Defaulting to `Broker` so that it works with 3.6 and folks can override it:

```
BROKER_KIND="ServiceBroker" ./scripts/run_latest_build.sh
```


shawn-hurley:
ACK

PR #438, opened 2017-09-15
Look at the apbs in the catalog for a matching name when creating a secret

fabianvf:
Running the `./scripts/create_broker_secret.py` script will now check the catalog for a FQName that looks like the image passed in. This should make it work with non-dockerhub registries as well as locally pushed images.

 - Added dependency on requests library to use the script
 - Get a list of apbs from `/v2/catalog`
 - Match the provided org, image, and tag to the apbs in the broker
  - If multiple match, prompt the user to pick one

depends on https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/143

### DIFF
--- a/scripts/create_broker_secret.py
+++ b/scripts/create_broker_secret.py
@@ -138,13 +138,13 @@ def format_config(config):
 
 
 def broker_auth(config):
-    credentials = {}
+    credentials = {'basic_auth_username': None, 'basic_auth_password': None}
     auth_settings = config['data']['broker-config']['broker'].get('auth')[0]
     if auth_settings.get('type') == 'basic' and auth_settings.get('enabled'):
         secret = yaml.load(runcmd('oc get secret asb-auth-secret -o yaml'))
         credentials = {
-            "basic_auth_username": base64.b64decode(secret['data']['username']),
-            "basic_auth_password": base64.b64decode(secret['data']['password'])
+            "basic_auth_{}".format(k): base64.b64decode(v)
+            for (k, v) in secret['data'].items()
         }
     return credentials
 
--- a/scripts/create_broker_secret.py
+++ b/scripts/create_broker_secret.py
@@ -1,7 +1,27 @@
 #! /usr/bin/env python
+
 import sys
 import yaml
 import subprocess
+
+try:
+    import yaml
+except Exception:
+    print("No yaml parsing modules installed, try: pip install pyyaml")
+    sys.exit(1)
+
+try:
+    import requests
+except Exception:
+    print("requests module not installed, try: pip install requests")
+    sys.exit(1)
+
+
+# Work around python2/3 input differences
+try:
+    input = raw_input
+except NameError:
+    pass
 
 USAGE = """USAGE:
   {command} NAME NAMESPACE IMAGE [KEY=VALUE]* [@FILE]*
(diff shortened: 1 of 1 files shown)


djwhatle:
This change allowed me to add broker secrets without needing to edit the config map for APBs that I had used "apb push" to get into the catalog.

shawn-hurley:
visual ACK

scripts/create_broker_secret.py
```diff
@@ -104,17 +124,49 @@ def format_config(config):
...
-    return '-'.join([registries[registry], org, image, tag])
+
+def get_all_apbs():
+    response = requests.get(broker_url() + '/v2/catalog', verify=False)
+    return response.json()['services']
+
+
+
+def fqname(apb):
+    parts = apb.split('/')
```
rthallisey: Would doing ```parts = apb.split('/')[-1]``` work here?
It looks like this breaks the image apart like this: ```rthallisey/mediawiki123:latest``` -> ```mediawiki123:latest```

fabianvf: no, given either `rthallisey/mediawiki123:latest` or `docker.io/rthallisey/mediawiki123:latest` it should produce `rthallisey-mediawiki123-latest`

scripts/create_broker_secret.py
```diff
@@ -104,17 +124,49 @@ def format_config(config):
...
+    if len(parts) == 3:
+        # Chop the service, we don't know what it will be translated to in the broker
+        parts = parts[1:]
+    if ':' in parts[-1]:
```
rthallisey: Would this if statement ever be skipped? If it can, shouldn't latest be added to the end of parts?

scripts/create_broker_secret.py
rthallisey: You could do this:
```python
if apb.count('/') >= 2:
    parts = apb.split("/", 1)[-1]
parts = parts.replace("/", "-").replace(":", "-")
```

scripts/create_broker_secret.py
```diff
@@ -1,8 +1,28 @@
 #! /usr/bin/env python
+
 import sys
 import yaml
 import subprocess
 
+try:
+    import yaml
```
djzager: You have `import yaml`, so I don't believe you would ever reach `print("No yaml parsing modules installed, try: pip install pyyaml")` as it stands now.

djzager: It isn't immediately obvious why we are try/catching this. It would be helpful, to me at least, if you could explain why you are doing this.

scripts/create_broker_secret.py
```diff
@@ -104,17 +124,49 @@ def format_config(config):
...
 
-    return '-'.join([registries[registry], org, image, tag])
+
+def get_all_apbs():
+    response = requests.get(broker_url() + '/v2/catalog', verify=False)
+    return response.json()['services']
+
+
+
+def fqname(apb):
```
djzager: It looks like I need to make updates to this in my PR https://github.com/openshift/ansible-service-broker/pull/433 since the `FQName` will be changing with those changes.

scripts/create_broker_secret.py
fabianvf: I'm just try-catching it because this isn't a python project and we can't really specifiy python dependencies, so I wanted a nice error message if they don't have the right dependencies installed. I messed up with the above import yaml though, oops.

scripts/create_broker_secret.py
fabianvf: yeah that's way cleaner

fabianvf:
@rthallisey @djzager I think I addressed your comments.

fabianvf:
This has been updated to deal with auth now

jwmatthews:
I removed the needs-bugzilla from this because this isn't part of the shipping broker, it's a tool used to aid testing.

PR #439, opened 2017-09-16
Additional workaround for the gate

rthallisey:
The gate still occasionally fails on a pod that isn't being removed.  Ignore the 'Terminating' pod by searching for 'Running'.

 - Look only for the running pod and ignore the terminating pod


rthallisey:
closing in favor of #453

PR #441, opened 2017-09-18
adding ability to pass in the CA Bundle for ServiceBroker

shawn-hurley:
Allows you to deploy the service broker with the CA Bundle defined

depends-on <PR>
https://github.com/fusor/catasb/pull/140


dymurray:
This PR worked for me with the associated catasb PR. I would suggest some logging messages in case someone is trying to use 3.7 and the cabundle can't be found.

PR #444, opened 2017-09-20
Update schema for instance-update

jmontleon:


jmontleon:
Update requests can be performed with asbcli as well, but for sake of demonstrating bad input"

```
## rhscl-postgresql-apb - no plan specified, continues with current plan (dev)
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/6119f273-5ef6-4899-89b0-bd7ad5c2e05a -d '{}'
{}

## rhscl-postgresql-apb - prod plan specified with current plan dev
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/6119f273-5ef6-4899-89b0-bd7ad5c2e05a -d '{"plan_id": "prod"}'
{}

## rhscl-postgresql-apb - trying to update to a non existent plan
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/6119f273-5ef6-4899-89b0-bd7ad5c2e05a -d '{"plan_id": "blablabla"}'
{
  "description": "plan not found"
}

## rhscl-postgresql-apb - trying to update from prod to dev (not possible)
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/6119f273-5ef6-4899-89b0-bd7ad5c2e05a -d '{"plan_id": "dev"}'
{
  "description": "plan update not possible"
}

## mediawiki123-apb - new parameters specified:
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/ef0efd9e-6982-44da-a600-1e05e2a1cca1 -d '{"parameters": {"mediawiki_site_name": "wackawacka"}}'
{}

## mediawiki123-apb - new plan_id and parameters specified:
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/ef0efd9e-6982-44da-a600-1e05e2a1cca1 -d '{"plan_id": "default", "parameters": {"mediawiki_site_name": "wackawacka"}}'
{}

## mediawiki123-apb - bad plan specified
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/ef0efd9e-6982-44da-a600-1e05e2a1cca1 -d '{"plan_id": "sdfgsdgfsdfgsdfg", "parameters": {"mediawiki_site_name": "wackawacka"}}'
{
  "description": "plan not found"
}

## mediawiki123-apb - trying to update a non-updatable parameter
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/ef0efd9e-6982-44da-a600-1e05e2a1cca1 -d '{"parameters": {"mediawiki_site_lang": "foo"}}'
{
  "description": "parameter not updatable"
}

## mediawiki123-apb - trying to update a non-existent parameter
$ curl -k -XPATCH -H "Authorization: Bearer $(oc whoami -t)" -H 'Content-Type: application/json'  https://asb-1338-ansible-service-broker.172.18.0.1.nip.io/ansible-service-broker/v2/service_instances/ef0efd9e-6982-44da-a600-1e05e2a1cca1 -d '{"parameters": {"bar": "foo"}}'
{
  "description": "parameter not found"
}
```

pkg/broker/broker.go
```diff
@@ -850,10 +859,194 @@ func (a AnsibleBroker) Unbind(
...
+	if err != nil {
+		// etcd return not found i.e. code 100
+		if client.IsKeyNotFound(err) {
+			return nil, ErrorNotFound
+		}
+		// otherwise unknown error bubble it up
+		return nil, err
+	}
+
+	// Make sure that we can find the plan we're updating from. This should probably never fail, but cover our tail.
```
eriknelson: I would abstract the ladder of validations that occurs here into their own, testable functions. Makes things read on a nice high level for the uninitiated and hides the ugly details, while also letting us easily change or add to their implementation without changing the core flow that will likely remain the same:

```
if err := validateCurrentPlan(); err != nil {
		a.log.Error("The plan %s, specified for updating from on instance %s, does not exist.", (*si.Parameters)["_apb_plan_id"], si.ID)
 		return nil, ErrorPlanNotFound
}

if err := validateNextPlan(); err != nil {
		a.log.Error("The plan %s, specified for updating from on instance %s, does not exist.", (*si.Parameters)["_apb_plan_id"], si.ID)
 		return nil, ErrorPlanNotFound
}

// Make sure that the current plan supports updating to the requested plan if a plan update was requested.
if err := validatePlanTransition(currentPlan, nextPlan); err != nil {
// Invalid Transition! throw error.
}

// Make sure that parameters requested for update exist on the new plan and that they are updatable.
if err := validateNextParameters(param, plan); err != nil { /* handle */ }

// elsewhere
func validateCurrentPlan(...) {
// Iterate spec plans and make sure the plan pulled from the si params actually exists
}

func validateNextPlan() {
// Iterate and confirm
		a.log.Error("The plan %s, specified for updating to on instance %s, does not exist.", req.PlanID, si.ID)
		return nil, ErrorPlanNotFound
}

// Etc...
```

pkg/broker/broker.go
```diff
@@ -850,10 +859,194 @@ func (a AnsibleBroker) Unbind(
...
+			}
+		}
+		if paramUpdated == false {
+			a.log.Error("Parameter %s, requested for update on instance %s, does not exist.", k, si.ID)
+			return nil, ErrorParameterNotFound
+		}
+	}
+
+	// We're ready to provision so save
+	if err = a.dao.SetServiceInstance(instanceUUID.String(), si); err != nil {
```
eriknelson: I think we might be doing this during normal provisions, but I'll raise the question: Do we want to be persisting service instances before we know they've *actually* been successfully created? Do they need to be extended to add some kind of `Status` field that's set to {Success|Error} at the conclusion of the APB run?

This is probably outside the scope of this PR, but worth consideration.

pkg/broker/broker.go
```diff
@@ -850,10 +859,194 @@ func (a AnsibleBroker) Unbind(
...
+		return nil, err
+	}
+
+	var token string
+
+	if async {
+		a.log.Info("ASYNC update in progress")
+		// asyncronously provision and return the token for the lastoperation
+		pjob := NewProvisionJob("update", si, a.clusterConfig, a.log)
```
eriknelson: We probably want a strongly typed `UpdateJob` suite, even if they do the same thing as provison. If the latter is true, we should refactor to account for all the shared stuff and better delegate specifics. This is kiind of a legacy thing, I'm happy  to help untangle this in another PR.

pkg/broker/broker.go
```diff
@@ -850,10 +859,194 @@ func (a AnsibleBroker) Unbind(
...
+		if err != nil {
+			a.log.Error("Failed to start new job for async update\n%s", err.Error())
+			return nil, err
+		}
+
+		// HACK: there might be a delay between the first time the state in etcd
+		// is set and the job was already started. But I need the token.
+		a.dao.SetState(instanceUUID.String(), apb.JobState{Token: token, State: apb.StateInProgress})
+	} else {
+		// TODO: do we want to do synchronous updating?
```
eriknelson: We'll likely only see async requests for update from the catalog, but we should probably support sync updates. If not, we should explicitly 422.

pkg/broker/provision_job.go
```diff
@@ -51,18 +52,19 @@ func (m ProvisionMsg) Render() string {
...
 		clusterConfig:   clusterConfig,
-		log:             log}
+		log:             log,
+		task:            task}
 }
 
 // Run - run the provision job.
 func (p *ProvisionJob) Run(token string, msgBuffer chan<- WorkMsg) {
-	podName, extCreds, err := apb.Provision(p.serviceInstance, p.clusterConfig, p.log)
+	podName, extCreds, err := apb.Provision(p.task, p.serviceInstance, p.clusterConfig, p.log)
```
eriknelson: I'm not a huge fan of parameterizing `Provision` with a couple variants of behavior; it breaks the semantics of the function from a caller's perspective. One thought would be to have two distinct `Provision` and `Update` available on the public interface, and behind the scenes they actually do some specific behavior for their case, and call into some shared function.

pkg/broker/util.go
```diff
@@ -219,7 +223,14 @@ func parametersToSchema(plan apb.Plan) Schema {
 					Required:   createRequired,
 				},
 			},
-			Update: map[string]*schema.Schema{},
+			Update: map[string]*schema.Schema{
+				"parameters": {
+					SchemaRef:  schema.SchemaURL,
+					Type:       []schema.PrimitiveType{schema.ObjectType},
+					Properties: updatableProperties,
+					Required:   updatableRequired,
```
eriknelson: Will this always be the set of required parameters that **also** are update-able?

pkg/broker/util.go
```diff
@@ -291,6 +302,33 @@ func extractRequired(params []apb.ParameterDescriptor) []string {
 	return req
 }
 
+func extractUpdatable(params []apb.ParameterDescriptor) map[string]*schema.Schema {
```
eriknelson: :+1: spec transforms look good.

pkg/handler/handler.go
```diff
@@ -286,21 +286,42 @@ func (h handler) update(w http.ResponseWriter, r *http.Request, params map[strin
...
+			writeResponse(w, http.StatusOK, resp)
+		case broker.ErrorNotFound:
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+		default:
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+		}
+	} else if async {
+		writeDefaultResponse(w, http.StatusAccepted, resp, err)
+	} else {
+		writeDefaultResponse(w, http.StatusCreated, resp, err)
```
eriknelson: This should be an http.StatusOK

pkg/broker/broker.go
```diff
@@ -850,10 +859,194 @@ func (a AnsibleBroker) Unbind(
...
+
+	si, err := a.dao.GetServiceInstance(instanceUUID.String())
+	if err != nil {
+		a.log.Debug("Error retrieving instance")
+		return nil, ErrorNotFound
+	}
+
+	// If no plan was specified we'll continue with the current plan
+	if req.PlanID == "" {
+		req.PlanID = (*si.Parameters)["_apb_plan_id"].(string)
```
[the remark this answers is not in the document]
eriknelson: One other note now that I look at it again: There's a const that should be used instead of the string literal `_apb_plan_id` for this key: https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/types.go#L35

So any `"_apb_plan_id"` should get updated to be `planParameterKey` instead.

pkg/apb/provision.go
```diff
@@ -21,85 +21,29 @@
...
-	}
-
-	executionContext, err := ExecuteApb(
-		"provision", clusterConfig, instance.Spec,
-		instance.Context, instance.Parameters, log,
+	// Nearly all of the logic for provisioning or updating is shared between
+	// provision and update, save for passing through the method type. Provision
+	// provides a nice public interface, but the bulk of the work is passed to
+	// provision_or_update as an implementation detail.
+	return provision_or_update(
```
shawn-hurley: methods should not have underscores in their names.

pkg/apb/provision.go
```diff
@@ -21,85 +21,29 @@
...
 	log.Notice("                       PROVISIONING                         ")
 	log.Notice("============================================================")
-	log.Notice(fmt.Sprintf("Spec.ID: %s", instance.Spec.ID))
-	log.Notice(fmt.Sprintf("Spec.Name: %s", instance.Spec.FQName))
-	log.Notice(fmt.Sprintf("Spec.Image: %s", instance.Spec.Image))
-	log.Notice(fmt.Sprintf("Spec.Description: %s", instance.Spec.Description))
+	log.Noticef("Spec.ID: %s", instance.Spec.ID)
+	log.Noticef("Spec.Name: %s", instance.Spec.FQName)
+	log.Noticef("Spec.Image: %s", instance.Spec.Image)
+	log.Noticef("Spec.Description: %s", instance.Spec.Description)
```
shawn-hurley: :+1:

pkg/apb/provision_or_update.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	logging "github.com/op/go-logging"
+	"github.com/openshift/ansible-service-broker/pkg/runtime"
+)
+
+type executionMethod string
+
+const (
+	executionMethodProvision executionMethod = "provision"
+	executionMethodUpdate    executionMethod = "provision"
```
shawn-hurley: Probably a type: `"provision"` -> `"update"`

pkg/handler/handler.go
```diff
@@ -286,21 +286,42 @@ func (h handler) update(w http.ResponseWriter, r *http.Request, params map[strin
 	defer r.Body.Close()
 	h.printRequest(r)
 
+	h.log.Debug(params["instance_uuid"])
```
shawn-hurley: Think we may want to remove or add more context to this debug statement?

pkg/apb/provision_or_update.go
eriknelson: woops, thanks @shawn-hurley

eriknelson:
Confirmed I'm able to transition from "dev" to "prod" plans as well as modify `postgresql_version` from 9.4 -> 9.5.

pkg/broker/broker.go
eriknelson: Do we think this is a blocking issue for merging this PR, or should this get logged as an issue and submitted in a separate PR?

shawn-hurley: @eriknelson I think a separate PR should be fine IMO

pkg/apb/provision_or_update.go
```diff
@@ -0,0 +1,96 @@
...
+		log.Error("No image field found on the apb instance.Spec (apb.yaml)")
+		log.Error("apb instance.Spec requires [name] and [image] fields to be separate")
+		log.Error("Are you trying to run a legacy ansibleapp without an image field?")
+		return "", nil, errors.New("No image field found on instance.Spec")
+	}
+
+	ns := instance.Context.Namespace
+	log.Info("Checking if project %s exists...", ns)
+	if !projectExists(ns) {
+		log.Error("Project %s does NOT exist! Cannot provision requested %s", ns, instance.Spec.FQName)
```
shawn-hurley: I think this should be `log.Errorf`

pkg/apb/provision_or_update.go
```diff
@@ -0,0 +1,96 @@
...
+		log.Errorf("apb::%s error occurred", string(method))
+		log.Error("%s", err.Error())
+		return executionContext.PodName, creds, err
+	}
+
+	return executionContext.PodName, creds, err
+}
+
+func projectExists(project string) bool {
+	_, _, code := runtime.RunCommandWithExitCode("kubectl", "get", "project", project)
```
shawn-hurley: I would really prefer to not rely on `kubectl` and use the namespace [client-go interface](https://godoc.org/k8s.io/client-go/kubernetes/typed/core/v1#NamespaceInterface). This is going to help us remove places that we are running commands rather than using the client. 

If you feel that we should let this pass through and deal with it during the other refactors I understand that just want to call it out here.

pkg/handler/handler.go
```diff
@@ -293,14 +293,33 @@ func (h handler) update(w http.ResponseWriter, r *http.Request, params map[strin
...
 		writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
 		return
 	}
 
-	resp, err := h.broker.Update(instanceUUID, req)
+	var async bool
 
-	writeDefaultResponse(w, http.StatusOK, resp, err)
+	// ignore the error, if async can't be parsed it will be false
+	async, _ = strconv.ParseBool(r.FormValue("accepts_incomplete"))
```
shawn-hurley: We must be checking for auto escalate and checking for users access to the namespace. see: https://github.com/openshift/ansible-service-broker/blob/master/pkg/handler/handler.go#L246

pkg/apb/provision_or_update.go
eriknelson: I agree, this code has been here for some time however and simply got moved over from Provision, so I'd rather address it in another PR.

pkg/handler/handler.go
eriknelson: I'll take a look at this, thanks for pointing it out @shawn-hurley

pkg/broker/broker.go
```diff
@@ -71,7 +81,7 @@ type Broker interface {
 	Bootstrap() (*BootstrapResponse, error)
 	Catalog() (*CatalogResponse, error)
 	Provision(uuid.UUID, *ProvisionRequest, bool) (*ProvisionResponse, error)
-	Update(uuid.UUID, *UpdateRequest) (*UpdateResponse, error)
+	Update(uuid.UUID, *UpdateRequest, bool) (*UpdateResponse, error)
```
jmrodri: good catch for async

pkg/broker/broker.go
```diff
@@ -411,11 +421,25 @@ func (a AnsibleBroker) Recover() (string, error) {
...
-			pjob := NewProvisionJob(instance, a.clusterConfig, a.log)
+			var job Work
+			var topic WorkTopic
+			if rs.State.Method == apb.JobMethodProvision {
+				job = NewProvisionJob(instance, a.clusterConfig, a.log)
+				topic = ProvisionTopic
+			} else if rs.State.Method == apb.JobMethodUpdate {
+				job = NewUpdateJob(instance, a.clusterConfig, a.log)
+				topic = UpdateTopic
+			} else {
```
jmrodri: Logic seems correct.

pkg/broker/broker.go
```diff
@@ -411,11 +421,25 @@ func (a AnsibleBroker) Recover() (string, error) {
...
+				a.log.Warningf(
+					"Attempted to recover job %s, but found an unrecognized "+
+						"MethodType: %s, skipping...",
+					rs.State.Token, rs.State.Method,
+				)
```
jmrodri: You really want me to make the line length 80 characters :)

pkg/broker/broker.go
```diff
@@ -993,10 +1017,256 @@ func (a AnsibleBroker) Unbind(
...
+		// will understand what to do with it, since APBs do not understand plan hashes.
+		toPlanName, err = a.dao.GetPlanName(req.PlanID)
+		if err != nil {
+			a.log.Error("Could not find requested PlanID %s in plan name lookup table", req.PlanID)
+			return nil, ErrorPlanNotFound
+		}
+	}
+
+	// Retrieve from/to plans by name, else respond with appropriate error
+	if fromPlan = spec.GetPlan(fromPlanName); fromPlan == nil {
```
jmrodri: I wasn't aware this syntax was legal :)

spadgett:
> I think the apb needs to have this field set on its metadata. I expect the catalog to look at that before calling update.

The web console will only allow you to change plans if planUpdatable is true.

jmrodri:
@spadgett thanks for the confirmation that's what I suspected based on the spec. So we'll fix it.

PR #445, opened 2017-09-20
Bearer Token Auth via kubernetes Apiserver

shawn-hurley:
Sets up ASB with the [apiserver](https://github.com/kubernetes/apiserver) kubernetes project. 
 - Serve ASB via [apiserver](https://github.com/kubernetes/apiserver) 
 - Set up delegated auth using [nonResoureceURLs](https://kubernetes.io/docs/admin/authorization/rbac/#role-examples) to control access
 - Setting `servicecatalog.ServiceBroker` with bearer token auth from a secret generated for a [service account](https://kubernetes.io/docs/admin/service-accounts-admin/#to-create-additional-api-tokens)
 - Will generate certs if no certs are given, therefore removing `insecure` argument.

depends-on <PR>
https://github.com/fusor/catasb/pull/142

Will neeed another PR for Documentation update.


pkg/clients/kubernetes.go
```diff
@@ -24,7 +24,7 @@ import (
 	"errors"
 
 	logging "github.com/op/go-logging"
-	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/rest"
```
rthallisey: can you move this down with the other k8s.io imports

pkg/handler/handler.go
```diff
@@ -94,33 +94,34 @@ func createVarHandler(r VarHandler) GorillaRouteHandler {
...
+func NewHandler(b broker.Broker, log *logging.Logger, brokerConfig broker.Config, prefix string) http.Handler {
 	h := handler{
 		router:       *mux.NewRouter(),
 		broker:       b,
 		log:          log,
 		brokerConfig: brokerConfig,
 	}
+	s := h.router.PathPrefix(prefix).Subrouter()
 
 	// TODO: Reintroduce router restriction based on API version when settled upstream
```
rthallisey: This covers the TODO here.

pkg/handler/handler_test.go
```diff
@@ -145,14 +145,14 @@ func init() {
 
 func TestNewHandler(t *testing.T) {
 	testb := MockBroker{Name: "testbroker"}
-	testhandler := NewHandler(testb, log, brokerConfig)
+	testhandler := NewHandler(testb, log, brokerConfig, "")
```
rthallisey: It might be worth adding a prefix to the test.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -197,18 +218,45 @@ objects:
...
+      kubernetes.io/service-account.name: ansibleservicebroker-client
+  type: kubernetes.io/service-account-token
 
 - apiVersion: servicecatalog.k8s.io/v1alpha1
   kind: ${BROKER_KIND}
   metadata:
     name: ansible-service-broker
   spec:
-    url: ${ASB_SCHEME}://asb-1338-ansible-service-broker.${ROUTING_SUFFIX}
+    url: https://asb.ansible-service-broker.svc:1338/ansible-service-broker/
```
rthallisey: This makes it a requirement to use https.  I'm not against that, but then ASB_SCHEME should be removed. 

Another way to do this is keep ASB_SCHEME and BROKER_AUTH in place but change BROKER_AUTH to:
```yaml
value: '{ "bearer": { "secretRef": { "kind": "Secret", "namespace": "ansible-service-broker", "name": "ansible-service-broker" } } }'
```

rthallisey: You also aren't using ```BROKER_URL_PREFIX```

pkg/app/app.go
```diff
@@ -24,12 +24,21 @@ import (
...
 	"encoding/json"
 	"fmt"
-	"net/http"
+	"net"
 	"os"
+	"strings"
 	"time"
 
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
```
rthallisey: Nit: Can you organize this so that imports with a name are together and those without a name are together.

```go
kapierrors "k8s.io/apimachinery/pkg/api/errors"
kubeversiontypes "k8s.io/apimachinery/pkg/version"
...

"k8s.io/apimachinery/pkg/runtime/serializer"
"k8s.io/apimachinery/pkg/util/wait"
...

File: pkg/app/app.go [hunk 6]
```diff
@@ -228,24 +301,31 @@ func (a *App) Start() {
...
+
+	var clusterURL string
+	if a.config.Broker.ClusterURL != "" {
+		if !strings.HasPrefix("/", a.config.Broker.ClusterURL) {
+			clusterURL = "/" + a.config.Broker.ClusterURL
+		} else {
+			clusterURL = a.config.Broker.ClusterURL
+		}
+	} else {
+		clusterURL = ClusterURLPreFix
```
rthallisey: Nit: Can you change ClusterURLPreFix to something like DefaultClusterURLPrefix

pkg/app/app.go
```diff
@@ -228,24 +301,31 @@ func (a *App) Start() {
...
 	}
+	daHandler := handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL)
+
+	genericserver.Handler.NonGoRestfulMux.HandlePrefix(fmt.Sprintf("%v/", clusterURL), daHandler)
+	a.log.Notice("Listening on https://%s", genericserver.SecureServingInfo.BindAddress)
+
+	a.log.Notice("Starting apiserver")
```
rthallisey: Shouldn't we note the Broker is started here? ```a.log.Notice("Ansible Service Broker Started")```

pkg/app/app.go
```diff
@@ -228,24 +301,31 @@ func (a *App) Start() {
...
+	err = genericserver.PrepareRun().Run(wait.NeverStop)
+	a.log.Errorf("unable to wait on run - %v", err)
+
+	//TODO: Add Flag so we can still use the old way of doing this.
```
rthallisey: I don't this it's necessary to also support the old way.

pkg/handler/handler.go
shawn-hurley: can you explain?

rthallisey: I thought 104 covered this todo. But, I think I'm confusing a Headers and a pathprefix. They aren't related.

pkg/app/app.go
shawn-hurley: I mean it does not actually start until the next line and could fail. I would prefer to make it `a.log.Notice("Ansible Service Broker Starting")`? is that agreeable?

pkg/app/app.go
shawn-hurley: go imports organize the imports like this I would prefer to continue to use go imports.

Does anyone else have any thoughts?

rthallisey: It's the style we've been using. I don't mind changing it.

pkg/app/args.go
```diff
@@ -30,7 +30,6 @@ import (
 type Args struct {
 	ConfigFile string `short:"c" long:"config" description:"Config File" default:"/etc/ansible-service-broker/config.yaml"`
 	Version    bool   `short:"v" long:"version" description:"Print version information"`
-	Insecure   bool   `long:"insecure" description:"Run Ansible Service Broker in insecure mode."`
```
rthallisey: This is going to require a change in [catasb](https://github.com/fusor/catasb/blob/78c4e35e45a61dcc953ef018d90be77220e1298a/ansible/roles/ansible_service_broker_setup/defaults/main.yml#L31).

shawn-hurley: I have the branch in my repo and will be posting a PR soon.

rthallisey:
@shawn-hurley I'm going to mess around with this a little by hand and get back to you.  I also want to make sure the gate passes on this which will require https://github.com/openshift/ansible-service-broker/pull/446

pkg/app/app.go
jmrodri: we could name them all and goimports will be fine. The problem with changing it is that @shawn-hurley and myself will flip it back when we save the file again. Not sure if there's a way to configure goimports. And I'd hate to stop using it since it does so much for me automatically.

rthallisey:
@jmrodri How close is your alternative to working? I'd like to evaluate both options.

jwmatthews:
@rthallisey it's been decided to go with Shawn's approach for delegated auth.  This is what the project needs at this point in time.

We are not pursuing the alternative approach.

jmrodri:
@rthallisey like @jwmatthews  mentioned it's been decided for us. I've deleted the bearerauth branch I had going.

jmrodri:
@shawn-hurley does basic auth still work?

shawn-hurley:
@jmrodri adding the ability that if basic auth if set, then we will use that instead of bearer token auth. This makes sense because if you set both authentication types on the `ServiceBroker` it will honor basic auth first. 

I still think that our default should be bearer token auth.

jwmatthews:
I expect default workflow will want Bearer Token Auth as well

enj:
LGTM

There are some followup items but this is in line with our discussions.

pkg/app/app.go
```diff
@@ -314,16 +313,16 @@ func (a *App) Start() {
...
-		clusterURL = ClusterURLPreFix
+		clusterURL = defaultClusterURLPreFix
 	}
 	daHandler := handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL)
 
 	genericserver.Handler.NonGoRestfulMux.HandlePrefix(fmt.Sprintf("%v/", clusterURL), daHandler)
 	a.log.Notice("Listening on https://%s", genericserver.SecureServingInfo.BindAddress)
 
-	a.log.Notice("Starting apiserver")
+	a.log.Notice("Ansible Serivce Broker Starting")
```
jmrodri: TYPO: Serivce -> Service

rthallisey:
Coorisponding catasb change: https://github.com/fusor/catasb/pull/142

PR #447, opened 2017-09-20
The docker organization name was changed in catasb

rthallisey:
dockerhub_org_name was changed to dockerhub_org.


rthallisey:
This will fix the error seen in #446 . And #446 should get the gate back to green.

PR #450, opened 2017-09-25
removing proposal that never happened

jmrodri:
Proposal is so wildly different from implementation, it will likely cause confusion.


jmrodri:
Might be this document: https://docs.google.com/document/d/1NvHiTl3zk_x2xxJM_moA6ky7tzGjnaw8aJABoTc4AoI/edit

PR #454, opened 2017-09-26
[Do Not Merge] Added versioning check when adding images

dymurray:
This PR implements version checking in the broker.

 - A new min/max version defined in a new package version
 - During image->spec process it does a validation check on the version

Note:
Ideally we will break this out of the adapter pkg and into the registry pkg so we aren't requiring new adapters to use imageToSpec. I will add this as an open issue.


pkg/version/version.go
```diff
@@ -1,7 +1,10 @@
-package app
+package version
 
 // Version contains the broker version.
 // DO NOT EDIT GENERATED FILE.
 // This file is generated by a tito tag
 // using templates/version.go.tmpl.
 var Version = "1.0.4"
+var MinAPBVersion = "1.0"
+var MaxAPBVersion = "1.0"
```
jmrodri: We might consider having an `apbversion.go` that contains the 2 values. Especially if they aren't filled in by tito.

So keep new `version` package. Move `version.go` like you did. Keep version.go just Version. Then add an `apbversion.go`.

.tito/tito.props
```diff
@@ -6,4 +6,4 @@ changelog_format = %s (%ae)
 
 [version_template]
 template_file = ./templates/version.go.tmpl
-destination_file = ./pkg/app/version.go
+destination_file = ./pkg/version/version.go
```
jmrodri: +1

pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +141,25 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
 
 	return spec, nil
 }
+
+func isCompatibleVersion(specVersion string, minVersion string, maxVersion string) bool {
+	specFloat, err := strconv.ParseFloat(specVersion, 64)
```
jmrodri: Create a unit test for this because I'm getting a sense this could be bad with bad input. This implementation won't allow 1.0.alpha or anything alphanumeric. So let's test with different inputs and expected errors etc.

pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +141,25 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
...
+	if err != nil {
+		return false
+	}
+
+	minFloat, err := strconv.ParseFloat(minVersion, 64)
```
jmrodri: the `ParseFloat` is raising my spidey senses but not sure the alternative is.

dymurray:
Closing this PR because I am way lost on what I did to it

PR #456, opened 2017-09-27
Remove provision parameters from being reused as binding parameters.

cfchase:
Currently, broker is listing all parameters intended for provision as binding parameters when creating a binding.  Until binding parameters in the broker is implemented, we need to remove these parameters as they are not currently used and add confusion to the user experience.

Removes the service_binding.create paramaters and openshift_form_definition from the catalog request.


jmrodri:
I thought this was put in for @jmontleon work.

PR #457, opened 2017-09-27
Added versioning check to Broker on bootstrap

dymurray:
This PR adds version checks to the broker when bootstrapping an APB

 - Move version.go to its own version pkg
 - Add version check on imageToSpec conversion
 - Add unit tests for compatible version checks

Note:
I plan to open an issue to move the version to the spec itself so that this check moves out of the adapter pkg.

### DIFF
--- a/pkg/registries/adapters/adapter.go
+++ b/pkg/registries/adapters/adapter.go
@@ -26,9 +26,12 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
+	"strings"
 
 	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
+	"github.com/openshift/ansible-service-broker/pkg/version"
 	yaml "gopkg.in/yaml.v1"
 )
 
@@ -111,6 +114,15 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
 		log.Infof("Didn't find encoded Spec label. Assuming image is not APB and skiping")
 		return nil, nil
 	}
+	if conf.Config.Label.Version == "" {
+		log.Infof("Didn't find encoded Version label. Assuming image is not APB and skipping")
+		return nil, nil
+	}
+	if isCompatibleVersion(conf.Config.Label.Version, version.MinAPBVersion, version.MaxAPBVersion) != true {
+		log.Infof("APB spec version was incompatible. Assuming image is incompatible and skipping")
+		return nil, nil
+	}
+
 	encodedSpec := conf.Config.Label.Spec
 	decodedSpecYaml, err := b64.StdEncoding.DecodeString(encodedSpec)
 	if err != nil {
@@ -129,4 +141,30 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
 	log.Debugf("Successfully converted Image %s into Spec", spec.Image)
 
 	return spec, nil
+}
+
+func isCompatibleVersion(specVersion string, minVersion string, maxVersion string) bool {
+	specMajorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[0])
+	specMinorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[1])
+	minMajorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[0])
+	minMinorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[1])
+	maxMajorVersion, err := strconv.Atoi(strings.Split(maxVersion, ".")[0])
+	maxMinorVersion, err := strconv.Atoi(strings.Split(maxVersion, ".")[1])
+	if err != nil {
+		return false
+	}
+	if minMajorVersion == maxMajorVersion {
+		if specMajorVersion >= minMajorVersion && specMajorVersion <= maxMajorVersion {
+			if specMinorVersion >= minMinorVersion && specMinorVersion <= maxMinorVersion {
+				return true
+			}
+		}
+	} else if minMajorVersion < maxMajorVersion {
+		if specMajorVersion == minMajorVersion && specMinorVersion >= minMinorVersion {
+			return true
+		} else if specMajorVersion == maxMajorVersion && specMinorVersion <= maxMinorVersion {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/registries/adapters/adapter.go
+++ b/pkg/registries/adapters/adapter.go
@@ -147,15 +147,7 @@ func isCompatibleVersion(specVersion string, minVersion string, maxVersion strin
 	if err != nil {
 		return false
 	}
-	specMinorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[1])
-	if err != nil {
-		return false
-	}
 	minMajorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[0])
-	if err != nil {
-		return false
-	}
-	minMinorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[1])
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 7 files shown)


pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +142,29 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
...
+
+func isCompatibleVersion(specVersion string, minVersion string, maxVersion string) bool {
+	specMajorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[0])
+	specMinorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[1])
+	minMajorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[0])
+	minMinorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[1])
+	maxMajorVersion, err := strconv.Atoi(strings.Split(maxVersion, ".")[0])
+	maxMinorVersion, err := strconv.Atoi(strings.Split(maxVersion, ".")[1])
+	if err != nil {
+		return false
```
dymurray: Oh crap, this is a typo. Let me fix this.

pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +142,45 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
 
 	return spec, nil
 }
+
+func isCompatibleVersion(specVersion string, minVersion string, maxVersion string) bool {
+	specMajorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[0])
+	if err != nil {
+		return false
+	}
+	specMinorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[1])
```
shawn-hurley: This could panic when I have set the version to `1` only.

pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +142,45 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
...
+		return false
+	}
+
+	if minMajorVersion == maxMajorVersion {
+		if specMajorVersion >= minMajorVersion && specMajorVersion <= maxMajorVersion {
+			if specMinorVersion >= minMinorVersion && specMinorVersion <= maxMinorVersion {
+				return true
+			}
+		}
+	} else if minMajorVersion < maxMajorVersion {
```
shawn-hurley: I think this is going to return false for `isCompatibleVersion("2.5", "1.0", "3.6")`
I don't know if this use case make sense or not.

pkg/version/apbversion.go
```diff
@@ -0,0 +1,4 @@
+package version
+
+var MinAPBVersion = "1.0"
```
shawn-hurley: I think that we may want these to be `const` rather than `var`.
Also, anything that is exported from the package needs a comment.

pkg/registries/adapters/adapter.go
dymurray: +1 good catch.

pkg/registries/adapters/adapter.go
dymurray: +1 another good catch

dymurray:
@shawn-hurley Thanks for the feedback. Just making it known that this will always fail CI until the APBs are updated with the proper version which I am working on doing today as well.

dymurray:
Confirmed when running catasb with the `canary` tag for images that are being built off recent commits that they are passing version validation.

jmrodri:
Linter errors:
```
pkg/version/apbversion.go:3:1: comment on exported const MinAPBVersion should be of the form "MinAPBVersion ..."

pkg/version/apbversion.go:7:7: exported const MaxAPBVersion should have comment or be unexported

Found 2 lint suggestions; failing.

make: *** [lint] Error 1
```

jmrodri:
You need to update the `version.go.tmpl` package as well.

jmrodri:
@dymurray forget my CI comment then, I see your comment about being dependent on apbs. But the linter is still broken so that comment still stands.

dymurray:
CI is failing on binding... unsure how my PR affects binding of mediawiki to postgres. I have confirmed the latest images on dockerhub have the proper version and the broker is bootstrapping them.

pkg/registries/adapters/adapter.go
```diff
@@ -130,3 +138,50 @@ func imageToSpec(log *logging.Logger, req *http.Request, apbtag string) (*apb.Sp
...
+	}
+	specMinorVersion, err := strconv.Atoi(strings.Split(specVersion, ".")[1])
+	if err != nil {
+		return false
+	}
+	minMajorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[0])
+	if err != nil {
+		return false
+	}
+	minMinorVersion, err := strconv.Atoi(strings.Split(minVersion, ".")[1])
```
rthallisey: How come we have to check minor versions? I thought we accept APBVersions >= 1.0 and < 2.0.

dymurray: We need to check minor versions because of the scenario where we support beyond 2.0. We need a way to check that the version fits outside of one major version bump.

pkg/version/apbversion.go
```diff
@@ -0,0 +1,11 @@
+package version
+
+// These constants describe the minimum and maximum
+// accepted APB spec versions. They are used to filter
+// acceptible APBs.
+
+// MinAPBVersion constant to describe minimum supported spec version
+const MinAPBVersion = "1.0"
```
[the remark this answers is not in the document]
dymurray: We *do* have a requirement for backwards compatibility we just have not gotten to a point where it is a problem yet. Also, I have updated the proposal with the notes we took on the 9/26 meeting. The broker version is not tied directly to the APB spec version. We want brokers to have a range of APB versions they support and not have to bump a release every time something changes in the broker when the APB spec itself has not changed.

pkg/registries/adapters/adapter.go
rthallisey: So you're saying we'd support >= 1.0 and < 2.x? If we do that then we can't really say that _only_ a major version bump will break the API.  IMO, that goes against the reason to have major vs minor versioning.

dymurray: I'm not sure how we can't say that? A broker that only knows how to deploy 1.X APBs will never be able to support 2.X APBs. However, I would expect a broker that knows how to deploy 2.X APBs to know how to deploy 1.X APBs as that would be supporting backwards compatibility. In this case the major version bump *does* break the API, it's just that we expect the broker to support 1.X APBs even as we move to a 2.X spec.

rthallisey: > In this case the major version bump does break the API, it's just that we expect the 
> broker to support 1.X APBs even as we move to a 2.X spec.
>
👍 

Here's an example of what I'm thinking:
- We move to APBVersion 2.3.
- APBVersion >= 1.0 and APBVersion < 3.0 are supported
   
Why would we need to check the minor version when we know that anything greater than or equal to the minimal major version and less than the maximum major version is valid? In other words, all breaking changes are per major version. If we support 1.0 then we also support 1.7 and 1.x so there's no reason to check beyond the major version.

dymurray: I see. Using your example, let's take a broker that is installed with 2.3 APBs. We then release a new version 2.4 (and while the upgrade to 2.4 will *not* be a breaking change, it would possibly expose a new option or spelling change). 

I see a use case where an installed broker will not support the new 2.4 option and may not want to deploy APBs that expose that option. It is still possible for the broker to be configured to have 1.0-3.0 as an acceptable range but gives the administrator the ability to fine tune what versions they want to accept and list. I don't see this being an immediate problem but I see this as something we will have to deal with as different versions of the service-catalog expose different metadata options. These are examples of non breaking changes but something the administrator may want to control.

It seems strange to me that we wouldn't be doing any validation on the minor version because we could expose new options that are *not* API breaking but might not be used in a 2.2 broker for example.

dymurray: Unless you feel that my example would imply a major version bump... then I could see why it would be trivial to check the minor version. I just worry about not checking the minor version at all and someone putting 1.Z which doesn't exist and has no associated schema.

rthallisey: > I just worry about not checking the minor version at all and someone putting 1.Z which doesn't 
> exist and has no associated schema.
>
I understand your concern, but if we follow the release plan correctly then we won't have this issue. And if we make a mistake then we can always change the version.

shawn-hurley: I wouldn't mind having Ryan's logic be the default, but if a user has defined some config then we override? This might just be a feature we add later and for now, just follow Ryan's logic for now.

dymurray: +1 sounds good to me

jmrodri: +1 @rthallisey example is how I mentally pictured how this would work. The scenario of 2.4 causing a problem means IMO we failed with the change APB spec change.

PR #462, opened 2017-10-02
Pass in args to the deploy scripts

rthallisey:
The gate showed a false positive on #434, this will fix it.


scripts/deploy.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
 
 BROKER_ROOT=$(dirname "${BASH_SOURCE}")/..
+ARGS="${@}"
```
djzager: Probably helps to include the args :+1:

PR #464, opened 2017-10-02
Creating proposals for keeping transient namespace alive

shawn-hurley:
Proposal for [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766)

### DIFF
--- a/docs/proposals/keeping-transient-pod-name-spaces-alive.md
+++ b/docs/proposals/keeping-transient-pod-name-spaces-alive.md
@@ -4,7 +4,7 @@
 The goal of this proposal is to give administrators options to keep namespaces and therefore APB pods after the execution. The use cases for this feature are demos and debugging APBs.
 
 ## Problem Description
-The [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) was created when we moved to creating transient namespaces during execution of the APB pod. What happens  This bug creates issues for debugging APB's as well as issues with demos. 
+[Bug 1497766](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) titled "APB Pods are deleted even when an error occurs" occurred when we moved to creating transient namespaces during the APB pod execution. This bug makes it difficult to debug APB development, debug issues in a production cluster using ASB, and creating demos because the pods do not live long enough to capture useful information.
 
 ## Implementation Details.
 
@@ -23,7 +23,7 @@ openshift:
 ...
 ```
 
-**NOTE: Both will default to false in openshift-ansible, but will be set to true for keep_namespace_on_error**
+**NOTE: Both configuration options will default to false in [openshift-ansible](https://github.com/openshift/openshift-ansible). When Deploying with CATASB or run_latest.sh we will set keep_namespace_on_error to true**
 
 ### Major Code Change
 
@@ -39,11 +39,27 @@ pod, err := k8scli.CoreV1().Pods(executionContext.Namespace).Get(executionContex
 if err != nil {
         s.log.Errorf("Unable to retrieve pod - %v", err)     
 }
-if !brokerConfig.keepNamespace || !((pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || err != nil) && brokerConfig.keepNamespaceOnError) {
+if shouldDeleteNamespace(clusterConfig, pod, err)  {
     ... Delete Namespace 
 }
 ...Delete role bindings.
 ```
+
+```golang
+func shouldDeleteNamespace(clusterConfig ClusterConfig, pod *apiv1.Pod, getPoderr error) bool {
+    if clusterConfig.KeepNamespaces {
+        return false
+    }
+    if clusterConfig.KeepNamespacesOnError {
+        if pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || getPoderr != nil {
+           return false
+        }
+    }
+    return true
+}
+```
+
+The outcome of keeping a namespace alive is that the RoleBindings and Service Accounts should be deleted. The only thing that should be left is the pod and the namespace.
 
 ## Work Items
 - Add Code Above
@@ -51,4 +67,3 @@ if !brokerConfig.keepNamespace || !((pod.Status.Phase == apiv1.PodFailed || pod.
 - update the deployed template to set default  values
 - CATASB change to allow for overriding the default values.
 - doc updates for config, deployment
- 
--- a/docs/proposals/keeping-transient-pod-name-spaces-alive.md
+++ b/docs/proposals/keeping-transient-pod-name-spaces-alive.md
@@ -10,8 +10,18 @@ The [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) was created when 
 
 ### Configuration Values
 
-* `keep_namespace` or `BrokerConfig.KeepNamespace` - Parameter to always keep name space no matter what.
-* `keep_namespace_on_error` or `BrokerConfig.KeepNamespaceOnError` -  parameter to keep name space around if an error occurs in the play book that is running.
+* `keep_namespace` or `ClusterConfig.KeepNamespace` - Parameter to always keep name space no matter what.
+* `keep_namespace_on_error` or `ClusterConfig.KeepNamespaceOnError` -  parameter to keep name space around if an error occurs in the play book that is running.
+
+Example:
+```yaml
+...
+openshift:
+  ....
+  keep_namespace: false
+  keep_namespace_on_error: true
+...
+```
 
 **NOTE: Both will default to false in openshift-ansible, but will be set to true for keep_namespace_on_error**
 
@@ -29,7 +39,7 @@ pod, err := k8scli.CoreV1().Pods(executionContext.Namespace).Get(executionContex
 if err != nil {
         s.log.Errorf("Unable to retrieve pod - %v", err)     
 }
-if !brokerConfig.keepNamespace || ((pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || err != nil) && !brokerConfig.keepNamespaceOnError {
+if !brokerConfig.keepNamespace || !((pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || err != nil) && brokerConfig.keepNamespaceOnError) {
     ... Delete Namespace 
 }
 ...Delete role bindings.
@@ -40,4 +50,5 @@ if !brokerConfig.keepNamespace || ((pod.Status.Phase == apiv1.PodFailed || pod.S
 - Add broker config values above during creation of Service Account.
 - update the deployed template to set default  values
 - CATASB change to allow for overriding the default values.
+- doc updates for config, deployment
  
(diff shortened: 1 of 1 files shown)


docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,43 @@
...
+The goal of this proposal is to give administrators options to keep namespaces and therefore APB pods after the execution. The use cases for this feature are demos and debugging APBs.
+
+## Problem Description
+The [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) was created when we moved to creating transient namespaces during execution of the APB pod. What happens  This bug creates issues for debugging APB's as well as issues with demos. 
+
+## Implementation Details.
+
+### Configuration Values
+
+* `keep_namespace` or `BrokerConfig.KeepNamespace` - Parameter to always keep name space no matter what.
```
shawn-hurley: I actually think that this should be in the `openshift`/clusterConfig section of the configuration.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
...
+```golang
+....
+pod, err := k8scli.CoreV1().Pods(executionContext.Namespace).Get(executionContext.PodName, metav1.GetOptions{})
+if err != nil {
+        s.log.Errorf("Unable to retrieve pod - %v", err)     
+}
+if !brokerConfig.keepNamespace || !((pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || err != nil) && brokerConfig.keepNamespaceOnError) {
+    ... Delete Namespace 
+}
+...Delete role bindings.
```
djzager: So the only thing(s) that should survive when we are keeping the namespace is 1) the namespace and 2) the pod? I imagine we would want to remove everything else.

rthallisey:
How much security are we gaining by having an ephemeral namespace versus a locked down persistent namespace for APBs?

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
+# Keeping transient pod name spaces alive
+
+## Introduction
+The goal of this proposal is to give administrators options to keep namespaces and therefore APB pods after the execution. The use cases for this feature are demos and debugging APBs.
+
+## Problem Description
+The [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) was created when we moved to creating transient namespaces during execution of the APB pod. What happens  This bug creates issues for debugging APB's as well as issues with demos.
```
jmrodri: TYPO: "What happens  This bug"  -> This bug creates ... 
Also what sort of problems are created? How about:

[Bug 1497766](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) titled "APB Pods are deleted even when an error occurs" occurred when we moved to creating transient namespaces during the APB pod execution. This bug makes it difficult to debug APB development and creating demos because the pods do not live long enough to capture useful information.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
...
+```yaml
+...
+openshift:
+  ....
+  keep_namespace: false
+  keep_namespace_on_error: true
+...
+```
+
+**NOTE: Both will default to false in openshift-ansible, but will be set to true for keep_namespace_on_error**
```
jmrodri: I don't quite understand the above NOTE. both will be false, but set to true for namespace_on_error. HUH? 😕

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Problem Description
+The [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497766) was created when we moved to creating transient namespaces during execution of the APB pod. What happens  This bug creates issues for debugging APB's as well as issues with demos. 
+
+## Implementation Details.
+
+### Configuration Values
+
+* `keep_namespace` or `ClusterConfig.KeepNamespace` - Parameter to always keep name space no matter what.
+* `keep_namespace_on_error` or `ClusterConfig.KeepNamespaceOnError` -  parameter to keep name space around if an error occurs in the play book that is running.
```
jmrodri: Have we considered a simple TTL? Keep pods alive for x amount of time, then possibly have a reaper job to clean things up. The TTL, say 1 day, would allow customer support to debug a pod that had an issue.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+The logic will be in the `pkg/apb/svc_account.go` file in the `DestroySandbox` method:
+```golang
+....
+pod, err := k8scli.CoreV1().Pods(executionContext.Namespace).Get(executionContext.PodName, metav1.GetOptions{})
+if err != nil {
+        s.log.Errorf("Unable to retrieve pod - %v", err)     
+}
+if !brokerConfig.keepNamespace || !((pod.Status.Phase == apiv1.PodFailed || pod.Status.Phase == apiv1.PodUnknown || err != nil) && brokerConfig.keepNamespaceOnError) {
```
jmrodri: Is there a nicer way to phrase this if statement? I have to read it several times to make ensure a particular state triggers it.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
...
+    ... Delete Namespace 
+}
+...Delete role bindings.
+```
+
+## Work Items
+- Add Code Above
+- Add broker config values above during creation of Service Account.
+- update the deployed template to set default  values
```
jmrodri: remove the extra space before values

jmrodri: The work items sound sane, depends on the actual configuration options we decide on.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
jwmatthews: If we are configured to keep the namespace, I assume we want to keep everything in it we created to aid in debugging.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
eriknelson: When I implemented the original sandboxes, there was an urgency around: the sandbox MUST be deleted following an APB method execution. I think there was a security concern with keeping resources like ServiceAccounts and RoleBindings around longer than required. We could keep the pod around and delete the SA/RB, but that might violate the idea of the atomic "sandbox" (deleting some resources, but not others). Maybe it's okay because the Pod itself isn't considered part of it, but those are my concerns with a TTL.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
```diff
@@ -0,0 +1,54 @@
+# Keeping transient pod name spaces alive
+
+## Introduction
+The goal of this proposal is to give administrators options to keep namespaces and therefore APB pods after the execution. The use cases for this feature are demos and debugging APBs.
```
eriknelson: There's also a use-case for production cluster operators; I don't think this is limited to demo's or debugging APBs. APB's could certainly fail under production scenarios, and having nothing to examine after something unexpected takes place is undesirable.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
eriknelson: Maybe the TTL is a 3rd configuration option?

docs/proposals/keeping-transient-pod-name-spaces-alive.md
eriknelson: Abstract it behind a function so the logic can be tweaked without altering the if?

```
 if shouldDeleteNamespace() {...}

func shouldDeleteNamespace{
    if brokerConfig.KeepNamespace {
        return false
    }
    // ... other conditions
}
```

docs/proposals/keeping-transient-pod-name-spaces-alive.md
eriknelson: Are we concerned about leaking role related resources like SA's and RoleBindings?

shawn-hurley: We are removing the role bindings, the thinking is that you can then give other users access to the new namespace, they can debug the APB without getting elevated permissions into the target namespace. 

@rthallisey 
> How much security are we gaining by having an ephemeral namespace versus a locked down persistent namespace for APBs?

The above is the added security that we will get.

jmrodri:
@eriknelson I like the abstraction behind shouldDeleteNamespace that will help.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
jmrodri: how do we help the production operator be able to debug things like this? would the workflow be if things fail, reconfigure the broker to keep the namespaces around, and retry?

jmrodri:
@eriknelson if we NEED to remove the namespaces immediately for security purposes, then I would assume the TTL won't really help. I was thinking the TTL would replace the other configs not be a 3rd.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
shawn-hurley: Was totally in my one head on this one, will make it more clear

docs/proposals/keeping-transient-pod-name-spaces-alive.md
shawn-hurley: I think this a great idea, but it strikes me as a feature rather than a bug fix.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
rthallisey: > We are removing the role bindings, the thinking is that you can then give other users access to the 
> new namespace, they can debug the APB without getting elevated permissions into the target 
> namespace.
>
>>    How much security are we gaining by having an ephemeral namespace versus a locked
>> down persistent namespace for APBs?
>>
> The above is the added security that we will get.
>
I disagree. The namespace is not the source for the vulnerability because by itself it's not dangerous. The vulnerability is with the rolebindings for the namespace. If we manage the rolebindings on every APB namespace, then I don't think there's any harm in keeping them around for at least a period of time.  This will allow folks debug a failure in their APBs or to read the output logs from a successful run.  I would much rather have a namespace deletion timer than having to check for APB pod failures in our code and require another config setting in the broker to enable persistent namespaces.

docs/proposals/keeping-transient-pod-name-spaces-alive.md
shawn-hurley: removing the extra space makes it not a list. 
example
-Hello

vs
- Hello

docs/proposals/keeping-transient-pod-name-spaces-alive.md
shawn-hurley: I don't disagree that the watcher that will delete them after a certain time is a good idea. We have captured that, and I think is just too much to implement in a bug fix. 

I have created a card on the backlog already:
https://trello.com/c/i5gUkUv4

PR #465, opened 2017-10-02
add 3.7 releaser to releasers.conf

jmrodri:
Had to add the 3.7 brew releaser so we can build against the 3.7 buildroot.


jmontleon:
ACK

PR #468, opened 2017-10-03
Bug 1498185 - Adjust versioning check so that it is done in the registry package

dymurray:
This PR adjusts for the change of moving the APB spec version from an image label into the Spec itself. It also moves the version check out of the adapter package so that it is now obfuscated out of each individual adapter.

 - Move APB spec version check to spec validation in registry package
 - Adds unit tests for specs without versions

This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1498185

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -61,6 +61,7 @@ type Plan struct {
 // Spec - A APB spec
 type Spec struct {
 	ID          string                 `json:"id"`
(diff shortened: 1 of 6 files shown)


dymurray:
FYI: This will fail travis until I update the upstream images so that they pass bootstrapping. Once I get approved reviews I will update the APBs and rerun travis.

Issue #385, opened 2017-08-21
APB pods should have descriptive names


djwhatle:
> # Feature Request: Use more descriptive names for APB pods

**What happened**:
 - Testing two APBs simultaneously (database and hello-world)
 - Running provision, bind, deprovision several times in same project
 - Want to check MySQL APB logs, but can't discern which APB to view logs for

```
NAME                                       READY     STATUS      RESTARTS   AGE
apb-040cb995-5c99-41b0-8c79-c1bb512d399e   0/1       Completed   0          2h
apb-10465d96-b7b6-46bb-858e-81236d8391de   0/1       Completed   0          2h
apb-11ed3a4d-ae70-4e9b-9787-9f042544bec0   0/1       Completed   0          2h
apb-dfe46c9e-89be-41b9-b7ce-07be8a1daf29   0/1       Completed   0          2h
docker-registry-1-lzwrb                    1/1       Running     0          2h
hello-world-1-3x7zf                        1/1       Running     0          2h
persistent-volume-setup-ct6k5              0/1       Completed   0          2h
router-1-8dls6                             1/1       Running     0          2h

```

**What you expected to happen**:
APBs have descriptive but unique names, similar to how normal pods are named in OpenShift. 
Example: `apb-hello-world-1-3x7zf`

Edit: It would be even more helpful to see the action that the APB pod is carrying out.
Example: `apb-provision-hello-world-1-lzwrb`

Perhaps the APB name can be used in the pod name as-is. Another option would be to have a "short name" parameter defined in apb.yml.

Relevant line in broker: https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/executor.go#L71

djwhatle:
Another idea: it would be helpful to see at a glance which action an APB pod is carrying out.

Example: `apb-provision-hello-world-1-lzwrb`

eriknelson:
We should be able to use a k8s construct: `metadata.generatedName` to address this:

https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#idempotency

> Names generated by the system may be requested using metadata.generateName. GenerateName indicates that the name should be made unique by the server prior to persisting it. A non-empty value for the field indicates the name will be made unique (and the name returned to the client will be different than the name passed). The value of this field will be combined with a unique suffix on the server if the Name field has not been provided.

[...]

PR #473, opened 2017-10-04
Bug 1497839 - copy secrets to transient namespace and always run

shawn-hurley:
Removes split code path based on secrets to run the APB in the transient namespace always
 - remove branch on secrets 
 - copy secrets to the transient namespace
 - panic the broker if we are attempting to delete its namespace.

depends-on <PR>
N/A
N/A

Fixes: #385

### DIFF
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -35,11 +35,6 @@ import (
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
 
-const (
-	transientNameSpaceKey        = "apb"
-	transientNameSpaceNamePrefix = "apb-namespace-"
-)
-
 // ExecuteApb - Runs an APB Action with a provided set of inputs
 func ExecuteApb(
 	action string,
@@ -83,12 +78,17 @@ func ExecuteApb(
 		return executionContext, err
 	}
 
+	labels := map[string]string{
+		"apb-fqname": spec.FQName,
+		"apb-action": action,
+	}
+
 	executionContext.Targets = append(executionContext.Targets, context.Namespace)
 	// Create namespace.
 	namespace := v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:       map[string]string{transientNameSpaceKey: spec.FQName},
-			GenerateName: transientNameSpaceNamePrefix,
+			Labels:       labels,
+			GenerateName: fmt.Sprintf("%s-%.4s-", spec.FQName, action),
 		},
 	}
 	ns, err := k8scli.CoreV1().Namespaces().Create(&namespace)
@@ -112,10 +112,8 @@ func ExecuteApb(
 	volumes, volumeMounts := buildVolumeSpecs(secrets)
 	pod := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: executionContext.PodName,
-			Labels: map[string]string{
-				"apb-fqname": spec.FQName,
-			},
+			Name:   executionContext.PodName,
+			Labels: labels,
 		},
 		Spec: v1.PodSpec{
 			Containers: []v1.Container{
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -203,10 +203,12 @@ func checkPullPolicy(policy string) (v1.PullPolicy, error) {
 	return value, nil
 }
 
-func copySecretsToNamespace(executionContext ExecutionContext,
+func copySecretsToNamespace(
+	executionContext ExecutionContext,
 	clusterConfig ClusterConfig,
 	k8scli *clientset.Clientset,
-	secrets []string) error {
+	secrets []string,
+) error {
 	for _, secrectName := range secrets {
 		secretData, err := k8scli.CoreV1().Secrets(clusterConfig.Namespace).Get(secrectName, metav1.GetOptions{})
 		if err != nil {
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -32,6 +32,7 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/pborman/uuid"
 	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/apb/executor.go
```diff
@@ -80,27 +81,28 @@ func ExecuteApb(
...
-				Name:   executionContext.Namespace,
-			},
-		}
-		_, err := k8scli.CoreV1().Namespaces().Create(&namespace)
-		if err != nil {
-			return executionContext, err
-		}
+
+	// Using a new UUID is sane, because it will have some gurantee
+	// of uniquenes and will meet DNS name requirements.
```
djzager: nit

Using a new UUID is sane, because it will have some **guarantee**
of **uniqueness** and will meet DNS name requirements.

pkg/apb/executor.go
```diff
@@ -80,27 +81,28 @@ func ExecuteApb(
...
+	executionContext.Namespace = uuid.New()
+	executionContext.Targets = append(executionContext.Targets, context.Namespace)
+	// Create namespace.
+	namespace := v1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{transientNameSpaceKey: spec.FQName},
+			Name:   executionContext.Namespace,
```
djzager: You could use `GenerateName`

https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#idempotency, but then you would have to get the name after you have created the namespace.

shawn-hurley: I think then we could give the namespace a name that is more obvious like apb-namespace-(unique suffix).

shawn-hurley:
[(2) Update AWS APB workflow on broker to match with transient namespace approach.](https://trello.com/c/PGDdRWD9/524-2-update-aws-apb-workflow-on-broker-to-match-with-transient-namespace-approach)

djzager:
Example Namespace:
```
- apiVersion: v1
  kind: Namespace
  metadata:
    annotations:
      openshift.io/sa.scc.mcs: s0:c13,c7
      openshift.io/sa.scc.supplemental-groups: 1000170000/10000
      openshift.io/sa.scc.uid-range: 1000170000/10000
    creationTimestamp: 2017-10-12T17:12:06Z
    generateName: dh-hello-world-apb-prov-
    labels:
      apb-action: provision
      apb-fqname: dh-hello-world-apb
    name: dh-hello-world-apb-prov-84s0z
    namespace: ""
    resourceVersion: "10004"
    selfLink: /api/v1/namespaces/dh-hello-world-apb-prov-84s0z
    uid: 7927f3e5-af70-11e7-a5b7-64006a6fb904
  spec:
    finalizers:
    - kubernetes
    - openshift.io/origin
  status:
    phase: Active
```

pkg/apb/executor.go
```diff
@@ -80,27 +77,31 @@ func ExecuteApb(
...
+		"apb-fqname": spec.FQName,
+		"apb-action": action,
 	}
+
+	executionContext.Targets = append(executionContext.Targets, context.Namespace)
+	// Create namespace.
+	namespace := v1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:       labels,
+			GenerateName: fmt.Sprintf("%s-%.4s-", spec.FQName, action),
```
eriknelson: What's the upper limit on the size of this field? 64 chars? We should guarantee we're under it.

pkg/apb/svc_acct.go
```diff
@@ -236,8 +237,11 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
 	if asbNamespace != executionContext.Namespace {
 		s.log.Debug("Deleting namespace %s", executionContext.Namespace)
 		k8scli.CoreV1().Namespaces().Delete(executionContext.Namespace, &metav1.DeleteOptions{})
-
+	} else {
+		//We should not be attempting to run pods in the ASB namespace, if we are, something is seriously wrong.
```
eriknelson: Nit: Space after //

pkg/apb/executor.go
```diff
@@ -201,3 +200,22 @@ func checkPullPolicy(policy string) (v1.PullPolicy, error) {
 
 	return value, nil
 }
+
+func copySecretsToNamespace(executionContext ExecutionContext,
```
eriknelson: Nit: We have a couple stylistic patterns for long function signatures in the codebase right now. Admittedly, it's not super consistent, but we should try to not introduce other forms as well IMO. Personally, I like a vertically favored format:

```
func copySecretsToNamespace(
	executionContext ExecutionContext,
	clusterConfig ClusterConfig,
	k8scli *clientset.Clientset,
	secrets []string,
) error { // Return values
	// Implementation
}
```

Also a longer form format, wrapping after the last full var/type crosses something like 80 chars:

```
func copySecretsToNamespace(executionContext ExecutionContext, clusterConfig ClusterConfig,
	k8scli *clientset.Clientset, secrets []string) error {
	// Implementation
}
```

I would just advocate picking one and sticking with it.

pkg/apb/svc_acct.go
```diff
@@ -236,8 +237,11 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
...
+		panic(fmt.Errorf("Broker is attempting to delete its own namespace"))
```
eriknelson: +1, let's not implode :tada:

eriknelson:
@shawn-hurley @djzager meant to request changes, apparently I can't operate github.

pkg/apb/executor.go
djzager: I can make this change if you like, but my understanding is that GenerateName handles this for you:

Names generated by the system may be requested using metadata.generateName. GenerateName indicates that the name should be made unique by the server prior to persisting it. A non-empty value for the field indicates the name will be made unique (and the name returned to the client will be different than the name passed). The value of this field will be combined with a unique suffix on the server if the Name field has not been provided. The provided value must be valid within the rules for Name, and **may be truncated by the length of the suffix required to make the value unique on the server**. If this field is specified, and Name is not present, the server will NOT return a 409 if the generated name exists - instead, it will either return 201 Created or 504 with Reason ServerTimeout indicating a unique name could not be found in the time allotted, and the client should retry (optionally after the time indicated in the Retry-After header).

From https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#idempotency

eriknelson: No change needed, that looks good! Thank you for the link :+1:

eriknelson:
doublecomboACK

Issue #391, opened 2017-08-23
Auto bootstrap is clearing apb-push sourced specs


eriknelson:
**What happened**:
APB developer with no existing APB's in their Dockerhub org wrote up a test APB, ran apb push, and the broker loaded it into it's `/spec` correctly. After waiting 5min, service still did not show up. Auto-bootstrap kicked in and cleared **all** specs from `/spec`, so it destroyed the one sourced from `apb-push`.
 
Log: [broker.txt](https://github.com/openshift/ansible-service-broker/files/1246547/broker.txt)

**What you expected to happen**:
Auto bootstrap should not be clearing apb-push sourced specs.

Longer term, we should be able to enable/disable, and configure auto-bootstrap on a registry basis. It should be disabled for the apb-push type registry adapter by default.

eriknelson:
Related #49

jmrodri:
Do we disable auto bootstrap in dev mode?

dymurray:
Working on a PR for this, associated bug is here: https://bugzilla.redhat.com/show_bug.cgi?id=1498933

dymurray:
https://github.com/openshift/ansible-service-broker/pull/477

PR #477, opened 2017-10-05
 Bug 1498933 - Do not delete apb-push sourced specs when bootstrapping 

dymurray:
This PR checks the list of bootstrapped specs for apb-push sourced specs before deleting the entire set of APBs. It then saves the apb-pushed specs to a new array, and adds them back in after the bootstrap. We need this because without apb-push sourced specs are being deleted.

 - Check for apb-push sourced specs, if they exist save them and append them to spec list after the bootstrap

fixes https://github.com/openshift/ansible-service-broker/issues/391


dymurray:
https://github.com/openshift/ansible-service-broker/issues/391

PR #480, opened 2017-10-10
Bug 1500048 - make plan ids globally unique

jmrodri:

With a recent service-catalog change, the serviceplans were moved to top level. We were using the plan.name as the plan.id which was not a problem before because they were namespaced by the serviceclass.

Now that the serviceplans are top level, the names need to be unique. Since we used to rely on the serviceclass name, the new id will be a UUID. We are now storing the id and name in etcd.

```
/plan_names/UUID: plan.name
```

fixes 1500048

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -50,6 +50,7 @@ type ParameterDescriptor struct {
 
 // Plan - Plan object describing an APB deployment plan and associated parameters
 type Plan struct {
+	ID             string                 `json:"id" yaml:"-"`
 	Name           string                 `json:"name"`
 	Description    string                 `json:"description"`
 	Metadata       map[string]interface{} `json:"metadata,omitempty"`
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -319,16 +319,20 @@ func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 
 		// update the id on the plans, doing it here avoids looping through the
 		// specs array again
-		addIDForPlan(spec.Plans)
+		addIDForPlan(spec.Plans, spec.FQName)
 	}
 }
 
 // addIDForPlan - for each of the plans create a new ID
-func addIDForPlan(plans []apb.Plan) {
+func addIDForPlan(plans []apb.Plan, FQSpecName string) {
 
 	// need to use the index into the array to actually update the struct.
-	for i := range plans {
-		plans[i].ID = uuid.New()
+	for i, plan := range plans {
+		//plans[i].ID = uuid.New()
+		FQPlanName := fmt.Sprintf("%s-%s", FQSpecName, plan.Name)
+		hasher := md5.New()
+		hasher.Write([]byte(FQPlanName))
+		plans[i].ID = hex.EncodeToString(hasher.Sum(nil))
 	}
 }
 
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -255,6 +255,7 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 			registryErrors = append(registryErrors, err)
 		}
 		imageCount += count
+		// this will also update the plan id
 		addNameAndIDForSpec(s, r.RegistryName())
 		specs = append(specs, s...)
 	}
@@ -266,12 +267,29 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 		return nil, errors.New("all registries failed on bootstrap")
 	}
 	specManifest := map[string]*apb.Spec{}
+	planNameManifest := map[string]string{}
+
 	for _, s := range specs {
 		specManifest[s.ID] = s
+
+		// each of the plans from all of the specs gets its own uuid. even
+		// though the names may be the same we want them to be globally unique.
+		for _, p := range s.Plans {
+			if p.ID == "" {
+				a.log.Errorf("We have a plan that did not get its id generated: %v", p.Name)
+			}
+			planNameManifest[p.ID] = p.Name
+		}
 	}
 	if err := a.dao.BatchSetSpecs(specManifest); err != nil {
 		return nil, err
 	}
+
+	// save off the plan names as well
+	if err = a.dao.BatchSetPlanNames(planNameManifest); err != nil {
+		return nil, err
+	}
+
 	apb.AddSecrets(specs)
 
 	return &BootstrapResponse{SpecCount: len(specs), ImageCount: imageCount}, nil
@@ -297,6 +315,19 @@ func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
 		hasher := md5.New()
 		hasher.Write([]byte(spec.FQName))
 		spec.ID = hex.EncodeToString(hasher.Sum(nil))
+
+		// update the id on the plans, doing it here avoids looping through the
+		// specs array again
+		addIDForPlan(spec.Plans)
+	}
+}
+
+// addIDForPlan - for each of the plans create a new ID
+func addIDForPlan(plans []apb.Plan) {
+
+	// need to use the index into the array to actually update the struct.
+	for i := range plans {
+		plans[i].ID = uuid.New()
 	}
 }
 
@@ -514,6 +545,7 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 	*/
 	var spec *apb.Spec
 	var err error
+	var planName string
 
 	// Retrieve requested spec
 	specID := req.ServiceID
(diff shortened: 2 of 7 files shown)


jmrodri:
```hs
$ oc get serviceplans
NAME                                   KIND
a78ed2af-ad55-11e7-9447-c85b76145add   ServicePlan.v1alpha1.servicecatalog.k8s.io
a79ac311-ad55-11e7-9447-c85b76145add   ServicePlan.v1alpha1.servicecatalog.k8s.io
a7a77477-ad55-11e7-9447-c85b76145add   ServicePlan.v1alpha1.servicecatalog.k8s.io
a7abbd8d-ad55-11e7-9447-c85b76145add   ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-etherpad-apb-default                ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-hastebin-apb-default                ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-hello-world-apb-default             ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-hello-world-db-apb-default          ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-jenkins-apb-default                 ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-mediawiki123-apb-default            ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-pyzip-demo-apb-default              ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-pyzip-demo-db-apb-default           ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-rhscl-mariadb-apb-dev               ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-rhscl-mariadb-apb-prod              ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-rhscl-postgresql-apb-dev            ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-rhscl-postgresql-apb-prod           ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-rocketchat-apb-default              ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-thelounge-apb-default               ServicePlan.v1alpha1.servicecatalog.k8s.io
dh-wordpress-ha-apb-default            ServicePlan.v1alpha1.servicecatalog.k8s.io
```

jmrodri:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1500048

eriknelson:
Do we need to be concerned about the length of the ids?

pkg/broker/util.go
```diff
@@ -46,20 +46,21 @@ func SpecToService(spec *apb.Spec) Service {
...
 	copy(retSvc.Tags, spec.Tags)
 	return retSvc
 }
 
-func toBrokerPlans(apbPlans []apb.Plan) []Plan {
+func toBrokerPlans(fqname string, apbPlans []apb.Plan) []Plan {
 	brokerPlans := make([]Plan, len(apbPlans))
 	i := 0
 	for _, plan := range apbPlans {
+		genID := fmt.Sprintf("%v-%v", fqname, plan.Name)
```
cfchase: Is there a character limit on these IDs and how does it compare to the char limit on the unique plan IDs making them not unique.  e.g. 
```yaml
serviceclass: dh-cfchase-reallyreallylongfqnameblahblah
  plans:
    - plan1
    - plan2
``` 
turning into 
```
dh-cfchase-reallyreallylongfqnameblahblah-plan1
dh-cfchase-reallyreallylongfqnameblahblah-plan2
```

and then getting cut short if really long to:
```
dh-cfchase-reallyreallylongfqnameblahblah-pla
dh-cfchase-reallyreallylongfqnameblahblah-pla
```

and then not being unique

cfchase: how about 
serviceclass name = 32 characters of service plan + 8 character generated (40 chars max)
plan name = 16 characters of english + 8 characters generated (24 chars max)
total of 64 characters max, pretty darn unique

djzager: @cfchase as far as serviceclass name is concerned I think we probably want to limit the number of places that have this logic of truncating names. Currently, in the broker:

```
                re := regexp.MustCompile(fqNameRegex)
                spec.FQName = re.ReplaceAllLiteralString(
                        fmt.Sprintf("%v-%v", registryName, spec.FQName),
                        "-")
                spec.FQName = fmt.Sprintf("%.51v", spec.FQName)
                if strings.HasSuffix(spec.FQName, "-") {
                        spec.FQName = spec.FQName[:len(spec.FQName)-1]
                }
```

`registryName` is typically something like `dh` (I've used `dz` when I've added my org as an additional registry) so we would have something like `dh-etherpad-apb`. We are truncating this to 51 characters wide but, I believe, this is sufficiently unique. My only concern is, **if** we want to shorten this, we should do it here in the broker **before** we create plan names.

#### Not really related to this PR

Follow up questions regarding serviceclass names, apb names, fqnames, etc:
1. Are we doing anything to prevent someone from using the same name 2x in the broker config?
1. How do we handle multiple APB containers whose apb.yml have the same name?

jmrodri: @eriknelson @djzager @dymurray @cfchase so does the ID need to be human readable? can't I just do a UUID? like f4b72d2a-bf85-4847-a19f-edd8beef0ddd

dymurray: @jmrodri I don't think there is any requirement that they are human readable. I think a UUID would work fine and in fact that is what the template broker is doing.

pkg/broker/broker.go
```diff
@@ -959,3 +965,11 @@ func ocLogin(log *logging.Logger, args ...string) error {
 	}
 	return nil
 }
+
+func extractNameFromPlanID(planID string) (string, error) {
+	splitPlan := strings.Split(planID, "-")
```
rthallisey: Can you get a planID without a '-' in it?

pkg/broker/broker.go
```diff
@@ -959,3 +965,11 @@ func ocLogin(log *logging.Logger, args ...string) error {
...
+
+	// TOOD: Error Handling
```
rthallisey: Can planID be just '-'? If neither of these cases can occur or are error checked elsewhere then you don't need any error checking. If they aren't, then they'll need to be checked because they will cause runtime errors.

pkg/broker/broker.go
jmrodri: @rthallisey not the way I generated them. There would always be at least one '-' in it.

pkg/broker/util.go
djzager: +1 to @dymurray comment

jmrodri:
I updated this PR with the new uuid and etcd changes. I also updated the output. The 145add are the template-service-brokers, the rest are ours.

jmrodri:
@eriknelson @shawn-hurley @dymurray @rthallisey Please REREVIEW!

pkg/apb/types.go
```diff
@@ -50,6 +50,7 @@ type ParameterDescriptor struct {
 
 // Plan - Plan object describing an APB deployment plan and associated parameters
 type Plan struct {
+	ID             string                 `json:"id" yaml:"-"`
```
eriknelson: Should these values be the same?

pkg/broker/broker.go
```diff
@@ -266,12 +267,29 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	planNameManifest := map[string]string{}
+
 	for _, s := range specs {
 		specManifest[s.ID] = s
+
+		// each of the plans from all of the specs gets its own uuid. even
+		// though the names may be the same we want them to be globally unique.
+		for _, p := range s.Plans {
+			if p.ID == "" {
+				a.log.Errorf("We have a plan that did not get its id generated: %v", p.Name)
```
eriknelson: Should this get continued?

pkg/broker/broker.go
```diff
@@ -539,10 +571,20 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
 		return nil, errors.New(errMsg)
 	}
 
+	planName, err = a.dao.GetPlanName(req.PlanID)
+	if err != nil {
+		// etcd return not found i.e. code 100
+		if client.IsKeyNotFound(err) {
```
eriknelson: +1

pkg/broker/broker.go
```diff
@@ -297,6 +315,19 @@ func addNameAndIDForSpec(specs []*apb.Spec, registryName string) {
...
+		addIDForPlan(spec.Plans)
+	}
+}
+
+// addIDForPlan - for each of the plans create a new ID
+func addIDForPlan(plans []apb.Plan) {
+
+	// need to use the index into the array to actually update the struct.
+	for i := range plans {
+		plans[i].ID = uuid.New()
```
shawn-hurley: What will happen here on restarting of the broker? Won't we blow away the spec and therefore get two different entries for the service plans much like what was happening with service instances names awhile ago and the reason we went to a hash of the FQname?

eriknelson: Yes, that's an issue. Good catch. It seems we need the lookup ID to be a similar hash.

pkg/apb/types.go
shawn-hurley: I think that we don't want a value to be set from the Spec yaml for the ID field.

pkg/broker/broker.go
jmrodri: UGH!

pkg/apb/types.go
jmrodri: 👍 to @shawn-hurley we don't want it to get set form the spec yaml. But I want it output via JSON.

pkg/broker/broker.go
jmrodri: added a continue earlier

PR #491, opened 2017-10-13
Gate Debugging

rthallisey:


rthallisey:
Abandoning this. This is a bug will be fixed in upstream k8s

PR #497, opened 2017-10-16
Adding prometheus metrics for ASB

shawn-hurley:
Adds Prometheus metrics for the ASB 
 - Metrics package to manage the Prometheus metrics 
 - Adding new endpoint from `apiserver` `/metrics`

depends-on <PR>
N/A
N/A

### DIFF
--- a/pkg/apb/svc_acct.go
+++ b/pkg/apb/svc_acct.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/openshift/ansible-service-broker/pkg/clients"
+	"github.com/openshift/ansible-service-broker/pkg/metrics"
 	"github.com/openshift/ansible-service-broker/pkg/runtime"
 	apicorev1 "k8s.io/kubernetes/pkg/api/v1"
 
@@ -55,6 +56,7 @@ func (s *ServiceAccountManager) CreateApbSandbox(
 	executionContext ExecutionContext,
 	apbRole string,
 ) (string, error) {
+	metrics.SandboxCreated()
 	apbID := executionContext.PodName
 	svcAccountName := executionContext.PodName
 	roleBindingName := executionContext.PodName
@@ -288,6 +290,7 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
 	// "If there is an error, it will be of type *PathError"
 	// We don't care, because it's gone
 	os.Remove(filePathFromHandle(executionContext.PodName))
+	metrics.SandboxDeleted()
 
 	return
 }
--- a/pkg/apb/svc_acct.go
+++ b/pkg/apb/svc_acct.go
@@ -123,7 +123,6 @@ func (s *ServiceAccountManager) CreateApbSandbox(
 
 	s.log.Info("Successfully created apb sandbox: [ %s ], with %s permissions in namespace %s", apbID, apbRole, executionContext.Namespace)
 
-	metrics.SandboxCreated()
 	return apbID, nil
 }
 
@@ -246,6 +245,8 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
 		if clusterConfig.Namespace != executionContext.Namespace {
 			s.log.Debug("Deleting namespace %s", executionContext.Namespace)
 			k8scli.CoreV1().Namespaces().Delete(executionContext.Namespace, &metav1.DeleteOptions{})
+			// This is keeping track of namespaces.
+			metrics.SandboxDeleted()
 		} else {
 			// We should not be attempting to run pods in the ASB namespace, if we are, something is seriously wrong.
 			panic(fmt.Errorf("Broker is attempting to delete its own namespace"))
@@ -290,7 +291,6 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
 	// "If there is an error, it will be of type *PathError"
 	// We don't care, because it's gone
 	os.Remove(filePathFromHandle(executionContext.PodName))
-	metrics.SandboxDeleted()
 
 	return
 }
--- a/pkg/apb/svc_acct.go
+++ b/pkg/apb/svc_acct.go
@@ -56,7 +56,6 @@ func (s *ServiceAccountManager) CreateApbSandbox(
 	executionContext ExecutionContext,
 	apbRole string,
 ) (string, error) {
-	metrics.SandboxCreated()
 	apbID := executionContext.PodName
 	svcAccountName := executionContext.PodName
 	roleBindingName := executionContext.PodName
@@ -124,6 +123,7 @@ func (s *ServiceAccountManager) CreateApbSandbox(
 
 	s.log.Info("Successfully created apb sandbox: [ %s ], with %s permissions in namespace %s", apbID, apbRole, executionContext.Namespace)
 
+	metrics.SandboxCreated()
 	return apbID, nil
 }
 
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -50,6 +50,7 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/openshift/ansible-service-broker/pkg/dao"
 	"github.com/openshift/ansible-service-broker/pkg/handler"
+	"github.com/openshift/ansible-service-broker/pkg/metrics"
 	"github.com/openshift/ansible-service-broker/pkg/registries"
 	"github.com/openshift/ansible-service-broker/pkg/version"
 	"github.com/prometheus/client_golang/prometheus"
@@ -231,6 +232,8 @@ func CreateApp() App {
 	app.log.Debugf("Active work engine topics: %+v", app.engine.GetActiveTopics())
 
 	apb.InitializeSecretsCache(app.config.Secrets, app.log.Logger)
+	// Initialize Metrics.
+	metrics.Init(app.log.Logger)
 	app.log.Debug("Creating AnsibleBroker")
 	if app.broker, err = broker.NewAnsibleBroker(
 		app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.Broker,
@@ -334,12 +337,15 @@ func (a *App) Start() {
 		clusterURL = defaultClusterURLPreFix
 	}
 
-	daHandler := handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules)
+	daHandler := prometheus.InstrumentHandler(
+		"ansible-service-broker",
+		handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules),
+	)
 
 	if clusterURL == "/" {
-		genericserver.Handler.NonGoRestfulMux.HandlePrefix("/", prometheus.InstrumentHandler("ansible-service-broker", daHandler))
+		genericserver.Handler.NonGoRestfulMux.HandlePrefix("/", daHandler)
 	} else {
-		genericserver.Handler.NonGoRestfulMux.HandlePrefix(fmt.Sprintf("%v/", clusterURL), prometheus.InstrumentHandler("ansible-service-broker", daHandler))
+		genericserver.Handler.NonGoRestfulMux.HandlePrefix(fmt.Sprintf("%v/", clusterURL), daHandler)
 	}
 
 	defaultMetrics := routes.DefaultMetrics{}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/apiserver/pkg/authentication/authenticatorfactory"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	genericoptions "k8s.io/apiserver/pkg/server/options"
+	"k8s.io/apiserver/pkg/server/routes"
 	authenticationclient "k8s.io/client-go/kubernetes/typed/authentication/v1beta1"
 	"k8s.io/kubernetes/pkg/apis/rbac"
 	v1beta1rbac "k8s.io/kubernetes/pkg/apis/rbac/v1beta1"
@@ -51,6 +52,7 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/handler"
 	"github.com/openshift/ansible-service-broker/pkg/registries"
 	"github.com/openshift/ansible-service-broker/pkg/version"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 var (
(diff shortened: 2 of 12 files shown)


pkg/app/app.go
```diff
@@ -335,11 +337,14 @@ func (a *App) Start() {
 	daHandler := handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules)
 
 	if clusterURL == "/" {
-		genericserver.Handler.NonGoRestfulMux.HandlePrefix("/", daHandler)
+		genericserver.Handler.NonGoRestfulMux.HandlePrefix("/", prometheus.InstrumentHandler("ansible-service-broker", daHandler))
```
djzager: This may make it clear my ignorance. Could you not update `daHandler` above to something like:
```
daHandler := prometheus.InstrumentHandler(
    "ansible-service-broker",
    handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules),
)
```

djzager:
Does this need a bugzilla since I'm not sure if it is being included in 3.7.

pkg/app/app.go
shawn-hurley: I bet you could, and that would probably be much more clear :) let me double check :+1:

shawn-hurley:
[(8) [CM-OPS-Tools] Ansible Service Broker Prometheus endpoint coverage](https://trello.com/c/Cruuo5Vl/397-8-cm-ops-tools-ansible-service-broker-prometheus-endpoint-coverage)

jwmatthews:
This PR does not need a bugzilla, it is one of two trello cards which have approval for being worked on post  the general feature complete date 3.7.  Other card is the service instance work.

Both features must be completed prior to this Friday.

jwmatthews:
@jcantrill @smarterclayton  @liggitt 
CC: @pweil- 

do you have any input on this PR.   It's our first cut at adding prometheus metrics for the Ansible Broker as per card:  https://trello.com/c/Cruuo5Vl

This is a sample of the output returned:
  https://hastebin.com/onaqimepux.coffeescript

Endpoint is accessed by:

curl -k -H "Authorization: bearer `oc whoami -t`" https://asb-1338-ansible-service-broker.172.17.0.1.nip.io/metrics 

Assuming that the logged in user has 'cluster-debugger-role' for accessing this endpoint.


Notes here on the metrics we are gathering:
   https://docs.google.com/document/d/1ui57sb3kMf2HEt6LelfuDEGjrAhbSaD4NhYPljecyRE/edit

pkg/apb/svc_acct.go
```diff
@@ -54,6 +55,7 @@ func (s *ServiceAccountManager) CreateApbSandbox(
 	executionContext ExecutionContext,
 	apbRole string,
 ) (string, error) {
+	metrics.SandboxCreated()
```
jmrodri: Do we actually mark created before it is really created? What happens if this method returns with an error?

pkg/apb/svc_acct.go
```diff
@@ -283,6 +285,7 @@ func (s *ServiceAccountManager) DestroyApbSandbox(executionContext ExecutionCont
 	// "If there is an error, it will be of type *PathError"
 	// We don't care, because it's gone
 	os.Remove(filePathFromHandle(executionContext.PodName))
+	metrics.SandboxDeleted()
```
jmrodri: This makes sense to do it once things have actually been deleted.

pkg/app/app.go
```diff
@@ -332,14 +337,20 @@ func (a *App) Start() {
...
+	)
 
 	if clusterURL == "/" {
 		genericserver.Handler.NonGoRestfulMux.HandlePrefix("/", daHandler)
 	} else {
 		genericserver.Handler.NonGoRestfulMux.HandlePrefix(fmt.Sprintf("%v/", clusterURL), daHandler)
 	}
 
+	defaultMetrics := routes.DefaultMetrics{}
+	defaultMetrics.Install(genericserver.Handler.NonGoRestfulMux)
```
jmrodri: What is this doing?

pkg/app/app.go
```diff
@@ -332,14 +337,20 @@ func (a *App) Start() {
 		clusterURL = defaultClusterURLPreFix
 	}
 
-	daHandler := handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules)
+	daHandler := prometheus.InstrumentHandler(
+		"ansible-service-broker",
+		handler.NewHandler(a.broker, a.log.Logger, a.config.Broker, clusterURL, providers, rules),
+	)
```
jmrodri: This makes sense, a "middleware" :) handler around our handler.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Init - Initialize the metrics package.
+func Init(logger *logging.Logger) {
+	log = logger
+}
+
+func recoverMetricPanic() {
+	if r := recover(); r != nil {
+		log.Errorf("Recovering from metric function - %v", r)
+	}
```
jmrodri: What does `recover` do? What are we recovering?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// SandboxCreated - Counter for how many sandbox created.
+func SandboxCreated() {
+	defer recoverMetricPanic()
+	sandboxCreated.Inc()
```
jmrodri: I'm going to assume that `Inc` can handle multiple things incrementing it at the same time and that we won't have a miscount due to a race condition. 🤞

pkg/registries/registry.go
```diff
@@ -122,6 +123,7 @@ func (r Registry) LoadSpecs() ([]*apb.Spec, int, error) {
 	} else {
 		r.log.Notice("All specs passed validation!")
 	}
+	metrics.SpecsLoaded(r.RegistryName(), len(validatedSpecs))
```
jmrodri: 👍 specs loaded looks good.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+const (
+	subsystem = "asb"
+)
+
+var (
+	sandboxCreated = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "sandbox_created",
+			Help:      "Counter of all sandbox namespaces that are created.",
```
jmrodri: I can't really comment on the validity of this metrics file :) I'm going to have to trust you that this is the proper way of doing it :)

pkg/broker/provision_subscriber.go
```diff
@@ -50,6 +51,7 @@ func (p *ProvisionWorkSubscriber) Subscribe(msgBuffer <-chan WorkMsg) {
 			msg := <-msgBuffer
 			var pmsg *ProvisionMsg
 			var extCreds *apb.ExtractedCredentials
+			metrics.RemoveProvisionJob()
```
jmrodri: Not entirely sure this is a valid state. We're not really removing the provision job, I guess remove is just the reference count so I guess that's ok.

pkg/broker/provision_job.go
```diff
@@ -62,6 +63,7 @@ func NewProvisionJob(serviceInstance *apb.ServiceInstance, clusterConfig apb.Clu
 
 // Run - run the provision job.
 func (p *ProvisionJob) Run(token string, msgBuffer chan<- WorkMsg) {
+	metrics.AddProvisionJob()
```
jmrodri: +1

pkg/broker/deprovision_subscriber.go
```diff
@@ -43,12 +44,13 @@ func NewDeprovisionWorkSubscriber(dao *dao.Dao, log *logging.Logger) *Deprovisio
...
 func (d *DeprovisionWorkSubscriber) Subscribe(msgBuffer <-chan WorkMsg) {
 	d.msgBuffer = msgBuffer
-	var dmsg *DeprovisionMsg
 
 	go func() {
 		d.log.Info("Listening for deprovision messages")
 		for {
 			msg := <-msgBuffer
+			var dmsg *DeprovisionMsg
+			metrics.RemoveDeprovisionJob()
```
jmrodri: Same comment as RemoveProvisionJob.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// SpecsReset - Counter for how many times the specs are reloaded.
+func SpecsReset() {
+	defer recoverMetricPanic()
+	specsReset.Inc()
+}
+
+// AddProvisionJob - Add a provision job to the counter.
+func AddProvisionJob() {
```
jmrodri: What about ProvisionJobStarted?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	defer recoverMetricPanic()
+	provisionJob.Inc()
+}
+
+// AddDeprovisionJob - Add a deprovision job to the counter.
+func AddDeprovisionJob() {
```
jmrodri: DeprovisionJobStarted

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	defer recoverMetricPanic()
+	deprovisionJob.Inc()
+}
+
+// RemoveProvisionJob - Remove a provision job to the counter.
+func RemoveProvisionJob() {
```
jmrodri: `ProvisionJobFinished`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	defer recoverMetricPanic()
+	provisionJob.Dec()
+}
+
+// RemoveDeprovisionJob - Remove a deprovision job to the counter.
+func RemoveDeprovisionJob() {
```
jmrodri: `DeprovisionJobFinished` or something like `PopDeprovisionJob` (of course make it consistent with provision) :)

pkg/app/app.go
jmrodri: I guess we couldn't just add a metrics endpoint like we did with /v2 and the dev endpoints?

pkg/apb/svc_acct.go
shawn-hurley: nothing bad would happen, our metrics would say we created a sandbox when it actually failed to create the sandbox, I think that moving it down to once it is actually created makes sense to me.

pkg/app/app.go
shawn-hurley: https://github.com/kubernetes/apiserver/blob/master/pkg/server/routes/metrics.go#L31

We totally could, we get metrics that the apiserver sets up by default. Some have to do with apiserver audit events, those I would not want to set up on my own. But the others, (golang process metrics) we could set up on our own. Figured we already have apiserver in might as well use the stuff it provides for "free".

pkg/metrics/metrics.go
shawn-hurley: > All exported functions and methods are safe to be used concurrently unless specified otherwise.

https://godoc.org/github.com/prometheus/client_golang/prometheus


:crossed_fingers: sounds like it should be fine.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+		}, []string{"registry_name"})
+
+	specsReset = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "specs_reset",
+			Help:      "Counter of how many times the specs have been reset.",
+		})
+
+	provisionJob = prometheus.NewGauge(
```
rthallisey: I think another good metric is a counter of the number of things provisioned and deprovisioned. Ditto bind and unbind.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	logging "github.com/op/go-logging"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	subsystem = "asb"
+)
+
+var (
+	sandboxCreated = prometheus.NewCounter(
```
rthallisey: Why do we want to count all the namespaces created? To me this metric means 'the total number of actions the broker has performed'.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+			Subsystem: subsystem,
+			Name:      "deprovision_jobs",
+			Help:      "How many deprovision jobs are actively in the buffer.",
+		})
+
+	log = logging.MustGetLogger("metrics")
+)
+
+func init() {
+	prometheus.MustRegister(sandboxCreated)
```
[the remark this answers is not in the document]
shawn-hurley: I think the pod should probably blow up if we can't expose metrics, this seems like the right thing to do if people are going to use the metrics to determine if things are wrong. 

This is different than a metric not working (guarding with the recover function), this is all of the Prometheus client is not working. I think this is why I would want to error the broker loudly and early,  thoughts?

pkg/metrics/metrics.go
shawn-hurley: Can you add to https://docs.google.com/document/d/1ui57sb3kMf2HEt6LelfuDEGjrAhbSaD4NhYPljecyRE/edit# 

I think that we can add better metrics for 3.8, but need to get some basics up today. 

Do you feel super strongly that we should have them for the initial PR?

pkg/metrics/metrics.go
shawn-hurley: I think sandbox is specific, this will give us some nice indirection metrics, say if we have a lot of sandbox creations and no sandbox deletions then things are going very wrong somewhere. This is not the actions, bind and unbind are configured by default to not launch APB's and therefore will not be counted in this metric.

pkg/metrics/metrics.go
eriknelson: Makes sense to me to get a baseline and then iterate once we have some feedback on what is useful in the wild.

pkg/metrics/metrics.go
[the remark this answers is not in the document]
eriknelson: I'm always a proponent of loud and early failure for systems deemed important enough. Curious, is it possible to configure metrics as enabled/disabled?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+// Therefore, we will recover our panics here and error log them
+// for later diagnosis but will never fail the app.
+func recoverMetricPanic() {
+	if r := recover(); r != nil {
+		log.Errorf("Recovering from metric function - %v", r)
+	}
+}
+
+// SandboxCreated - Counter for how many sandbox created.
+func SandboxCreated() {
```
eriknelson: +1 for these hooks.

pkg/metrics/metrics.go
rthallisey: > I think sandbox is specific, this will give us some nice indirection metrics, say if we have a lot of
> sandbox creations and no sandbox deletions then things are going very wrong somewhere.
>
It's a good metric to be aware that the sandboxes are being cleaned up, but we should do that with a gauge instead of two counters.

> This is not the actions, bind and unbind are configured by default to not launch APB's and
> therefore will not be counted in this metric.
>
We can still count them if you increment in the if block. 
https://github.com/openshift/ansible-service-broker/blob/81381b1600c489a193fd13afb0a4862ed7b878fa/pkg/broker/broker.go#L865

pkg/metrics/metrics.go
rthallisey: IMO keeping track of number of bind/unbind is important. It's one the four APIs we track and care about. We can add better metrics for 3.8, but to me this seems like a core metric we want.

pkg/metrics/metrics.go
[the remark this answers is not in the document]
rthallisey: We still are going to error loudly and early. It's just like how we check if we can connect to the etcd endpoint.  But, we will have control over the error message so we can have a smoother exit.

shawn-hurley:
@rthallisey Can you double check that the agreed upon changes are what you were thinking?

PR #498, opened 2017-10-16
Bug 1499622 - Return 202 if provisioning job is in progress

dymurray:
This PR fixes the scenario when the catalog sends two provisioning requests and we return a 200 without checking for jobs in progress.

 - Check for provisioning jobs and return 202 if jobs are running.

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -626,7 +626,7 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 		// away from []byte it can still be evaluated.
 		if uuid.Equal(si.ID, serviceInstance.ID) {
 			if reflect.DeepEqual(si.Parameters, serviceInstance.Parameters) {
-				alreadyInProgress, jobToken, err := a.isProvisionInProgress(serviceInstance)
+				alreadyInProgress, jobToken, err := a.isJobInProgress(serviceInstance, apb.JobMethodProvision)
 				if err != nil {
 					return nil, fmt.Errorf("An error occurred while trying to determine if a provision job is already in progress for instance: %s", serviceInstance.ID)
 				}
@@ -716,7 +716,7 @@ func (a AnsibleBroker) Deprovision(
 		return nil, err
 	}
 
-	alreadyInProgress, jobToken, err := a.isDeprovisionInProgress(&instance)
+	alreadyInProgress, jobToken, err := a.isJobInProgress(&instance, apb.JobMethodDeprovision)
 	if err != nil {
 		return nil, fmt.Errorf("An error occurred while trying to determine if a deprovision job is already in progress for instance: %s", instance.ID)
 	}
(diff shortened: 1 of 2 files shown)


pkg/handler/handler.go
```diff
@@ -268,6 +268,8 @@ func (h handler) provision(w http.ResponseWriter, r *http.Request, params map[st
 		switch err {
 		case broker.ErrorDuplicate:
 			writeResponse(w, http.StatusConflict, broker.ProvisionResponse{})
+		case broker.ErrorProvisionInProgress:
+			writeResponse(w, http.StatusAccepted, broker.ProvisionResponse{})
```
jmrodri: I would change `broker.ProvisionResponse{}` to `resp`.

pkg/broker/broker.go
```diff
@@ -619,6 +621,16 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
...
 	// if err is not nil, we will just bubble that up
 
+	alreadyInProgress, err := a.isProvisionInProgress(serviceInstance)
+	if err != nil {
+		return nil, fmt.Errorf("An error occurred while trying to determine if a provision job is already in progress for instance: %s", serviceInstance.ID)
+	}
+
+	if alreadyInProgress {
+		a.log.Infof("Provision requested for instance %s, but job is already in progress", serviceInstance.ID)
+		return nil, ErrorProvisionInProgress
```
jmrodri: I would change `nil` to `&ProvisionResponse{Operation: token}` somehow we'll need the token not sure yet.

pkg/broker/broker.go
```diff
@@ -619,6 +621,16 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
 	//
 	// if err is not nil, we will just bubble that up
 
+	alreadyInProgress, err := a.isProvisionInProgress(serviceInstance)
```
shawn-hurley: I think this call needs to be within the `if uuid.Equal....` block. 

Reason being if it is a duplicate we want to return duplicate and not give the job token to the other service instance

jmrodri:
@eriknelson thoughts on my comments about returning a ProvisionResponse with Operation when we are still in progress?

pkg/handler/handler.go
shawn-hurley: I think that we need to return the job token, like: `ProvisionResponse{Operation: token}`

pkg/broker/broker.go
shawn-hurley: :+1: deleted my comment that was essentially the same as this.

eriknelson:
Will review this first thing tomorrow morning.

dymurray:
Please look at the actual branch... for some reason the PR is showing out of date diffs.

dymurray:
@jmrodri @shawn-hurley Updated with your feedback.

eriknelson:
@jmrodri Does the spec call for the operation token on an InProgress response?

dymurray:
@eriknelson I believe it does here: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#body-3

pkg/broker/broker.go
```diff
@@ -706,14 +716,14 @@ func (a AnsibleBroker) Deprovision(
...
-	alreadyInProgress, err := a.isDeprovisionInProgress(&instance)
+	alreadyInProgress, jobToken, err := a.isDeprovisionInProgress(&instance)
 	if err != nil {
 		return nil, fmt.Errorf("An error occurred while trying to determine if a deprovision job is already in progress for instance: %s", instance.ID)
 	}
 
 	if alreadyInProgress {
 		a.log.Infof("Deprovision requested for instance %s, but job is already in progress", instance.ID)
-		return nil, ErrorDeprovisionInProgress
+		return &DeprovisionResponse{Operation: jobToken}, ErrorDeprovisionInProgress
```
shawn-hurley: :+1:

eriknelson:
@jmrodri good catch re: the Op!

pkg/broker/broker.go
```diff
@@ -764,14 +774,32 @@ func (a AnsibleBroker) validateDeprovision(instance *apb.ServiceInstance) error
...
 
+	var token string
+	proJobs := dao.MapJobStatesWithMethod(allJobs, apb.JobMethodProvision)
+	if len(proJobs) > 0 {
+		token = proJobs[0].Token
+	}
+	return len(proJobs) > 0, token, nil
+}
+
+func (a AnsibleBroker) isDeprovisionInProgress(instance *apb.ServiceInstance) (bool, string, error) {
```
djzager: This may be more of a nitpick but I am not a big fan of 2 functions that are so identical to each other. I believe there are 2 ways that you could improve this:

PR #500, opened 2017-10-17
Bug 1503233 - Add liveness and readiness checks to ASB dc

dymurray:
This PR adds liveness and readiness checks to the ASB template on the ASB pod.

 - Adds liveness and readiness check on /healthz


shawn-hurley:
ACK

PR #502, opened 2017-10-18
Bug 1503289 - Move registry credentials to a secret

dymurray:
Move registry credentials to a secret and consumed from a different file

This PR moves the registry credentials into a secret so that it is not visible in plaintext from the configmap. 

 - Move registry credentials into a new file at /etc/ansible-service-broker/registry-auth.yaml
 - Create a secret out of base64 encoded credentials and inject them as environment variables to be consumed by the broker
 - Broker creates /etc/ansible-service-broker/registry-auth.yaml in entrypoint script
 - Updates all prep scripts to handle new behavior

IMPORTANT:
You now need your Dockerhub environment variables to be base64 or else the secret will fail to be created. Be sure to update catasb or the template vars.

### DIFF
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 USER_ID=$(id -u)
-REGISTRY_AUTH=/etc/ansible-service-broker/registry-auth.yaml
 if [ ${USER_UID} != ${USER_ID} ]; then
   sed "s@${USER_NAME}:x:\${USER_ID}:@${USER_NAME}:x:${USER_ID}:@g" ${BASE_DIR}/etc/passwd.template > /etc/passwd
 fi
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -16,11 +16,4 @@ if [ ! -f "$BROKER_CONFIG" ] ; then
 fi
 echo "Using config file mounted to $BROKER_CONFIG"
 
-cat << EOF  > "${REGISTRY_AUTH}"
----
-- type: dockerhub
-  user: ${DOCKERHUB_USER}
-  pass: ${DOCKERHUB_PASS}
-EOF
-
-asbd -c $BROKER_CONFIG --registryauth=$REGISTRY_AUTH $FLAGS
+asbd -c $BROKER_CONFIG $FLAGS
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-
 USER_ID=$(id -u)
+REGISTRY_AUTH=/etc/ansible-service-broker/registry-auth.yaml
 if [ ${USER_UID} != ${USER_ID} ]; then
   sed "s@${USER_NAME}:x:\${USER_ID}:@${USER_NAME}:x:${USER_ID}:@g" ${BASE_DIR}/etc/passwd.template > /etc/passwd
 fi
@@ -16,4 +16,11 @@ if [ ! -f "$BROKER_CONFIG" ] ; then
 fi
 echo "Using config file mounted to $BROKER_CONFIG"
 
-asbd -c $BROKER_CONFIG $FLAGS
+cat << EOF  > "${REGISTRY_AUTH}"
+---
+- type: dockerhub
+  user: ${DOCKERHUB_USER}
+  pass: ${DOCKERHUB_PASS}
+EOF
+
+asbd -c $BROKER_CONFIG --registryauth=$REGISTRY_AUTH $FLAGS
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 USER_ID=$(id -u)
 if [ ${USER_UID} != ${USER_ID} ]; then
   sed "s@${USER_NAME}:x:\${USER_ID}:@${USER_NAME}:x:${USER_ID}:@g" ${BASE_DIR}/etc/passwd.template > /etc/passwd
--- a/etc/example-registry-auth.yaml
+++ b/etc/example-registry-auth.yaml
@@ -1,4 +1,3 @@
 ---
-- type: dockerhub
-  user: DOCKERHUB_USER
-  pass: DOCKERHUB_PASS
+username: DOCKERHUB_USER
+password: DOCKERHUB_PASS
--- a/etc/example-registry-auth.yaml
+++ b/etc/example-registry-auth.yaml
@@ -1,0 +1,4 @@
+---
+- type: dockerhub
+  user: DOCKERHUB_USER
+  pass: DOCKERHUB_PASS
--- a/etc/generated_registry_auth.yaml
+++ b/etc/generated_registry_auth.yaml
@@ -1,0 +1,4 @@
+---
+- type: dockerhub
+  user: dymurray
+  pass: qci-test
--- a/etc/generated_registry_auth.yaml
+++ b/etc/generated_registry_auth.yaml
@@ -1,4 +1,0 @@
----
-- type: dockerhub
-  user: dymurray
-  pass: qci-test
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -153,7 +153,7 @@ func CreateApp() App {
 
 	// TODO: Let's take all these validations and delegate them to the client
 	// pkg.
-	if app.config, err = CreateConfig(app.args.ConfigFile, app.args.RegistryAuthFile); err != nil {
+	if app.config, err = CreateConfig(app.args.ConfigFile); err != nil {
 		os.Stderr.WriteString("ERROR: Failed to read config file\n")
 		os.Stderr.WriteString(err.Error() + "\n")
 		os.Exit(1)
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -153,7 +153,7 @@ func CreateApp() App {
 
 	// TODO: Let's take all these validations and delegate them to the client
 	// pkg.
-	if app.config, err = CreateConfig(app.args.ConfigFile); err != nil {
+	if app.config, err = CreateConfig(app.args.ConfigFile, app.args.RegistryAuthFile); err != nil {
 		os.Stderr.WriteString("ERROR: Failed to read config file\n")
 		os.Stderr.WriteString(err.Error() + "\n")
 		os.Exit(1)
--- a/pkg/app/config.go
+++ b/pkg/app/config.go
@@ -22,10 +22,8 @@ package app
 
 import (
 	"fmt"
-	logging "github.com/op/go-logging"
 	yaml "gopkg.in/yaml.v2"
 	"io/ioutil"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"strings"
 
@@ -86,7 +84,7 @@ func CreateConfig(configFile string) (Config, error) {
 
 	for regCount, reg := range config.Registry {
 		if reg.AuthType == "secret" {
-			data, err = getSecretData(reg.AuthName, config.Openshift.Namespace)
+			data, err = clients.GetSecretData(reg.AuthName, config.Openshift.Namespace)
 			if err != nil {
 				fmt.Println("Unable to get secret data")
 				// NEW ERROR
@@ -146,25 +144,4 @@ func validateConfig(c Config) error {
 
 	}
 	return nil
-}
-
-// Returns the data inside of a given secret
-func getSecretData(secretName, namespace string) (map[string][]byte, error) {
-	var log logging.Logger
-	k8scli, err := clients.Kubernetes(&log)
-	if err != nil {
-		return nil, err
-	}
-	var ret = make(map[string][]byte)
-
-	secretData, err := k8scli.CoreV1().Secrets(namespace).Get(secretName, meta_v1.GetOptions{})
-	if err != nil {
-		fmt.Printf("Unable to load secret '%s' from namespace '%s'", secretName, namespace)
-		return ret, nil
-	}
-	fmt.Printf("Found secret with name %v\n", secretName)
-
-	ret = secretData.Data
-
-	return ret, nil
 }
--- a/pkg/app/config.go
+++ b/pkg/app/config.go
@@ -21,6 +21,7 @@
 package app
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -35,17 +36,27 @@ import (
 
 // Config - The base config for the pieces of the applcation
 type Config struct {
-	Registry   []registries.Config
-	Dao        dao.Config
-	Log        LogConfig
-	Openshift  apb.ClusterConfig
-	ConfigFile string
-	Broker     broker.Config
-	Secrets    []apb.SecretsConfig
+	Registry         []registries.Config
+	Dao              dao.Config
+	Log              LogConfig
+	Openshift        apb.ClusterConfig
+	ConfigFile       string
+	RegistryAuthFile string
+	Broker           broker.Config
+	Secrets          []apb.SecretsConfig
+}
+
+type RegistryAuth struct {
+	Credentials []regCreds
+}
+type regCreds struct {
+	Type string
+	User string
+	Pass string
 }
 
 // CreateConfig - Read config file and create the Config struct
-func CreateConfig(configFile string) (Config, error) {
+func CreateConfig(configFile string, registryAuthFile string) (Config, error) {
 	var err error
 
 	// Confirm file is valid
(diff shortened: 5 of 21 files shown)


jmrodri:
This doesn't seem to allow me to specify multiple dockerhub credentials. With our current configuration I can set the broker to talk to dockerhub for `ansibleplaybookbundle` AND `jmrodri` orgs with two different logins.

dymurray:
@jmrodri initially I had the credentials in the auth file stored by name i.e.:
```
---
- name: dh
  user: <user>
  pass: <pass>
```
And then was storing the user/pass based on the name. There is a good reason why I didn't continue down that approach because working with secrets around independent registries just didn't seem like a scalable way to continue. 

Since we are working to remove the credentials completely I think it makes sense to consolidate the credentials based on registry type. This way you can declare rhcc creds (not needed currently), isv registry creds (will become important), without redeclaring for every different adapter.

dymurray:
@jmrodri Is it a valid use case to want to use two different credentials for two different adapters? It currently doesn't matter which credentials you use as they are just used to obtain a bearer token that allows us to search the registries index. One credential would work for any dockerhub org. The only scenario I could see that not being true would be private organizations but it would just require the most elevated credentials.

build/entrypoint.sh
```diff
@@ -16,4 +16,11 @@ if [ ! -f "$BROKER_CONFIG" ] ; then
...
 
-asbd -c $BROKER_CONFIG $FLAGS
+cat << EOF  > "${REGISTRY_AUTH}"
+---
+- type: dockerhub
+  user: ${DOCKERHUB_USER}
+  pass: ${DOCKERHUB_PASS}
+EOF
+
+asbd -c $BROKER_CONFIG --registryauth=$REGISTRY_AUTH $FLAGS
```
jmrodri: I prefer NOT to have this as a separate option. Why can't the admin specify the location of the secret in the registry config section?

Something like `auth: TYPE|NAME` where `TYPE` can be `secret` or `file`. Then `NAME` can be the name of the secret OR the fully qualified filename: `/etc/ansible-service-broker/dockerhub-auth`

```yaml
registry:
  - type: dockerhub
    name: docker
    url: https://registry.hub.docker.com
    auth: TYPE|NAME
    org: DOCKERHUB_ORG
    fail_on_error: false
  - type: rhcc
    name: rhcc
    url: registry.access.stage.redhat.com
    fail_on_error: true
    white_list:
      - "^legitimate.*-apb$"    
      # will allow all the APBs to be included. You must have at least 1 white 
      # list to retrieve APBs and this is the most permissive 
      - ".*-apb$" 
    black_list:
      - "malicious.*-apb$"
      - "^specific-blacklist-apb$"
dao:
```

OR we could use
```yaml
   auth_type: secret|file
   auth_name: NAME_OF_SECRET | FQFN
```

This gives us the flexibility to know how to read different ways. File makes it easier to run locally. But secret might make it easier in OpenShift by not needing a mount. Just go read the secret.

etc/example-registry-auth.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- type: dockerhub
+  user: DOCKERHUB_USER
+  pass: DOCKERHUB_PASS
```
jmrodri: This looks like it supports multiples by being an array but how do I know which registry this is for?

pkg/app/app.go
```diff
@@ -150,7 +150,7 @@ func CreateApp() App {
 
 	// TODO: Let's take all these validations and delegate them to the client
 	// pkg.
-	if app.config, err = CreateConfig(app.args.ConfigFile); err != nil {
+	if app.config, err = CreateConfig(app.args.ConfigFile, app.args.RegistryAuthFile); err != nil {
```
jmrodri: Like I mentioned above, I would prefer specifying the auth stuff as a pointer to some other location still in the main registry config.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -175,6 +185,15 @@ objects:
...
 - apiVersion: v1
+  kind: Secret
+  metadata:
+    name: registry-auth-secret
+    namespace: ansible-service-broker
+  data:
+    username: ${DOCKERHUB_USER}
+    password: ${DOCKERHUB_PASS}
+
+- apiVersion: v1
```
jmrodri: The secret is still ok, but instead of passing it as a parameter on the cli to the broker, it should be in the registry section of the config.

build/entrypoint.sh
jmrodri: OR we could reuse the auth format I did for basic auth and you might even be able to use the same structs (maybe not entire sure).

jmrodri:
We could support file in the broker as an initial cut, then just create that file from a secret. But it would be nice to read a secret directly.

shawn-hurley:
@dymurray @jmrodri I see a case where I have two private registries with two different "service accounts" that have access to one or the other.

build/entrypoint.sh
dymurray: +1, you're right no reason to make this a separate config. I agree with you. I think I pidgeonholed myself there because I was initially mounting the secret as a volume.

dymurray:
Thanks for the feedback, I'm pretty much completely refactoring this to fit Jesus suggestions.

etc/generated_registry_auth.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- type: dockerhub
+  user: dymurray
+  pass: qci-test
```
shawn-hurley: Probably want to ignore the generated file

dymurray: Oops, this shouldn't be there.

pkg/app/config.go
```diff
@@ -74,6 +84,40 @@ func CreateConfig(configFile string) (Config, error) {
...
+				return Config{}, err
+			}
+			if err = yaml.Unmarshal(dat, &regCreds); err != nil {
+				return Config{}, err
+			}
+			config.Registry[regCount].User = regCreds.Username
+			config.Registry[regCount].Pass = regCreds.Password
+
+		} else {
+			// ERROR
```
dymurray: Should we still allow the user to specify the user/pass in the config and use them here? This way we don't break old brokers and the real motivation was this for when we are running in openshift.

dymurray: @shawn-hurley that was how I previously implemented this before the refactoring :). That worked fine but required entrypoint changes. Jesus' brought up some good feedback that it would be cleaner to read the secret from the broker itself. We also control the template so its arbitrary which one we pick.

build/entrypoint.sh
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
```
jmrodri: Remove this change from this patch. It just adds noise and isn't necessary.

pkg/app/config.go
jmrodri: @dymurray If you are saying allow the user to use the old configuration values if they want, that's ok with me. If someone chooses to put a password in a config it is up to them.

pkg/app/config.go
```diff
@@ -103,3 +147,24 @@ func validateConfig(c Config) error {
 	}
 	return nil
 }
+
+// Returns the data inside of a given secret
+func getSecretData(secretName, namespace string) (map[string][]byte, error) {
```
jmrodri: I thought we already had a method to get secrets. I seem to recall a secrets.go that @fabianvf added in a previous patch. Seems that would be useful. If not this seems useful in general to me.

scripts/prep_local_devel_env.sh
```diff
@@ -58,7 +58,8 @@ fi
 
 
 # Determine the name of the secret which has the 'asb' service account info
-BROKER_SVC_ACCT_SECRET_NAME=`kubectl get serviceaccount asb -n ansible-service-broker -o jsonpath='{.secrets[0].name}'`
+BROKER_SVC_ACCT_SECRET_NAME=`kubectl get serviceaccount asb -n ansible-service-broker -o yaml | grep -Po asb-token-[a-z0-9]*`
```
jmrodri: why is this being changed in this PR?

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -187,8 +196,8 @@ objects:
         - type: "${REGISTRY_TYPE}"
           name: "${REGISTRY_NAME}"
           url: "${REGISTRY_URL}"
-          user: "${DOCKERHUB_USER}"
-          pass: "${DOCKERHUB_PASS}"
+          auth_type: "${REGISTRY_AUTH_TYPE}"
+          auth_name: "${REGISTRY_SECRET_NAME}"
```
jmrodri: +1 I like it.

pkg/app/config.go
```diff
@@ -22,13 +22,16 @@ package app
 
 import (
 	"fmt"
+	logging "github.com/op/go-logging"
+	yaml "gopkg.in/yaml.v2"
 	"io/ioutil"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jmrodri: I find it weird to have this imported into the config. Seems that the thing that uses this should be in a util package or part of the clients package directly. @shawn-hurley @eriknelson thoughts?

jmrodri:
make lint failed:
```
pkg/app/config.go:50:6: exported type RegCreds should have comment or be unexported
Found 1 lint suggestions; failing.
make: *** [Makefile:25: lint] Error 1
```

jmrodri:
I also saw you updated `config.go` but there are no change to `config_test.go`.

pkg/app/config.go
dymurray: Yeah we did have a method but it was being used to get secrets associated with a specific APB. It also required me to initialize a secrets cache in order to use those methods. I'm not against making this globally available and putting it in a utils package.

pkg/app/config.go
dymurray: This would be the getSecretData method so I'm fine with moving that out to a util package.

pkg/registries/adapters/dockerhub_adapter.go
```diff
@@ -241,13 +241,24 @@ func (r DockerHubAdapter) loadSpec(imageName string) (*apb.Spec, error) {
...
 func (r DockerHubAdapter) getBearerToken(imageName string) (string, error) {
-	req, err := http.NewRequest("GET",
-		fmt.Sprintf("https://auth.docker.io/token?grant_type=password&service=registry.docker.io&scope=repository:%v:pull", imageName),
-		nil)
-	if err != nil {
-		return "", err
+	var err error
+	var req *http.Request
+	if r.Config.User == "" {
+		r.Log.Debugf("hereeee")
```
shawn-hurley: Probably should remove this debug statement, my bad @dymurray.

dymurray:
@shawn-hurley The secret data *is* base64 encoded. Right now I'm expecting the user to set the user/pass creds already base64 encoded much like we already do when passing in broker user/pass.

dymurray:
https://github.com/fusor/catasb/pull/172

rthallisey:
[(5) Update Configuration for ASB Deployment to consume a Secret for Dockerhub credentials [asb]](https://trello.com/c/ALeto6O3/221-5-update-configuration-for-asb-deployment-to-consume-a-secret-for-dockerhub-credentials-asb)

pkg/app/config.go
```diff
@@ -74,6 +82,34 @@ func CreateConfig(configFile string) (Config, error) {
...
+			}
+
+			config.Registry[regCount].User = username
+			config.Registry[regCount].Pass = password
+
+		} else if reg.AuthType == "file" {
+			if dat, err = ioutil.ReadFile(reg.AuthName); err != nil {
+				return Config{}, err
+			}
+			if err = yaml.Unmarshal(dat, &regCred); err != nil {
```
rthallisey: The logging above here is really good. We could improve the logging here by adding a message about what failed. Could be something like 'missing or invalid username/password'.

pkg/registries/registry.go
```diff
@@ -40,14 +40,16 @@ var regex = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-
...
-	URL    string
-	User   string
-	Pass   string
-	Org    string
-	Tag    string
-	Type   string
-	Name   string
-	Images []string
+	URL      string
+	AuthType string `yaml:"auth_type"`
```
rthallisey: Can you put a comment about exactly what AuthType and AuthName are.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -455,6 +462,16 @@ parameters:
   name: BROKER_USER
   value: YWRtaW4=
 
+- description: Dockerhub user password
+  displayname: Dockerhub user password
+  name: DOCKERHUB_PASS
+  value: Y2hhbmdlbWUK
```
rthallisey: Is this a real docker user and pw?

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -189,6 +189,15 @@ objects:
     password: ${BROKER_PASS}
```
rthallisey: Do we need any changes in ```deploy-local-dev-changes.yaml``` for the local broker?

templates/deploy-ansible-service-broker.template.yaml
dymurray: This is "changeme/changeme" base64 encoded so I figured it was safe. I can make it an empty string too.

templates/deploy-ansible-service-broker.template.yaml
dymurray: I don't believe so because that template doesn't include the broker config.

templates/deploy-ansible-service-broker.template.yaml
shawn-hurley: @dymurray now that we have the ability to do empty username/password I don't see why we wouldn't do that?

templates/deploy-ansible-service-broker.template.yaml
dymurray: I don't believe we should go through the trouble of creating an empty secret for the user every time since it won't be used unless someone is going through the trouble of configuring a private registry. In that case they can configure it themself.

dymurray:
@rthallisey Thanks updated!

pkg/clients/kubernetes.go
```diff
@@ -49,6 +50,27 @@ func KubernetesConfig(log *logging.Logger) (*rest.Config, error) {
...
+	secretData, err := k8scli.CoreV1().Secrets(namespace).Get(secretName, meta_v1.GetOptions{})
+	if err != nil {
+		log.Error("Unable to load secret '%s' from namespace '%s'", secretName, namespace)
+		return ret, nil
+	}
+	log.Debug("Found secret with name %v\n", secretName)
+
+	ret = secretData.Data
+
+	return ret, nil
```
jmrodri: feels weird to assign a value to ret just to return it on the next line, I'd expected: `return secretData.Data` I'd get rid of ret entirely. Because you're creating an array that you ultimately throwaway on every successful call. I'd rather create an array only during the error condition.

```golang
// GetSecretData - Returns the data inside of a given secret
func GetSecretData(secretName, namespace string) (map[string][]byte, error) {
    var log logging.Logger
    k8scli, err := Kubernetes(&log)
    if err != nil {
        return nil, err
    }

    secretData, err := k8scli.CoreV1().Secrets(namespace).Get(secretName, meta_v1.GetOptions{})
    if err != nil {
        log.Errorf("Unable to load secret '%s' from namespace '%s'", secretName, namespace)
        return make(map[string][]byte), nil
    }
    log.Debugf("Found secret with name %v\n", secretName)

    return secretData.Data, nil
```

pkg/registries/adapters/dockerhub_adapter.go
```diff
@@ -241,13 +241,23 @@ func (r DockerHubAdapter) loadSpec(imageName string) (*apb.Spec, error) {
...
+			return "", err
+		}
+	} else {
+		req, err = http.NewRequest("GET",
+			fmt.Sprintf("https://auth.docker.io/token?grant_type=password&service=registry.docker.io&scope=repository:%v:pull", imageName),
+			nil)
+		if err != nil {
+			return "", err
+		}
+		req.SetBasicAuth(r.Config.User, r.Config.Pass)
```
jmrodri: these changes look okay 👍

pkg/registries/registry.go
```diff
@@ -60,6 +66,11 @@ func (c Config) Validate() bool {
 	if c.Name == "" {
 		return false
 	}
+	if c.AuthType == "file" || c.AuthType == "secret" {
```
jmrodri: do we need to validate the `AuthType` at all? i.e. if I enter "foobarauth" what will happen?

pkg/app/testdata/test-config-admin.yaml
```diff
@@ -3,12 +3,14 @@ registry:
   - type: dockerhub
     name: docker
     url: https://registry.hub.docker.com
+    auth_type: config
```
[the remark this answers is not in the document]
dymurray: Yes it is, to simply declare user/pass in the config. I have updated the test cases and switch statement to reflect that.

dymurray:
Valid options for registry config `auth_type` are:
* `secret`
* `file`
* `config`
* ``

If set to ``, then we are expecting that the registry does NOT require credentials.

If set to `secret`, or `file` then we expect `auth_name` to be set to either the secret name or the file location.

If set to `config` then we expect user/pass to be declared in the config.

Issue #475, opened 2017-10-05
Last Operation description


maleck13:
**This form is for bug reports and feature requests. Major features will go through a [spec process](https://github.com/openshift/ansible-service-broker/blob/master/CONTRIBUTING.md).**

> # Feature:

It would be nice for long running apbs, if a description could somehow be provided via apb about what the apb was doing to the user, both to inform and indicate progress. The service broker api supports a last_operation, which I see is in place, however the description appears to always be empty.
Wondering if there are any thoughts on how this might be achieved and if it is someting the asb wants to support

rthallisey:
https://github.com/openshift/ansible-service-broker/pull/505

maleck13:
@rthallisey is this actually solved by the PR? Or is it something not wanted? Looking at the PR the description is still empty.

rthallisey:
@maleck13 Ok I thought that PR covered it. Reopening.

maleck13:
On this particular feature, would a proposal be welcome?
An option I was considering:
- Writing the description of the last operation to a known location and have it collected by the broker periodically, I believe something similar happens for collecting the credentials for the secret. An example of this description would perhaps be:
```progress 60%: creating initial realm and clients```

rthallisey:
@maleck13 Proposals are always welcome!! 👍 

> Writing the description of the last operation to a known location and have it collected by the broker
> periodically, I believe something similar happens for collecting the credentials for the secret. An
> example of this description would perhaps be:
> ```progress 60%: creating initial realm and clients```
>
Looks good to me 👍.  On the apb side, we could have a simple module handle putting the description in place for the broker to find.  With bind, we exec into the pod.  So like you said, we can use the same workflow.

PR #505, opened 2017-10-19
Bug 1504729 - Log job state when getting last op

djzager:
Simple enhancement so that we can see the state of the last operation when the service-catalog requests it.

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -1012,7 +1012,7 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
 	*/
 	a.log.Debugf("service_id: %s", req.ServiceID)
 	a.log.Debugf("plan_id: %s", req.PlanID)
-	a.log.Debugf("operation:  %s", req.Operation) // this is provided with the provision. task id from the work_engine
+	a.log.Debugf("operation:  %s", req.Operation) // Operation is the job token id from the work_engine
 
 	// TODO:validate the format to avoid some sort of injection hack
 	jobstate, err := a.dao.GetState(instanceUUID.String(), req.Operation)
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -1010,9 +1010,9 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
 
 		if async, provision: it should create a Job that calls apb.Provision. And write the output to etcd.
 	*/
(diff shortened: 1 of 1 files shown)


pkg/broker/broker.go
```diff
@@ -1002,9 +1002,9 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
 
 		if async, provision: it should create a Job that calls apb.Provision. And write the output to etcd.
 	*/
-	a.log.Debug(fmt.Sprintf("service_id: %s", req.ServiceID)) // optional
-	a.log.Debug(fmt.Sprintf("plan_id: %s", req.PlanID))       // optional
-	a.log.Debug(fmt.Sprintf("operation:  %s", req.Operation)) // this is provided with the provision. task id from the work_engine
+	a.log.Debugf("service_id: %s", req.ServiceID)
```
eriknelson: Thanks for killing the fmt.Sprint, there's a lot of that left over from the old days :)

pkg/broker/broker.go
```diff
@@ -1002,9 +1002,9 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
...
+	a.log.Debugf("plan_id: %s", req.PlanID)
+	a.log.Debugf("operation:  %s", req.Operation) // this is provided with the provision. task id from the work_engine
```
eriknelson: Nit: the comment is a little inaccurate, it'll be there with every async op, not just provision.

pkg/broker/broker.go
```diff
@@ -1004,7 +1004,7 @@ func (a AnsibleBroker) LastOperation(instanceUUID uuid.UUID, req *LastOperationR
 	*/
 	a.log.Debugf("service_id: %s", req.ServiceID)
 	a.log.Debugf("plan_id: %s", req.PlanID)
-	a.log.Debugf("operation:  %s", req.Operation) // this is provided with the provision. task id from the work_engine
```
jmrodri: Why is this an erroneous comment? The operation is expected to be the job token (task_id) from the work_engine that happened during the provision. At a minimum I'd remove provision and make it clear that we expect operation to be the job token id.

eriknelson: I just remarked it wasn't entirely accurate since all async jobs will set that to be the job token, not just provision.

jmrodri:
If this was changing something in docs or build file stuff I would let it slide. This is actually changing mainline broker code so a bugzilla is required.

rthallisey:
https://github.com/openshift/ansible-service-broker/issues/475

PR #510, opened 2017-10-20
Bug 1502044 - add buffer size and work_engine test

jmrodri:
during a refactoring of the work engines to support async deprovision
and update we missed the buffer size.

fixes 1502044

### DIFF
--- a/pkg/broker/work_engine_test.go
+++ b/pkg/broker/work_engine_test.go
@@ -1,0 +1,106 @@
+package broker
+
+import (
+	"fmt"
+	"testing"
+
+	ft "github.com/openshift/ansible-service-broker/pkg/fusortest"
+)
+
+var engine *WorkEngine
+
+func init() {
+	engine = NewWorkEngine(10)
+}
+
+type mockSubscriber struct {
+	buffer <-chan WorkMsg
+	called bool
+}
+
+func (ms *mockSubscriber) Subscribe(buffer <-chan WorkMsg) {
+	ms.buffer = buffer
+	ms.called = true
+}
+
+type mockMsg struct {
+	msg string
+}
+
+func (mm mockMsg) Render() string {
+	return mm.msg
+}
+
+type mockWorker struct {
+	called bool
+}
+
+func (mw mockWorker) Run(token string, buffer chan<- WorkMsg) {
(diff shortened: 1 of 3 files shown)


pkg/broker/work_engine_test.go
```diff
@@ -0,0 +1,106 @@
...
+	buffer <- mockMsg{msg: "hello"}
+}
+
+func TestNewWorkEngine(t *testing.T) {
+	we := NewWorkEngine(10)
+	ft.AssertNotNil(t, we)
+	ft.AssertEqual(t, we.bufsz, 10)
+}
+
+func TestGetActiveTopics(t *testing.T) {
```
shawn-hurley: :tada: :+1:

pkg/broker/work_engine_test.go
```diff
@@ -0,0 +1,106 @@
...
+	faketopic = "fake"
+	dasub := mockSubscriber{}
+	err := engine.AttachSubscriber(&dasub, faketopic)
+	if err == nil {
+		t.Fail()
+	}
+	ft.AssertEqual(t, "invalid work topic", err.Error(), "invalid error")
+}
+
+func TestStartNewJob(t *testing.T) {
```
rthallisey: This function still needed?

pkg/broker/work_engine_test.go
```diff
@@ -0,0 +1,106 @@
...
+	return mm.msg
+}
+
+type mockWorker struct {
+	called bool
+}
+
+func (mw mockWorker) Run(token string, buffer chan<- WorkMsg) {
+	fmt.Println("xxxxxxxxxxxxxxxxxxxxx run called")
+	mw.called = true
```
rthallisey: The code where this gets used is commented out.  Does ```TestStartNewJob``` need to be uncommented or should we remove the mockWorker struct?

jmrodri: I'll clean that up. I think I wsa in the middle of creating a test that didn't quite work because I needed to mock something else.

pkg/broker/work_engine_test.go
jmrodri: @rthallisey I changed that test to actually do the test I was expecting. Should work now.

PR #519, opened 2017-10-26
update resource field names

jmontleon:

 - Update resource field names in CI files

depends-on N/A

fixes N/A

### DIFF
--- a/scripts/broker-ci/mediawiki123.yaml
+++ b/scripts/broker-ci/mediawiki123.yaml
@@ -4,8 +4,8 @@ metadata:
   name: mediawiki
   namespace: default
 spec:
-  externalClusterServiceClassName: dh-mediawiki123-apb
-  externalClusterServicePlanName: default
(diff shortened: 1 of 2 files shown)


scripts/broker-ci/mediawiki123.yaml
```diff
@@ -4,8 +4,8 @@ metadata:
   name: mediawiki
   namespace: default
 spec:
-  externalClusterServiceClassName: dh-mediawiki123-apb
-  externalClusterServicePlanName: default
+  clusterServiceClassExternalName: dh-mediawiki123-apb
+  clusterServicePlanExternalName: default
```
jmrodri: We should not be leaking service-catalog names into our apbs. It shouldn't matter what these are called as long as they get translated on the broker side. IMO. We should define what we want them to be and use that.

jmrodri:
Disregard my deleted comment. I thought I was looking at apb yaml. +1

PR #520, opened 2017-10-27
Bug 1476173 - Skip deprovision if the namespace is being deleted

cfchase:
Skip deprovision if the namespace is being deleted since we can't access it and it's going to be deleted anyway.

### DIFF
--- a/pkg/apb/deprovision.go
+++ b/pkg/apb/deprovision.go
@@ -22,8 +22,11 @@ package apb
 
 import (
 	logging "github.com/op/go-logging"
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/openshift/ansible-service-broker/pkg/metrics"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/apb/deprovision.go
```diff
@@ -73,3 +87,17 @@ func Deprovision(
...
+	if err != nil {
+		return false, err
+	}
+
+	namespace, err := k8scli.CoreV1().Namespaces().Get(name, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	return namespace == nil || namespace.Status.Phase == "Terminating", nil
```
shawn-hurley: Would prefer to use the [constant](https://godoc.org/k8s.io/api/core/v1#NamespacePhase) to compare the value with.

cfchase: thanks, didn't realize that existed.  Fixed

PR #525, opened 2017-10-31
when building the broker for image always build for linux OS.

### LINKED ISSUE #521: ASBD fails to run in image when created on a Mac
When you run ‘make build-image’ on a Mac the container that is created copy’s the Mac generated binary into the Linux container.

shawn-hurley:
Always build the broker binary for OS of Linux and output to the build directory

fixes #521

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,8 @@ $(KUBERNETES_FILES):
 prepare-local-env: $(KUBERNETES_FILES) ## Prepare the local environment for running the broker locally
 	@echo > /dev/null
 
-build-image: broker ## Build a docker image with the broker binary
-	cp broker build/broker
+build-image: ## Build a docker image with the broker binary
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,8 +47,8 @@ $(KUBERNETES_FILES):
 prepare-local-env: $(KUBERNETES_FILES) ## Prepare the local environment for running the broker locally
 	@echo > /dev/null
 
-build-image: broker ## Build a docker image with the broker binary
-	cp broker build/broker
+build-image: ## Build a docker image with the broker binary
+	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/broker ./cmd/broker
```
jmrodri: The only downside to this is that it will build everytime, where the previous version only did it if the broker was out of date. One could argue though you wouldn't be calling build-image if you didn't cause a change

PR #527, opened 2017-11-01
Bug 1507111 - Add support for a local OpenShift Registry adapter

dymurray:
This PR addresses Bug 1507111 to add an adapter for the internal OpenShift registry.

 - New adapter file
 - New clusterrolebindings for asb service account to access imagestream resource
 - More openshift client changes so that we can access the imagestream resource

### DIFF
--- a/pkg/clients/openshift.go
+++ b/pkg/clients/openshift.go
@@ -1,9 +1,11 @@
 package clients
 
 import (
+	b64 "encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"unsafe"
 
 	logging "github.com/op/go-logging"
@@ -29,7 +31,38 @@ var oldAllowAllPolicyRule = PolicyRule{APIGroups: nil, Verbs: []string{"*"}, Res
 
 // OpenshiftClient - Client to interact with openshift api
 type OpenshiftClient struct {
-	restClient rest.Interface
+	authRestClient  rest.Interface
+	imageRestClient rest.Interface
+}
+
+type imageLabel struct {
+	Spec string `json:"com.redhat.apb.spec"`
+}
+
+type containerConfig struct {
+	Labels imageLabel `json:"Labels"`
+}
+
+type imageMetadata struct {
+	ContainerConfig containerConfig `json:"ContainerConfig"`
+}
+
+type image struct {
+	DockerImage string        `json:"dockerImageReference"`
+	Metadata    imageMetadata `json:"dockerImageMetadata"`
+}
+
+// ImageList is a resource you can create to determine which actions another user can perform in a namespace
+type ImageList struct {
+	metav1.TypeMeta `json:",inline"`
+	// Items holds the image data
+	Items []image `json:"items"`
+}
+
+// FQImage is a struct to map FQNames to Imagestreams
+type FQImage struct {
+	Name        string
+	DecodedSpec []byte
 }
 
 // SubjectRulesReview is a resource you can create to determine which actions another user can perform in a namespace
@@ -235,22 +268,31 @@ func newOpenshift(log *logging.Logger) (*OpenshiftClient, error) {
 }
 
 func newForConfig(c *rest.Config) (*OpenshiftClient, error) {
-	config := *c
-	if err := setConfigDefaults(&config); err != nil {
+	authConfig := *c
+	imageConfig := *c
+	if err := setConfigDefaults(&authConfig, "/apis/authorization.openshift.io"); err != nil {
 		return nil, err
 	}
-	client, err := rest.RESTClientFor(&config)
+	authClient, err := rest.RESTClientFor(&authConfig)
 	if err != nil {
 		return nil, err
 	}
 
-	return &OpenshiftClient{restClient: client}, nil
+	if err := setConfigDefaults(&imageConfig, "/apis/image.openshift.io"); err != nil {
+		return nil, err
+	}
+	imageClient, err := rest.RESTClientFor(&imageConfig)
+	if err != nil {
+		return nil, err
+	}
+	return &OpenshiftClient{authRestClient: authClient, imageRestClient: imageClient}, nil
 }
 
-func setConfigDefaults(config *rest.Config) error {
+func setConfigDefaults(config *rest.Config, APIPath string) error {
 	gv := v1.SchemeGroupVersion
 	config.GroupVersion = &gv
-	config.APIPath = "/apis/authorization.openshift.io"
+	config.APIPath = APIPath
+	//	config.APIPath = "/apis/authorization.openshift.io"
 	config.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
 
 	if config.UserAgent == "" {
@@ -271,7 +313,7 @@ func (o OpenshiftClient) SubjectRulesReview(user, namespace string, log *logging
 	body.APIVersion = "authorization.openshift.io/v1"
 	b, _ := json.Marshal(body)
 	r := &SubjectRulesReview{}
-	err = o.restClient.Post().
+	err = o.authRestClient.Post().
 		Namespace(namespace).
 		Resource("subjectrulesreviews").
 		Body(b).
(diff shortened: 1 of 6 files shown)


dymurray:
https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/159

dymurray:
Only thing I'm still working on is a unit test.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -81,6 +81,35 @@ objects:
     verbs: ["get", "post", "put", "patch", "delete"]
 
 - apiVersion: v1
+  kind: ClusterRole
```
shawn-hurley: Can we just add these to the existing cluster role for asb service account?

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	logging "github.com/op/go-logging"
+	"github.com/openshift/ansible-service-broker/pkg/apb"
+	"github.com/openshift/ansible-service-broker/pkg/clients"
+	yaml "gopkg.in/yaml.v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
+)
+
+const localOpenShiftName = "openshift-registry"
```
shawn-hurley: If the registry name is hardcoded we should make sure that one someone makes a registry they don't have this name for the registry.

jmrodri:
Please run `make check`. Travis failed because the linter failed. The CI actually worked.

pkg/clients/openshift.go
```diff
@@ -288,3 +330,56 @@ func (o OpenshiftClient) SubjectRulesReview(user, namespace string, log *logging
...
+		return nil, err
+	}
+
+	return fqList, nil
+}
+
+// ListRegistryImages - List images in internal OpenShift registry
+func (o OpenshiftClient) ListRegistryImages(log *logging.Logger) (images []string, err error) {
+	r := &ImageList{}
+	var imageList []string
```
jmrodri: Flip lines 371 and 370. I almost didn't see where `r` was being assigned.

pkg/clients/openshift.go
```diff
@@ -288,3 +330,56 @@ func (o OpenshiftClient) SubjectRulesReview(user, namespace string, log *logging
...
+	err = o.imageRestClient.Get().
+		Resource("images").
+		Do().
+		Into(r)
+
+	for _, image := range r.Items {
+		imageList = append(imageList, strings.Split(image.DockerImage, "@")[0])
+	}
+	if err != nil {
```
jmrodri: This is checking the error return from the `imageRestClient`? If there's an error from the client how can `r` be filled out and have items?

I think line 380 error check should be on line 377 BEFORE you loop through `r.Items`.

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+func (r LocalOpenShiftAdapter) RegistryName() string {
+	return localOpenShiftName
+}
+
+// GetImageNames - retrieve the images
+func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
+	r.Log.Debug("LocalOpenShiftAdapter::GetImageNames")
+	r.Log.Debug("BundleSpecLabel: %s", BundleSpecLabel)
+
+	openshiftClient, err := clients.Openshift(r.Log)
```
jmrodri: you are eating the `err`. If that `clients.Openshift` failed, line 51 will die a miserable death which might be a panic.

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return images, nil
+}
+
+// FetchSpecs - retrieve the spec for the image names.
+func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, error) {
+	r.Log.Debug("LocalOpenShiftAdapter::FetchSpecs")
+	r.Log.Debugf("HERE %v", r.Config.Namespaces)
+	specList := []*apb.Spec{}
+	registryIP, err := r.getServiceIP("docker-registry", "default")
```
jmrodri: eating another `err`, you might be able to do something like
```
var registryIP WHATEVERITRETURNS
var err error

if registryIP, err := r.getServiceIP("docker-registry", "default"); err != nil {
   r.logErrorf("Failed to load registry images")
   return nil, err
}
```

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+
+	openshiftClient, err := clients.Openshift(r.Log)
```
jmrodri: another err eaten

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+			r.Log.Debugf("Image does not have proper registry IP prefix. Something went wrong.")
+		}
+	}
+
+	return specList, nil
+}
+
+func (r LocalOpenShiftAdapter) loadSpec(yamlSpec []byte) (*apb.Spec, error) {
+	r.Log.Debug("LocalOpenShiftAdapter::LoadSpec")
+	var err error
```
jmrodri: weird to declare a var when you can easily do `err := yaml.Unmarshal` on line 103.

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,125 @@
...
+func (r LocalOpenShiftAdapter) getServiceIP(service string, namespace string) (string, error) {
+	k8scli, err := clients.Kubernetes(r.Log)
+	if err != nil {
+		return "", err
+	}
+
+	serviceData, err := k8scli.CoreV1().Services(namespace).Get(service, meta_v1.GetOptions{})
+	if err != nil {
+		r.Log.Warningf("Unable to load service '%s' from namespace '%s'", service, namespace)
+		return "", nil
```
jmrodri: So we got an error but we're returning nil?

pkg/registries/adapters/local_openshift_adapter.go
dymurray: For this I was following what was done in the other adapters. For example the dockerhub adapter hardcodes the names to `docker.io`. But yes I agree.

templates/deploy-ansible-service-broker.template.yaml
dymurray: I actually don't think I can because the other cluster roles are being created under the rbac apiGroup (apiVersion `rbac.authorization.k8s.io/v1beta1`. I am going to test it to see if there's a way to consolidate them but I thought that would cause roadbumps.

dymurray:
@jmrodri @shawn-hurley Thanks for the feedback guys. I think I've addressed everything except for the hardcoded name. Let me know your thoughts on that.

pkg/clients/openshift.go
```diff
@@ -288,3 +330,56 @@ func (o OpenshiftClient) SubjectRulesReview(user, namespace string, log *logging
...
 	return authorization.ConvertAPIPolicyRulesToRBACPolicyRules(pr), nil
 }
+
+// ConvertRegistryImagesToSpecs - Return APB specs from internal OCP registry
+func (o OpenshiftClient) ConvertRegistryImagesToSpecs(log *logging.Logger, imageList []string) ([]FQImage, error) {
+	var fqList []FQImage
+	fqImage := FQImage{}
+	var err error
+	r := &ImageList{}
+	err = o.imageRestClient.Get().
```
shawn-hurley: Sorry, didn't catch this one before, but should check the error before working with the items.

dymurray: Thanks, fixed

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -0,0 +1,137 @@
...
+
+// LocalOpenShiftAdapter - Docker Hub Adapter
+type LocalOpenShiftAdapter struct {
+	Config Configuration
+	Log    *logging.Logger
+}
+
+// RegistryName - Retrieve the registry name
+func (r LocalOpenShiftAdapter) RegistryName() string {
+	return localOpenShiftName
```
rthallisey: I think the name can be changed of the registry. Think this should be like rhcc: ```return r.Config.URL.String()```

rthallisey: The internal registry uses the api. Name will always be the same.

PR #529, opened 2017-11-02
Bug 1476173 - Cleanup deleting namespaces

cfchase:
Redo deprovision of apbs in terminating namespaces to work when autoescalate is false.  Skip user permission verification if the namespace is being deleted and clean up our etcd without running deprovision.

### DIFF
--- a/pkg/broker/deprovision_subscriber.go
+++ b/pkg/broker/deprovision_subscriber.go
@@ -74,7 +74,7 @@ func (d *DeprovisionWorkSubscriber) Subscribe(msgBuffer <-chan WorkMsg) {
 			}
 
 			// Job is not reporting error, cleanup after deprovision
-			err = cleanupDeprovision(dmsg.PodName, instance, d.dao, d.log)
+			err = cleanupDeprovision(instance, d.dao, d.log)
 			if err != nil {
 				d.log.Error("Failed cleaning up deprovision after job, error: %s", err.Error())
 				// Cleanup is reporting something has gone wrong. Deprovision overall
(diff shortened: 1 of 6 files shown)


pkg/handler/handler.go
```diff
@@ -521,9 +538,11 @@ func (h handler) unbind(w http.ResponseWriter, r *http.Request, params map[strin
 			})
 			return
 		}
-		if ok, status, err := h.validateUser(userInfo.Username, serviceInstance.Context.Namespace); !ok {
-			writeResponse(w, status, broker.ErrorResponse{Description: err.Error()})
-			return
+		if !nsDeleted {
```
shawn-hurley: Don't we need to check and not run the pod during unbind like deprovision?

pkg/broker/deprovision_subscriber.go
```diff
@@ -105,7 +105,7 @@ func setFailedDeprovisionJob(dao *dao.Dao, dmsg *DeprovisionMsg) {
 }
 
 func cleanupDeprovision(
-	podName string, instance *apb.ServiceInstance, dao *dao.Dao, log *logging.Logger,
```
shawn-hurley: 👍

pkg/handler/handler.go
cfchase: I haven't put this in for now, since I believe launch_apb_on_bind would have to be turned on.  I can add it, as long as we think it's worth doing now.

cfchase: Fixed

PR #536, opened 2017-11-06
Bug 1508724 - Improve logging when using an incorrect tag

rthallisey:
If you set an incorrect tag for your images in the broker config, the logging
does not indicate what the error is.  Instead, make a suggestion
about the error and print out the request body.

 - Improve logging when an incorrect tag is used

### DIFF
--- a/pkg/registries/adapters/adapter.go
+++ b/pkg/registries/adapters/adapter.go
@@ -21,8 +21,10 @@
 package adapters
 
 import (
+	"bytes"
 	b64 "encoding/base64"
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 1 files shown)


pkg/registries/adapters/adapter.go
```diff
@@ -91,8 +93,14 @@ func imageToSpec(log *logging.Logger, req *http.Request, image string) (*apb.Spe
...
 
-	err = json.NewDecoder(resp.Body).Decode(&hist)
+	// resp.Body is an io.Reader, which are a one time use.  Save the
+	// contents to a byte[] for debugging, then remake the io.Reader for the
+	// JSON decoder.
+	debug, _ := ioutil.ReadAll(resp.Body)
+	r := bytes.NewReader(debug)
+	err = json.NewDecoder(r).Decode(&hist)
 	if err != nil {
+		log.Errorf("Image '%s' may not exist in registry: %s", image, debug)
```
shawn-hurley: I think using the registry name here, rather then the response body, might make more sense and then you can remove the copy of the byte array?

rthallisey: This error check really is catching two error scenarios: A json decoding error and a request error.  I'm going to change this to two different checks by looking for a bad http status code.

I thought about that adding in the registry, but it's embedded in the image variable: ```http://<registry>/<image>/<tag>```.

rthallisey:
The logging error came from https://bugzilla.redhat.com/show_bug.cgi?id=1508724
Do we want an addition bz for this?

shawn-hurley:
I just re-use that BZ, I think this is something worth merging IMO

PR #537, opened 2017-11-06
First pass at last_operation description proposal

maleck13:

Proposal to allow APBs to write a last operation description and have the broker collect and return it as part of the last_operation API
https://github.com/openshift/ansible-service-broker/issues/475

**Note**
While I am becoming more familiar with the broker code, I am still very new to the code base so would appreciate an feedback on the approach outlined here.

On the APB side I would need to investigate the code in order to better understand how we could add a module to allow writing to the log file.

### DIFF
--- a/docs/proposals/last-operation-description.md
+++ b/docs/proposals/last-operation-description.md
@@ -20,27 +20,23 @@ Without the description, we limit the feedback that can be provided to the user 
 - The last_operation demonstrates APB progress.
 - No restrictions or requirements for user to demonstrate APB progress.
 - No guarantee that last_operation shows every operation.
-- The final_operation is the last_operation gathered before the APB is deleted.
 
 
 ## Terms
 
-**Last Operation:**  the most recent operation an apb performed 
-
-**Final operation:** the operation that indicates the apb is 100% complete
+**Last Operation:**  the most recent operation an APB performed 
                      
 
 ## Proposed Solution
 
-Using a new apb module, allow for a description to be added by the apb developer for the last operation the apb took along with the final operation the apb took. This module would take advantage of env vars,  provided to it via the downward api, that reference the pod name and namespace the apb is executing within. 
-A PR is already in place to expose this information: https://github.com/openshift/ansible-service-broker/pull/546
-When called this apb module would update known annotations on the pod ie: ```apb_last_operation``` and ``` apb_final_operation```
-with the description provided by the apb developer. This would be part of the 
+Using a new APB module, allow for a description to be added by the APB developer for the last operation the APB took. This module would take advantage of env vars, provided to it via the downward api, that reference the pod name and namespace the APB is executing within. 
+A PR is already in place to expose this information: https://github.com/openshift/ansible-service-broker/pull/546 .
+When called this APB module would update a known annotation on the pod ie: ```apb_last_operation```
+with the description provided by the APB developer. This would be part of the 
 [ansible playbook modules](https://github.com/ansibleplaybookbundle/ansible-asb-modules).
 
 In order to collect this information we would use a watch via the Kubernetes client on the pod resource within the temporary namespace [Pod Rest API](https://docs.openshift.com/container-platform/3.5/rest_api/kubernetes_v1.html#list-or-watch-objects-of-kind-pod).
-This would allow us to react to changes made (i.e to the annotations) on Pod Object. Whenever a change occurred, an update to the JobStatus would happen. If the ```apb_final_operation``` annotation is present this would take precedence over the last_operation annotation. 
-Once the pod was deleted we would stop the watch on the pod and update the JobStatus ```final_operation``` annotation value.
+This would allow us to react to changes made (i.e to the annotations) on Pod Object. Whenever a change occurred, an update to the JobStatus would happen. Once the pod was deleted we would stop the watch on the pod and update the JobStatus with the final ```last_operation``` annotation value.
 
 Sudo code example
 
@@ -60,38 +56,34 @@ changes := wi.ResultChan()
 
 ```
 
-As this would block, it would need to be done in a background go routine. Using a watch in a background routine should allow us to update the JobState independent of the actual execution of the apb.    
+As this would block, it would need to be done in a background go routine. Using a watch in a background routine should allow us to update the JobState independent of the actual execution of the APB.    
 
 
 ### Broker changes
 Below are a set of changes that I believe are in line with the current design. The exact implementation would likely differ but the gist would be the same.
 
 - A new field would be added to the JobState type:
-```Description string ``` 
-The string value , added by the apb module and gathered from the pod annotation would be stored here.
+```Description string ``` .
+The string value , added by the APB module and gathered from the pod annotation, would be stored here.
 
 - Modify the existing provision and deprovision subscribers along with the corresponding work messages.
-A namespace field would be added to these messages as the pod name is already present. 
-Inside the subscriber, the go routine to watch the pod and update the JobStatus would start when a new work msg was recieved.
-This routine would stop once the a deleted change was received in the watch.
+Only a namespace field would need be added to these messages as the pod name is already present. 
+Inside the subscriber, the go routine to watch the pod and update the JobStatus would start when a new work msg was received.
+This routine would stop once a deleted change was received in the watch.
 
-- Add a new subscriber for bind
-async bind will be part of the service catalog, so having a Status and Description for an async binding will also be needed.
-Adding a new subscriber and workmsg for binding operations will allow us to update the JobStatus once async binding arrives.
+- Add a new subscriber for bind. Async bind will be part of the service catalog, so having a Status and Description for an async binding will also be needed.
+Adding a new subscriber and work message for binding operations will allow us to update the JobStatus once async binding arrives.
 
 - Modify last operation handler to pull the description out of the Job state and send it back as part of the response.
 
-### APB changes
-Add a module that would handle putting the content from a last operation description into the right place on disk. 
+### APB Module changes
+Add a module that would handle putting the content from a last operation into the correct annotation on the pod.
 
 Something like the following may make sense:
 
 ```
- asb_last_operation_description:
+ asb_last_operation:
    description:"10%: creating deployment and routes"
-
-asb_final_operation_description:
-   description:"100%: keycloak succesfully provisioned"   
 ```   
 
-***Note not very familiar with how the ansible apb works under the hood so would need some guidance here***
+***Note not very familiar with how the ansible apb modules works under the hood so would need some guidance here***
--- a/docs/proposals/last-operation-description.md
+++ b/docs/proposals/last-operation-description.md
@@ -2,10 +2,10 @@
 
 ## Introduction
 As per the OpenService Broker API spec, the last operation response can contain
-a description as well as a status [last operation response](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#body).
+a description as well as a status: [last operation response](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#body).
 This can provide useful information about what is happening during a broker action along with details of the overall progress.
-UI implementors could make valuable use of this information
-to provide a progress bar or just simply a log of actions to show progress.
+UI implementors could make valuable use of this information to provide an indicator of progress and provide feedback while long running
+actions are in progress. An example of a possible description is shown below:
 ```
 60%: succesfully created realm in keycloak
 ```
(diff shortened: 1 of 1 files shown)


maleck13:
@rthallisey As per discussion on the issue, I have made an attempt at a proposal here to implement last operation description. Any feedback welcome when you get a chance

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Problem Description
+Currently we are only making use of the status field: 
+[last operation in broker.go](https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/broker.go#L1311)
+The difficulty is around how we get that detail back from the pod.
+Without the description, we limit the feedback that can be provided to the user.
+
+
+## Proposed Implementation
+Create an append only log file in a specified location ```/var/log/last_operation```, which is collected periodically and also a final time after
```
rthallisey: terms: 
final operation - the operation that indicates the apb is 100% complete
last_operation - the most recent operation an apb performed

One thing I like about waiting to collect the final operation is that we know the apb is complete.  That might come in handy with dependencies.  On the other hand, does the broker really need the final operation if it's only a description?  How do you view the importance of the final operation?

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,68 @@
...
+the apb has completed but before the pod is deleted. 
+In the bind workflow, we already do something similar by execing into the container in order to collect
```
djzager: I'm wondering if all of this is possible without using a file inside the APB container. Kubernetes already has a [downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) for exposing things like Pod Name and Pod Namespace combined with [Kubernetes annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) means that we could simply have an annotation(s) for `last_operation` and `final_operation` available through API calls w/o needing to exec.

```
$ kubectl annotate pods asb-1-h597s last_operation='Last operation added'
pod "asb-1-h597s" annotated
$ kubectl get pods asb-1-h597s -o go-template='{{ .metadata.annotations.last_operation }}'
Last operation added%
```

maleck13: @djzager Interesting that you mention the annotations approach. This was something I also thought about. I shied away from it as the file approach seemed more fitting given the pod was already being exec'd into. I agree though this could be a simpler approach, particularly taking into account using the downward API to inform the apb. 
Curious on your thoughts about using a watch while the pod was executing, would this be a reasonable way to pick up updates to the annotations in the broker?
I will rework the proposal to make use of the downward API and pod annotations.

shawn-hurley: So, the APB pod that is running would add annotations on itself?

maleck13: Yes that is my understanding. I believe the apb module could make this transparent to the apb developer

djzager: I like the idea of using watch, unfortunately this is where I get a little fuzzy with how the broker handles the provision/deprovision jobs, their associated channels, and writing state information to etcd. My immediate impression is that we may need to make some changes to the broker in order to support additional status changes if we are going to write the last_operation to etcd.

You _may_ be able to get by with handling the `in progress` state with an associated request for the `last_operation` annotation on the pod when the service catalog is asking for the last operation.

shawn-hurley: I am wondering if using the labels instead of the annotation might make more sense. I don't fully understand the difference but just starting the discussion.

maleck13: I am a bit hazy on the difference also.  However I know labels are used as selectors etc and for querying via the client. Whereas I have often seen annotations used for specific meta data for example https://github.com/openshift/origin/blob/master/examples/quickstarts/cakephp-mysql-persistent.json#L11
So I think that in this case annotations would make more sense.

maleck13: A question on this approach: If we go with the watch on the pod, I am wondering would it at some point also make sense to do the same thing for the credentials from a bind. Rather than execing into the pod, have the pod create a secret in the pods namespace and have the broker watch for secrets in that namespace. 
Obviously this is separate from this work but it could be a good paradigm to use for these kind of interactions? I can create a separate issue for this to discuss further.

rthallisey: I don't know the security impact, if at all, of the apb creating a secret and populating it. But, I think it's a good idea.  The broker could also create it and the apb populate it.  There are a few avenues we could take here. We'll have to explore this separately.

> So I think that in this case annotations would make more sense.
>
+1 for annotations. I think labels are meant for kubernetes services identifying other kubernetes services.

> Curious on your thoughts about using a watch while the pod was executing, would this be a 
> reasonable way to pick up updates to the annotations in the broker?
>
If the broker watches the resource,  it should allow us to always get the final operation because the logic will exist in the broker.  We can grab the final operation right before the namespace is cleaned up. +1

jmrodri:
@djzager we may need an experiment with the annotations to see if that is something we can definitely do.

jmrodri:
I'm assuming this is what we're talking about: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

maleck13:
@jmrodri yes that is what we're talking about

djzager:
It would be nice if I could thread conversations. I'm working on an experiment for you @jmrodri by creating a broker that passes the `pod_name` and `pod_namespace` to the pod and creating a hack APB that annotates itself.

shawn-hurley:
@djzager I think you should be able to pull those values out of the container that it is running in. I think you should be able to get the [pod name](https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/downward-api.md) as well

shawn-hurley:
I thought there was some way to pull this data from the pod, I know namespace is possible, but pod name might be a little harder

djzager:
You may have heard me celebrating @shawn-hurley that this worked.

```
	pod := &v1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Name:   executionContext.PodName,
			Labels: labels,
		},
		Spec: v1.PodSpec{
			Containers: []v1.Container{
				{
					Name:  "apb",
					Image: spec.Image,
					Args: []string{
						action,
						"--extra-vars",
						extraVars,
					},
					Env: []v1.EnvVar{
						v1.EnvVar{
							Name: "POD_NAME",
							ValueFrom: &v1.EnvVarSource{
								FieldRef: &v1.ObjectFieldSelector{
									FieldPath: "metadata.name",
								},
							},
						},
						v1.EnvVar{
							Name: "POD_NAMESPACE",
							ValueFrom: &v1.EnvVarSource{
								FieldRef: &v1.ObjectFieldSelector{
									FieldPath: "metadata.namespace",
								},
							},
						},
					},
					ImagePullPolicy: pullPolicy,
					VolumeMounts:    volumeMounts,
				},
			},
			RestartPolicy:      v1.RestartPolicyNever,
			ServiceAccountName: executionContext.ServiceAccount,
			Volumes:            volumes,
		},
	}
```

djzager:
I'm actually genuinely surprised that it was this easy and it worked on the first ~~try~~tries. I updated the broker with these lines (we may want to use volumes in practice but :man_shrugging:):

```
// pkg/apb/executor.go
		Spec: v1.PodSpec{
			Containers: []v1.Container{
				{
					Name:  "apb",
					Image: spec.Image,
					Args: []string{
						action,
						"--extra-vars",
						extraVars,
					},
					Env: []v1.EnvVar{
						v1.EnvVar{
							Name: "POD_NAME",
							ValueFrom: &v1.EnvVarSource{
								FieldRef: &v1.ObjectFieldSelector{
									FieldPath: "metadata.name",
								},
							},
						},
						v1.EnvVar{
							Name: "POD_NAMESPACE",
							ValueFrom: &v1.EnvVarSource{
								FieldRef: &v1.ObjectFieldSelector{
									FieldPath: "metadata.namespace",
								},
							},
						},
					},
					ImagePullPolicy: pullPolicy,
					VolumeMounts:    volumeMounts,
				},
			},
			RestartPolicy:      v1.RestartPolicyNever,
			ServiceAccountName: executionContext.ServiceAccount,
			Volumes:            volumes,
		},
```

And then I created a custom hello-world-apb whose last step was to annotate itself:

```
- name: annotate myself
  k8s_v1_pod:
    name: "{{ lookup('env', 'POD_NAME') }}"
    namespace: "{{ lookup('env', 'POD_NAMESPACE') }}"
    annotations:
      status: Finished
```

Then I ran the apb, keeping the namespace after completion:

```
$ oc get pods -n apb-push-hello-world-apb-prov-6l746 -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
      openshift.io/scc: restricted
      status: Finished
```

:man_dancing:

jmrodri:
@djzager so would an apb be annotating itself regularly? Like if I'm giving status of my installation, will I have to call that annotation multiple times? What are the restrictions for the status or annotation fields?

With the logfile approach, we can put ANYTHING in there. So I'd want to be able to get as much information as possible. Also, will annotations overwrite themselves? Say the apb writes 50%, then 60%, then 70%, then 80%. We only see the 50 and 80 because we didn't catch the 60 and 70 in time. Could this happen?

I like the changes, just want to see what the gotchas will be if any.

jmrodri:
@shawn-hurley you mentioned "pull this data". Are you wondering if there is a need to have the apb annotate itself for the name, etc?

shawn-hurley:
@jmrodri I was referring to being able to use the known location of `/var/run/secretes/serviceaccount/namespace` to retrieve the namespace, and was wondering if the same thing existed for pod name

djzager:
@jmrodri 
```
so would an apb be annotating itself regularly? Like if I'm giving status of my installation, will I have to call that annotation multiple times?
```

An apb could be annotating itself as often as it wants, my guess is that we would want to support annotations like `last_operation` and `final_operation` and there are 2 ways prominent to me on how you could do that 1) use `asb-modules` to have something like `asb_last_operation` and `asb_final_operation` to do this kind of thing under the covers (I like this because it makes it easier for APB developers) 2) do what we are doing in my experiment.

```
What are the restrictions for the status or annotation fields?
```

Annotations are [unstructured key:value pairs](https://github.com/ansible/ansible-kubernetes-modules/blob/master/library/k8s_v1_pod.py#L15-L20) so we could put whatever we want in there but I think simpler the better.

```
With the logfile approach, we can put ANYTHING in there. So I'd want to be able to get as much information as possible. Also, will annotations overwrite themselves? Say the apb writes 50%, then 60%, then 70%, then 80%. We only see the 50 and 80 because we didn't catch the 60 and 70 in time. Could this happen?
```

Everytime we update the annotation field (ie. `last_operation`) it is overwritten so this scenario you describe is possible and likely a common one.

jmrodri:
> > With the logfile approach, we can put ANYTHING in there. So I'd want to be 
able to get as much information as possible. Also, will annotations overwrite 
themselves? Say the apb writes 50%, then 60%, then 70%, then 80%. We
only see the 50 and 80 because we didn't catch the 60 and 70 in time. 
Could this happen?

> Everytime we update the annotation field (ie. `last_operation`) it is overwritten so this scenario you describe is possible and likely a common one.

I only ask in order to set expectations. I can see bugs being written against this when the progress goes from 0 to 100% because we missed the 2 operations in between.

The workflow would be service-catalog is polling the broker's `last_operation` endpoint on a periodic basis. The broker will be *watching* the pod for annotation changes, upon change it will send a message to the channel with the status update. The apb in the pod will be doing its work and calling a module to update the pod annotation.

My original concern might not be a problem there are enough latency in this whole work flow from the polling of the last_operation endpoint to the watching interval down to time it takes for apb to update status. I think it will be okay if we *miss* a few. Also, getting some status is better than what we have today, NO STATUS!

maleck13:
> An apb could be annotating itself as often as it wants, my guess is that we would want to support annotations like last_operation and final_operation and there are 2 ways prominent to me on how you could do that 1) use asb-modules to have something like asb_last_operation and asb_final_operation to do this kind of thing under the covers (I like this because it makes it easier for APB developers)

I think going the asb module route also makes a lot of sense.  That module could then handle retrieving the values of those annotations and then updating the value by appending to the existing one. Ensuring an append only log.
I think from an end user point of view, if it goes from ```10%: creating X``` to then ``` 10%: creating X, 20%: updating Y, 40%: creating X, 80% deleting Z``` you are still getting the full picture which I think is fine.

One thing I think we should discuss is error scenarios. Is there a common or best way to "trap" an error during the apb execution to allow the developer to add some context and update the log before the apb exits?

djzager:
@maleck13 
```
I think from an end user point of view, if it goes from 10%: creating X to then 10%: creating X, 20%: updating Y, 40%: creating X, 80% deleting Z you are still getting the full picture which I think is fine.
```

We could when going the `asb-module` route: 1) get the `last_operation` and 2) make the annotation be `{{ previous_operation}}, 80% deleting` to do as you are suggesting. Personally, I just don't think that we should do that. I believe the `asb-module` should keep writing it's `last_operation` annotation like:

```
// at start
  metadata:
    annotations:
      last_operation: "10%: creating"
// in the middle
  metadata:
    annotations:
      last_operation: "50%: running"
// in the end
  metadata:
    annotations:
      last_operation: "100%: done"
      final_operation: "Woot, we're done"
```

I say this because it seems appropriate for us to be reporting the `last_operation` when asked not all the operations performed.

maleck13:
@djzager Yeah I see where you are coming from, and strictly speaking I think you are correct, it should be the last operation only. The concern I would have then is around consistency, as mentioned in previous comment, you could potentially get very different feedback depending on the the timings of the polling.
Is it stretching the use of this API too far to have the last operation present (ie the last thing append to the log) along with the previous operations?
Perhaps this is something that needs discussion with the service catalog to see is there interest in having support for something like ```{status:"in progress", "description":"doing something", "operation_log":["something i did before"]}``` ?

maleck13:
I will wait to update the proposal until we iron out these different pieces

rthallisey:
>@djzager Yeah I see where you are coming from, and strictly speaking I think you are correct, it should 
> be the last operation only. The concern I would have then is around consistency, as mentioned in 
> previous comment, you could potentially get very different feedback depending on the the timings of the 
> polling. Is it stretching the use of this API too far to have the last operation present (ie the last thing 
> append to the log) along with the previous operations?
>
I think it's going to be hard to expect consistency from last_operation because the description field can be anything.  It's plain text so we're not requiring a percentage or anything, but it gives the user a way to demonstrate progress that is meaningful to them.  So to me, it doesn't matter if we're missing any messages because the description isn't guaranteed to be a definition of state.

Here are a summary of my expectations:
 1) The last_operation demonstrates APB progress.
 2) No restrictions or requirements for user to demonstrate APB progress.
 3) No guarantee that last_operation shows every operation.
 4) The final_operation is the last_operation gathered before the APB is deleted.

maleck13:
Ok this sounds like a good definition. Based on the thumbs up looks like there is agreement, so I think I can proceed with an update to the proposal.

maleck13:
Updated proposal based on discussions here

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+Currently we are only making use of the status field: 
+[last operation in broker.go](https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/broker.go#L1311)
+The difficulty is around how we get that detail back from the pod.
+Without the description, we limit the feedback that can be provided to the user for actions performed against the service catalog.
+
+## Expectations
+- The last_operation demonstrates APB progress.
+- No restrictions or requirements for user to demonstrate APB progress.
+- No guarantee that last_operation shows every operation.
+- The final_operation is the last_operation gathered before the APB is deleted.
```
djzager: ~~deleted~~ completes

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+## Terms
+
+**Last Operation:**  the most recent operation an apb performed 
+
+**Final operation:** the operation that indicates the apb is 100% complete
```
djzager: ~~o~~Operation

djzager: I'd also recommend lining this up with the expectations, `**Final Operation:** the last operation before APB completed.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+## Expectations
+- The last_operation demonstrates APB progress.
+- No restrictions or requirements for user to demonstrate APB progress.
+- No guarantee that last_operation shows every operation.
+- The final_operation is the last_operation gathered before the APB is deleted.
+
+
+## Terms
+
+**Last Operation:**  the most recent operation an apb performed
```
djzager: nit: `strings.ToUpper("apb")`, I see this elsewhere

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+
+In order to collect this information we would use a watch via the Kubernetes client on the pod resource within the temporary namespace [Pod Rest API](https://docs.openshift.com/container-platform/3.5/rest_api/kubernetes_v1.html#list-or-watch-objects-of-kind-pod).
+This would allow us to react to changes made (i.e to the annotations) on Pod Object. Whenever a change occurred, an update to the JobStatus would happen. If the ```apb_final_operation``` annotation is present this would take precedence over the last_operation annotation. 
+Once the pod was deleted we would stop the watch on the pod and update the JobStatus ```final_operation``` annotation value.
+
+Sudo code example
+
+```go
+
+wi, err := k8client.CoreV1().Pods(ns).Watch(meta_v1.ListOptions{})
```
djzager: :tada: It wasn't immediately obvious to me how to accomplish this, nice.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+Below are a set of changes that I believe are in line with the current design. The exact implementation would likely differ but the gist would be the same.
+
+- A new field would be added to the JobState type:
+```Description string ``` 
+The string value , added by the apb module and gathered from the pod annotation would be stored here.
+
+- Modify the existing provision and deprovision subscribers along with the corresponding work messages.
+A namespace field would be added to these messages as the pod name is already present. 
+Inside the subscriber, the go routine to watch the pod and update the JobStatus would start when a new work msg was recieved.
+This routine would stop once the a deleted change was received in the watch.
```
djzager: nit: This routine would stop once ~~the~~ a deleted change was received in the watch.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+
+- Add a new subscriber for bind
+async bind will be part of the service catalog, so having a Status and Description for an async binding will also be needed.
+Adding a new subscriber and workmsg for binding operations will allow us to update the JobStatus once async binding arrives.
+
+- Modify last operation handler to pull the description out of the Job state and send it back as part of the response.
+
+### APB changes
```
djzager: Actually don't think that you are targeting the APB specifically here, more the underlying [ansible-asb-modules](https://github.com/ansibleplaybookbundle/ansible-asb-modules/blob/master/library/). I don't think that you would be putting the last operation on disk, you should just modify the running pod's annotation directly with the new `asb_last_operation_description`.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+Add a module that would handle putting the content from a last operation description into the right place on disk. 
+
+Something like the following may make sense:
+
+```
+ asb_last_operation_description:
```
djzager: This applies to both new ansible-asb-module actions, ~~_description~~ from the name.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,97 @@
...
+A PR is already in place to expose this information: https://github.com/openshift/ansible-service-broker/pull/546
+When called this apb module would update known annotations on the pod ie: ```apb_last_operation``` and ``` apb_final_operation```
+with the description provided by the apb developer. This would be part of the 
+[ansible playbook modules](https://github.com/ansibleplaybookbundle/ansible-asb-modules).
+
+In order to collect this information we would use a watch via the Kubernetes client on the pod resource within the temporary namespace [Pod Rest API](https://docs.openshift.com/container-platform/3.5/rest_api/kubernetes_v1.html#list-or-watch-objects-of-kind-pod).
+This would allow us to react to changes made (i.e to the annotations) on Pod Object. Whenever a change occurred, an update to the JobStatus would happen. If the ```apb_final_operation``` annotation is present this would take precedence over the last_operation annotation. 
+Once the pod was deleted we would stop the watch on the pod and update the JobStatus ```final_operation``` annotation value.
```
rthallisey: Do we need two annotations?  If the pod completes, the final_operation = last_operation.  Since we're watching the pod anyway, we'll know when it's complete. That will allow us to get both data points with a single annotation.

maleck13: Right that's actually a good point, and simplifies things further. I will update based on this

docs/proposals/last-operation-description.md
maleck13: 👍

maleck13:
@jmrodri @djzager I have updated based on all feedback here now. PTA and let me know if there is anything further required.

docs/proposals/last-operation-description.md
```diff
@@ -0,0 +1,89 @@
...
+Add a module that would handle putting the content from a last operation into the correct annotation on the pod.
+
+Something like the following may make sense:
+
+```
+ asb_last_operation:
+   description:"10%: creating deployment and routes"
+```   
+
+***Note not very familiar with how the ansible apb modules wors under the hood so would need some guidance here***
```
djzager: nit: ~~wors~~works

djzager: https://github.com/ansibleplaybookbundle/ansible-asb-modules/pull/8 actually has an example for you of using the kubernetes API to create a secret. If you need any more guidance, I think I can help you out.

maleck13:
Fixed the typo. Thanks for the help guys. I am hoping to start looking at implementation next week. However we have a lot going at the moment with the mobile next project as we have finished our PoC and are planning for tech preview, so I will have to see how things pan out.

maleck13:
@jmrodri @rthallisey After spending sometime looking at the implementation of this, I have a question about the subscribers and the role they play, in particular the setting of the JobState. Does it make sense for these subscribers to set the JobState or would it make sense to change the jobs to have access to the DAO and set the JobState directly? 
I believe this would simplify the overall design somewhat. Or is there a principal or design philosophy that I am unaware of here for why the subscribers should be doing this? The channels seems to be about sending a workmsg (so something that should be done async as a piece of background work) whereas setting the state of an executing job seems like something that should be reflected as soon as possible?

PR #540, opened 2017-11-07
Bug 1507111 - Do not force image tag to be IP + Port

dymurray:
This PR changes default behavior of local registry adapter to not error out if APB name does not start with IP+port. This was erroring out when a user's image was being tagged with the svc name instead.

 - Change error to debug statement and continue
 - Set default namespace to `openshift` if not set

### DIFF
--- a/pkg/registries/adapters/local_openshift_adapter.go
+++ b/pkg/registries/adapters/local_openshift_adapter.go
@@ -91,7 +91,7 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
 			continue
 		}
 		if strings.HasPrefix(image.Name, registryIP) == false {
-			r.Log.Debugf("Image does not have a registry IP as prefix. Assuming it is using hostname.")
+			r.Log.Debugf("Image does not have a registry IP as prefix. This might cause problems but not erroring out.")
 		}
 		if r.Config.Namespaces == nil {
 			r.Log.Debugf("Namespace not set. Assuming `openshift`")
--- a/pkg/registries/adapters/local_openshift_adapter.go
+++ b/pkg/registries/adapters/local_openshift_adapter.go
@@ -90,18 +90,20 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
 			r.Log.Errorf("Failed to load image spec")
 			continue
 		}
-		if strings.HasPrefix(image.Name, registryIP) {
-			// Image has proper registry IP prefix
-			spec.Image = image.Name
-			namespace := strings.Split(image.Name, "/")[1]
-			for _, ns := range r.Config.Namespaces {
-				if ns == namespace {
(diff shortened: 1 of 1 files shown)


pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -90,18 +90,20 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
...
-				}
+		if strings.HasPrefix(image.Name, registryIP) == false {
+			r.Log.Debugf("Image does not have a registry IP as prefix. This might cause problems but not erroring out.")
+		}
+		if r.Config.Namespaces == nil {
+			r.Log.Debugf("Namespace not set. Assuming `openshift`")
+			r.Config.Namespaces = append(r.Config.Namespaces, "openshift")
+		}
+		spec.Image = image.Name
+		namespace := strings.Split(image.Name, "/")[1]
```
shawn-hurley: I don't like blindly accessing index of an array

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -90,18 +90,32 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
...
+		}
+		if r.Config.Namespaces == nil {
+			r.Log.Debugf("Namespace not set. Assuming `openshift`")
+			r.Config.Namespaces = append(r.Config.Namespaces, "openshift")
+		}
+		spec.Image = image.Name
+		ns_list := strings.Split(image.Name, "/")
+		var namespace string
+		if len(ns_list) == 0 {
+			r.Log.Errorf("Image [%v] is not in the proper format. Erroring.")
```
shawn-hurley: missing image name in print statement.

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -90,18 +90,32 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
...
-			spec.Image = image.Name
-			namespace := strings.Split(image.Name, "/")[1]
-			for _, ns := range r.Config.Namespaces {
-				if ns == namespace {
-					r.Log.Debugf("Image [%v] is in configured namespace [%v]. Adding to SpecList.", image.Name, ns)
-					specList = append(specList, spec)
-				}
-			}
+		if strings.HasPrefix(image.Name, registryIP) == false {
+			r.Log.Debugf("Image does not have a registry IP as prefix. This might cause problems but not erroring out.")
```
djzager: Why not `Warningf` then?

dymurray: @djzager Because if you look a few lines below it is possible that S2I images will not have any registry prefix at all. I'm indifferent to which type of logging statement it is but with 10 S2I images that would output 10 warning statements in the logs for something that is expected behavior.

Feels weird to warn the user over expected behavior.

djzager: I figured there was a reason. In this case, the text of the message is the warning. Works for me.

Issue #542, opened 2017-11-08
Proposal: using interfaces for DAO operations to allow simple unit tests


maleck13:
> # Feature:
When looking at how I would implement the last_operation description https://github.com/openshift/ansible-service-broker/pull/537
I wondered how I would go about adding tests for it also. I know there are a number of end to end tests that are used, but wondered what appetite there was to improving the ability to add unit tests?

If we were to use interfaces to pick out the relevant parts of the DAO object, I think it would give a good starting place.

I have created a reference PR to show what I mean and am interested in thoughts and feedback.

jmrodri:
I started trying to get coverage report in a branch so that we could see how bad the gap was in our testing. Mocking is something we want to do, the dao, and `RunCommand` are 2 big ones that are holding a lot of this back.

rthallisey:
@maleck13 did code from #543 make it into any of your PR? I don't want to lose out on all those changes to broker_test.go.

maleck13:
The changes to broker_test are not in the PRs but the concepts are there. There is an interface around the DAO actions for the Jobs for instance. Which I think can form the basis for future changes to how testing and dependency injection is approached.
I was worried about the PR getting stale as so many changes are happening

rthallisey:
Ok. We can circle back to this after #619 is merged

PR #543, opened 2017-11-08
[wip] pr to get feedback on approach of using interfaces to aid unit te…

maleck13:
…sting

This is a ***reference*** implementation for issue https://github.com/openshift/ansible-service-broker/issues/542 to show the approach I am outlining.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -239,9 +239,15 @@ func CreateApp() App {
 	metrics.Init(app.log.Logger)
 	app.log.Debug("Creating AnsibleBroker")
 	if app.broker, err = broker.NewAnsibleBroker(
-		app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.Broker,
+		app.dao, app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.Broker,
 	); err != nil {
 		app.log.Error("Failed to create AnsibleBroker\n")
+		app.log.Error(err.Error())
+		os.Exit(1)
+	}
+
+	if err := app.broker.Login(); err != nil {
+		app.log.Error("Failed to login to OpenShift with oc \n")
 		app.log.Error(err.Error())
 		os.Exit(1)
 	}
--- a/pkg/broker/broker_test.go
+++ b/pkg/broker/broker_test.go
@@ -23,6 +23,7 @@ import (
 	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 	ft "github.com/openshift/ansible-service-broker/pkg/fusortest"
+	"github.com/pborman/uuid"
 )
 
 var log = logging.MustGetLogger("handler")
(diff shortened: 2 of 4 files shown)


maleck13:
@jmrodri Ok great. If this approach is something we want to pursue (ie mocks, interfaces and table tests), then it can be done in an incremental approach,  similar to what is shown here. I can tidy this up a little more and maybe create a mocks package to hold some of the more useful mocks. The coverage stats you link to would certainly be very useful also.

pkg/app/app.go
```diff
@@ -243,13 +243,19 @@ func CreateApp() App {
 	metrics.Init(app.log.Logger)
 	app.log.Debug("Creating AnsibleBroker")
 	if app.broker, err = broker.NewAnsibleBroker(
-		app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.Broker,
+		app.dao, app.dao, app.log.Logger, app.config.Openshift, app.registry, *app.engine, app.config.Broker,
```
jmrodri: I can see the dao ending up with a bunch of interfaces, which could blow up this construction method.

pkg/broker/broker_test.go
```diff
@@ -55,3 +56,93 @@ func TestAddIdForPlan(t *testing.T) {
...
+
+func TestAnsibleBroker_LastOperation(t *testing.T) {
+	cases := []struct {
+		Name                 string
+		JobStateDAO          func() JobStateDAO
+		ExpectError          bool
+		InstanceID           uuid.UUID
+		LastOperationRequest *LastOperationRequest
+		Validate             func(t *testing.T, resp *LastOperationResponse)
+		ExpectedCalls        map[string]int
```
jmrodri: I would want to simplify this a bit. Definitely the right idea but this test looks very painful to debug and understand in 30s. I prefer a bunch of small tests vs a test that loops through all cases.

pkg/broker/interfaces.go
```diff
@@ -0,0 +1,10 @@
+package broker
+
+import "github.com/openshift/ansible-service-broker/pkg/apb"
+
+// JobStateDAO defines the actions that can be preformed on the JobState resource
+type JobStateDAO interface {
+	GetState(instanceUUID, operation string) (apb.JobState, error)
+	SetState(id string, state apb.JobState) error
+	GetSvcInstJobsByState(instanceID string, reqState apb.State) ([]apb.JobState, error)
+}
```
jmrodri: I'd like to map out what the interfaces for the dao will be, like I mentioned earlier there could be many which could blow up some of the methods. I think the interfaces are the only solution just need to be smart in how we define them.

pkg/broker/broker_test.go
maleck13: Interesting. I want to be clear that I have no problem in changing this to follow the patterns you feel are best for this project, however in the interest of discussion I will explain my reasons here. You may already know all the following, if so feel free to ignore.

These kind of tests are quite common in go projects, including the stdlib, the built in testing package has builtin support for running sub tests in this way. 

Using the ```t.Run ``` method runs each of these cases as their own test. This produces output as follows:

```
--- PASS: TestAnsibleBroker_LastOperation (0.00s)
    --- PASS: TestAnsibleBroker_LastOperation/test_get_last_operation_successful
```
I think this is quite clear and gives a nice view of what has been tested. Also if the sub test: ``` TestAnsibleBroker_LastOperation/test_get_last_operation_successful ``` fails, you can re run just that single sub test by using ```go test -run ^TestAnsibleBroker_LastOperation/test_get_last_operation_successful$``` to isolate it down, no matter how many cases may be present. This should allow you to debug quite quickly if an individual case is failing.
In my experience have a structure for your test case makes it quite simple not only to add a new case but also to understand the pieces involved in the test and what can be controlled.

As mentioned these are only my own thoughts and I will happily adjust.

pkg/app/app.go
maleck13: Yes agree. Something that may help here, would be to split the broker into the different actions.

In sudo code. Something like :

```go
package broker

type provision struct{}

func NewProvision(<dao interface>, <dao interface>, logger)*provision{}

func (p *provision)Handle(req, ProvisionRequest, ...)(ProvisionResponse,error){}

```
could work?

pkg/broker/broker_test.go
jmrodri: @maleck13 well you taught me something new today. It's an interesting approach for doing positive and negative testing on a single piece of code. The examples on the go blog were easier since they just had strings in their test data. I would not be opposed to it using table driven tests.

pkg/app/app.go
maleck13: I could try this approach for the last operation handler to aid discussion.

maleck13:
@jmrodri going to close this PR as many of the concepts made it into my other PRs

PR #545, opened 2017-11-09
Create release notes script to grab changes

djzager:
First iteration on the idea of automatically generating release notes
emails that we may or may not want to send out.

Example usage:

```
./scripts/release_notes.sh ansible-service-broker-1.0.5-1
```

### DIFF
--- a/scripts/release_notes.sh
+++ b/scripts/release_notes.sh
@@ -1,19 +1,5 @@
 #!/bin/env bash
 
-if [ ! -n "$GH_USER" ] || [ ! -n "$GH_TOKEN" ]; then
-    echo "You must export GH_USER - your github user id"
-    echo "You must export GH_TOKEN - your github access token"
-    exit 1
-fi
-
-if [ ! -n "$1" ]; then
-    echo "Must provide arg"
-    exit 1
-fi
-
-GH_CURL="curl -u $GH_USER:$GH_TOKEN"
-GIST_API="https://api.github.com/gists"
-
 create_gist () {
     local date=$1
     local notes_file=$2
@@ -88,6 +74,22 @@ EOF
     rm $notes_file
 }
 
+##############################
+# Main
+##############################
+if [ ! -n "$GH_USER" ] || [ ! -n "$GH_TOKEN" ]; then
+    echo "You must export GH_USER - your github user id"
+    echo "You must export GH_TOKEN - your github access token"
+    exit 1
+fi
+
+if [ ! -n "$1" ]; then
+    echo "Must provide start tag/branch/sha to compare against"
+    exit 1
+fi
+
+GH_CURL="curl -u $GH_USER:$GH_TOKEN"
+GIST_API="https://api.github.com/gists"
 generate_notes $1
 
 
--- a/scripts/release_notes.sh
+++ b/scripts/release_notes.sh
@@ -56,7 +56,7 @@ EOF
         pr_number=$(echo $commit | sed -n 's/.*(#\([[:digit:]]\+\))$/\1/p')
         bug_number=$(echo $commit| sed -n 's/Bug\s\+\([[:digit:]]\+\).*/\1/p')
         if [ -n "$pr_number" ]; then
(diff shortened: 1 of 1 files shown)


scripts/release_notes.sh
```diff
@@ -0,0 +1,110 @@
...
+    rm $other_file
+
+    create_gist $date $notes_file
+    rm $notes_file
+}
+
+generate_notes $1
+
+
+###
```
djzager: I left this because it was kind of a pain to learn how to use curl to make requests to the GitHub API and I don't want anyone, including myself to have to go through that again if we decide we want more of this.

djzager:
An example generated release notes: https://gist.github.com/djzager/a361f6cf302d7adce62533a1433179e9

scripts/release_notes.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/env bash
+
+if [ ! -n "$GH_USER" ] || [ ! -n "$GH_TOKEN" ]; then
+    echo "You must export GH_USER - your github user id"
+    echo "You must export GH_TOKEN - your github access token"
+    exit 1
+fi
+
+if [ ! -n "$1" ]; then
+    echo "Must provide arg"
```
jmrodri: I'd say what the arg is. I.e. "Must provide start commit sha" or something more explicit. Otherwise I'll just enter "fu"

scripts/release_notes.sh
```diff
@@ -0,0 +1,110 @@
...
+    cat $bug_file >> $notes_file
+    rm $bug_file
+    cat $other_file >> $notes_file
+    rm $other_file
+
+    create_gist $date $notes_file
+    rm $notes_file
+}
+
+generate_notes $1
```
jmrodri: Add a comment above so it is easy to know this is the main :)

```bash
#
# MAIN
#
```

there are some checks at the top, and if you read it top down, it's easy to forget that the main is at the bottom, etc.

scripts/release_notes.sh
```diff
@@ -0,0 +1,110 @@
...
+
+if [ ! -n "$1" ]; then
+    echo "Must provide arg"
+    exit 1
+fi
+
+GH_CURL="curl -u $GH_USER:$GH_TOKEN"
+GIST_API="https://api.github.com/gists"
+
+create_gist () {
```
jmrodri: I would probably add a comment up here to suggest these are the methods or functions for the script:

```bash
#
# FUNCTIONS
#
```

If it looks weird leave it out, but definitely include the one for MAIN.

scripts/release_notes.sh
```diff
@@ -0,0 +1,110 @@
...
+        bug_number=$(echo $commit| sed -n 's/Bug\s\+\([[:digit:]]\+\).*/\1/p')
+        if [ -n "$pr_number" ]; then
+            detailed_commit=$(echo $commit | sed -n 's,\(.*\)(#\([[:digit:]]\+\))$,\1[(#\2)](https://github.com/openshift/ansible-service-broker/pulls/\2),p')
+            if [ ! -z "$bug_number" ]; then
+                echo "* $detailed_commit" | sed -n 's,Bug\s\+\([[:digit:]]\+\)\(.*\),[Bug \1](https://bugzilla.redhat.com/show_bug.cgi?id=\1)\2,p' >> $bug_file
+            else
+                echo "* $detailed_commit" >> $other_file
+            fi
+        fi
+    done <<< "$commits"
```
jmrodri: Nice work parsing the pr numbers and bug numbers :)

djzager:
[Ansible Service Release Notes](https://trello.com/c/qGOrpcaW/555-ansible-service-release-notes)

scripts/release_notes.sh
djzager: I'm skipping this in favor of moving all of the pre-checks down into main for readability.

scripts/release_notes.sh
djzager: Yeah, that was a poor decision on my part.

PR #546, opened 2017-11-10
Expose the pod name/namespace to APB

djzager:
Use kubernetes downward API to expose the pod name and namespace to the
APB. This will give us more flexibility w.r.t. managing APB execution,
secrets, and annotations.

### DIFF
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -128,6 +128,24 @@ func ExecuteApb(
 						"--extra-vars",
 						extraVars,
 					},
+					Env: []v1.EnvVar{
+						v1.EnvVar{
+							Name: "POD_NAME",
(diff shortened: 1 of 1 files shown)


djzager:
I know that @maleck13 was working on the proposal in #537 where APBs could annotate themselves. Also, if we wish to move away from `exec`ing into containers to get APB credentials then the APB will need to know the pod name/namespace information.

@shawn-hurley I remember you suggesting using volumes, I can make that change, I just wanted to get eyeballs on this change.

maleck13:
This looks pretty reasonable to me. Wondering what advantages we would gain from using a volume rather than an envvar?

shawn-hurley:
@djzager I am starting to agree with @maleck13. I think that the env variables will be easy to interact with.

PR #548, opened 2017-11-10
Integrate with coveralls for code coverage

jmrodri:

 - integrating travis-ci with goveralls (coveralls for golang)
 - generate coverage reports of our code to better understand our testing
 - also added a target for developers to see the coverage locally: `make ci-test-coverage`

none

none

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,16 @@ ansible/main.retry
 # rendered templates
 templates/k8s-ansible-service-broker.yaml
 apiserver.local.config*
+
+# test artifacts
+adapters.out
+apb.out
+app.out
+auth.out
+broker.out
+coverage-all.out
+coverage.out
+handler.out
+registries.out
+validation.out
+coverage.html
--- a/.gitignore
+++ b/.gitignore
@@ -56,4 +56,3 @@ coverage.out
 handler.out
 registries.out
 validation.out
-coverage.html
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ test-coverage-html: ## checkout the coverage locally of your tests
 	@$(foreach pkg,$(PACKAGES),\
 		go test -coverprofile=coverage.out -covermode=count $(pkg);\
 		tail -n +2 coverage.out >> coverage-all.out;)
-	@go tool cover -html=coverage-all.out # -o coverage.html
+	@go tool cover -html=coverage-all.out
 
 ci-test-coverage: ## CI test coverage, upload to coveralls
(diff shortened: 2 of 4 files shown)


jmrodri:
TEST
-----
* you can see the output of the coverage quite easily just run `make test-coverage-html`. A browser will open with the output. That same file is used to feed coveralls.

* The `ci-test-coverage` will NOT work locally because it uses the `-service travis-ci` option which works ONLY in travis. I was able to get goveralls to update coveralls locally but it requires the project API KEY in the command which I didn't want to commit to the repo.

Makefile
```diff
@@ -31,7 +33,21 @@ fmtcheck: ## Check go formatting
...
 test: ## Run unit tests
-	@go test ./pkg/...
+	@go test -cover ./pkg/...
+
+test-coverage-html: ## checkout the coverage locally of your tests
+	@echo "mode: count" > coverage-all.out
+	@$(foreach pkg,$(PACKAGES),\
+		go test -coverprofile=coverage.out -covermode=count $(pkg);\
+		tail -n +2 coverage.out >> coverage-all.out;)
+	@go tool cover -html=coverage-all.out # -o coverage.html
```
djzager: So if you are excluding `-o coverage.html` where is the file that I would open in my browser to see it ?

Makefile
```diff
@@ -31,7 +33,21 @@ fmtcheck: ## Check go formatting
...
+
+ci-test-coverage: ## CI test coverage, upload to coveralls
+	@echo "mode: count" > coverage-all.out
+	@$(foreach pkg,$(PACKAGES),\
```
djzager: This is a duplicate from `test-coverage-html`, is there a way for you to not repeat yourself?

.gitignore
```diff
@@ -44,3 +44,16 @@ ansible/main.retry
...
+apb.out
+app.out
+auth.out
+broker.out
+coverage-all.out
+coverage.out
+handler.out
+registries.out
+validation.out
+coverage.html
```
djzager: I don't get the impression that you need this anymore

Makefile
jmrodri: running `go tool cover -html=coverage-all.out` will actually open the browser window for you. With the `-o` it will simply write it to the file. I'll remove the reference to the coverage.html

Makefile
jmrodri: I will look at deduping

jmrodri:
@djzager I have deduped the coverage generation for you. I also removed references to coverage.html since I no longer use it.

PR #549, opened 2017-11-11
remove unneeded paragraph from license header

jmrodri:
We were asked to remove the unneeded paragraph regarding trademarks as it was older language that is no longer used. I copied the original header from some old code :)

### DIFF
--- a/cmd/broker/main.go
+++ b/cmd/broker/main.go
@@ -13,10 +13,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-// Red Hat trademarks are not licensed under Apache License, Version 2.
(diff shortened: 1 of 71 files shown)


jmrodri:
TEST
------
* make sure everything compiles and builds.
* inspect the diff ensure that there is only either a NEW addition of the header OR the trademark paragraph was removed.

FAIL if you see any CODE that was removed because that means it probably didn't have a header to begin with. I didn't see any but I want extra eyes on it.

jmrodri:
[(1) Remove last paragraph in license headers [1]](https://trello.com/c/QE1h5XsT/580-1-remove-last-paragraph-in-license-headers-1)

PR #550, opened 2017-11-13
Proposal to improve bind credential extraction

djzager:
A proposal to remove (really deprecate) `kubectl exec` from bind
credential extraction process.

### DIFF
--- a/docs/proposals/prop-apb-gen-creds.md
+++ b/docs/proposals/prop-apb-gen-creds.md
@@ -1,0 +1,41 @@
+# APB Generate Credentials Secret Proposal
+
+## Introduction
+
+This proposal aims to improve the broker process by which we extract
+credentials from bindable [Ansible Playbook
+Bundles (APB)](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle)
+by allowing the APB to generate a secret with needed credentials instead of
+using `kubectl exec` into the running pod to grab credentials from a file.
+
+Reference Issue #544
+
+## Problem Description
+
+The broker should seek to use available kubernetes API calls to get necessary
+information whenever possible instead of relying on `kubectl exec`. Now that
+the broker supplies pod name and namespace information to running APBs via the
+[Kubernetes Downward
+API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/),
+APBs have all they need to generate secrets inside the APB sandbox namespace to
+be retrieved later by the broker.
+
+## Implementation Details
+
+- [ ] Update the [`asb_encode_binding`
+module](https://github.com/ansibleplaybookbundle/ansible-asb-modules/blob/master/library/asb_encode_binding.py)
+to create a kubernetes secret in the APB sandbox namespace.
+- [ ] Update
+  [`apb-base`](https://github.com/ansibleplaybookbundle/apb-base/tree/master/files/usr/bin):
+  we no longer need to run `bind-init` or `broker-bind-creds`.
+- [ ] Bump the APB version to `1.1`. This will prevent older broker's from grabbing
+  new APBs that it won't be able to handle and allow us to centrally locate our
+  backwards compatibility in the broker.
+- [ ] Update
+  [`pkg/apb/ext_creds.go
+  ExtractCredentials`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/apb/ext_creds.go#L33):
+  If APB version is `1.0`, do it the way we are currently doing it.
+  If APB version is `1.1`, 1) watch pod and wait for it to complete 2) evaluate
+  success/failure of APB execution 3) read credentials from secret.
+- [ ] Bump [`MaxAPBVersion`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/version/apbversion.go#L27)
+  to `1.1`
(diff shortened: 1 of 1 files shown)


djzager:

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Problem Description
+
+The broker should seek to use available kubernetes API calls to get necessary
+information whenever possible instead of relying on `kubectl exec`. Now that
+the broker supplies pod name and namespace information to running APBs via the
+[Kubernetes Downward
+API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/),
+APBs have all they need to generate secrets inside the APB sandbox namespace to
+be retrieved later by the broker.
```
jmrodri: I would split the last sentence fragment so that it is more clear:

APBs will have all they need to generate secrets inside the APB sandbox namespace. The secret will be retrieved by the broker after the pod finishes.

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Implementation Details
+
+- [ ] Update the [`asb_encode_binding`
+module](https://github.com/ansibleplaybookbundle/ansible-asb-modules/blob/master/library/asb_encode_binding.py)
+to create a kubernetes secret in the APB sandbox namespace.
```
jmrodri: Are there any error conditions we need to be able to handle? Anything that could cause the information not to appear in the secret?

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+- [ ] Update
+  [`apb-base`](https://github.com/ansibleplaybookbundle/apb-base/tree/master/files/usr/bin):
+  we no longer need to run `bind-init` or `broker-bind-creds`.
```
jmrodri: 👍 to removing the need for these scripts

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+- [ ] Bump the APB version to `1.1`. This will prevent older broker's from grabbing
```
jmrodri: Are you bumping the min and the max? or something else? Should newer brokers be able to handle older APBs?

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+  new APBs that it won't be able to handle and allow us to centrally locate our
+  backwards compatibility in the broker.
+- [ ] Update
+  [`pkg/apb/ext_creds.go
+  ExtractCredentials`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/apb/ext_creds.go#L33):
+  If APB version is `1.0`, do it the way we are currently doing it.
+  If APB version is `1.1`, 1) watch pod and wait for it to complete 2) evaluate
+  success/failure of APB execution 3) read credentials from secret.
```
jmrodri: This answers my question above.

docs/proposals/prop-apb-gen-creds.md
djzager: Could you elaborate more on this. Off the top of my head the only scenario that I would be concerned about would be if we fail to create the secret. I'm struggling to think of a scenario where we would be able to create the secret but it would be empty.

jmrodri: I don't know :) just trying to come up with worst case scenarios. Can creating a secret fail? if we get errors what will happen? will we mark the pod as failed? today if we don't write the creds in the file we return an error code that exec can see. That way the broker knows when to stop looking. If the pod completes, but no secret exists does this mean there are no secrets or that we failed to write them?

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -0,0 +1,41 @@
...
+  [`apb-base`](https://github.com/ansibleplaybookbundle/apb-base/tree/master/files/usr/bin):
+  we no longer need to run `bind-init` or `broker-bind-creds`.
+- [ ] Bump the APB version to `1.1`. This will prevent older broker's from grabbing
+  new APBs that it won't be able to handle and allow us to centrally locate our
+  backwards compatibility in the broker.
+- [ ] Update
+  [`pkg/apb/ext_creds.go
+  ExtractCredentials`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/apb/ext_creds.go#L33):
+  If APB version is `1.0`, do it the way we are currently doing it.
+  If APB version is `1.1`, 1) watch pod and wait for it to complete 2) evaluate
```
eriknelson: Do we want to deprecate and warn v1.0 APBs?

docs/proposals/prop-apb-gen-creds.md
djzager: Yes. The APB would fail if we couldn't create the secret.

docs/proposals/prop-apb-gen-creds.md
rthallisey: Would this be a major version bump since it breaks compatibility with the 1.0 Broker?

It would be good to have a table somewhere tracking which Broker goes with which APBs that go with which release.  Might come in handy.

PR #551, opened 2017-11-13
using table driven testing

jmrodri:
Using a table driven test case similar to what @maleck13 has proposed. This is using an existing test to see how it compares.

### DIFF
--- a/pkg/broker/util_test.go
+++ b/pkg/broker/util_test.go
@@ -18,7 +18,9 @@ package broker
 
 import (
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


jmrodri:
Let me know what folks think. It might be better for the more advanced tests.

maleck13:
These look reasonable to me. 👍

jmrodri:
@maleck13 thanks for the enlightenment. I never knew you could do this. Once you get the hang of it, it's quite easy to understand and read. 👍

PR #552, opened 2017-11-14
Use the Kubernetes API for namespaces check

rthallisey:
Verify that a namespace exists using the kubernetes API.

 - Fixes issue in k8s where projects don't exist

### DIFF
--- a/pkg/apb/provision_or_update.go
+++ b/pkg/apb/provision_or_update.go
@@ -18,10 +18,8 @@ package apb
 
 import (
 	"errors"
-	"fmt"
 	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/metrics"
-	"github.com/openshift/ansible-service-broker/pkg/runtime"
 )
 
 type executionMethod string
@@ -47,13 +45,6 @@ func provisionOrUpdate(
 		log.Error("apb instance.Spec requires [name] and [image] fields to be separate")
 		log.Error("Are you trying to run a legacy ansibleapp without an image field?")
 		return "", nil, errors.New("No image field found on instance.Spec")
-	}
-
-	ns := instance.Context.Namespace
-	log.Info("Checking if project %s exists...", ns)
-	if !projectExists(ns) {
-		log.Errorf("Project %s does NOT exist! Cannot provision requested %s", ns, instance.Spec.FQName)
-		return "", nil, fmt.Errorf("Project %s does not exist", ns)
 	}
 
 	metrics.ActionStarted(string(method))
@@ -86,9 +77,4 @@ func provisionOrUpdate(
 	}
 
 	return executionContext.PodName, creds, err
-}
-
-func projectExists(project string) bool {
-	_, _, code := runtime.RunCommandWithExitCode("kubectl", "get", "project", project)
-	return code == 0
 }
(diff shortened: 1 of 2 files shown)


shawn-hurley:
NACK, it is assumed that the project is created I think this is a valid check IMO. 

This is probably a pretty good place to determine if that assumption makes sense.

rthallisey:
You can't create a service instance without a valid namespace/project.  I don't think there's a scenario where this code will be false.  Let me know if you have one in mind.

```
[rhallisey@rhev-i16c-04 ansible-service-broker]$ oc create -f  scripts/broker-ci/mediawiki123.yaml 
Error from server (NotFound): error when creating "scripts/broker-ci/mediawiki123.yaml": namespaces "non-existent-namespace" not found
```

shawn-hurley:
When someone hits the endpoint without going through the catalog. We are
just a spec conforming thing, we should not assume our callers are doing
the right thing IMO.

rthallisey:
I think this check came from the asbcli era which used the scenario you mentioned.  There isn't any harm in keeping another check around, but I think the scenario where this check fails is going to be rare. I have the patch to convert to a namespace already around. I'll swap to that one.

pkg/apb/provision_or_update.go
```diff
@@ -49,11 +52,17 @@ func provisionOrUpdate(
...
+
 	ns := instance.Context.Namespace
-	log.Info("Checking if project %s exists...", ns)
-	if !projectExists(ns) {
-		log.Errorf("Project %s does NOT exist! Cannot provision requested %s", ns, instance.Spec.FQName)
-		return "", nil, fmt.Errorf("Project %s does not exist", ns)
+	log.Info("Checking if namespace %s exists.", ns)
+	_, err = k8scli.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
+	if err != nil {
+		return "", nil, errors.New(fmt.Sprintf("Project %s does NOT exist!", ns))
```
djzager:
```
=================================
              Lint               
=================================
pkg/apb/provision_or_update.go:65:19: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)
Found 1 lint suggestions; failing.
Makefile:27: recipe for target 'lint' failed
make: *** [lint] Error 1
```

pkg/apb/provision_or_update.go
```diff
@@ -49,11 +52,17 @@ func provisionOrUpdate(
...
+
 	ns := instance.Context.Namespace
-	log.Info("Checking if project %s exists...", ns)
-	if !projectExists(ns) {
-		log.Errorf("Project %s does NOT exist! Cannot provision requested %s", ns, instance.Spec.FQName)
-		return "", nil, fmt.Errorf("Project %s does not exist", ns)
+	log.Info("Checking if namespace %s exists.", ns)
+	_, err = k8scli.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
+	if err != nil {
+		return "", nil, fmt.Errorf("Project %s does NOT exist!", ns)
```
djzager: One more :wink: 
```
pkg/apb/provision_or_update.go:65:30: error strings should not be capitalized or end with punctuation or a newline
```

You can use `make check` to run through most of the tests before pushing.

rthallisey: I'll move back to what was there :).

rthallisey:
thanks @djzager

PR #555, opened 2017-11-15
Broker should extract credentials from secret

### LINKED ISSUE #544: Proposal using a secret in the bind namespace for credentials
> # Feature:

Based on discussions here:
https://github.com/openshift/ansible-service-broker/pull/537
Creating follow up issue to allow discussion of using a secret in the temporary bind namespace to retrieve the credentials rather than the current approach: having the apb write them to a specific file and have the broker exec into the container in a background loop to check for their existence.

In the referenced issue, we discuss using the pod for the apb to add meta data about the last operation and then retrieving this information via a watch on the pod resource. This raised the question of whether we should use a similar paradigm for the credentials from a bind.

This needs a full proposal, but the basic premise would be to have the apb either create or update a secret with the credentials for the binding. The broker then, instead of exec'ing into the container, would set up a watch on secrets within the temporary namespace that the apb was being executed. It would use the contents of this secret in order to return the credentials to the service catalog.

### LINKED ISSUE #553: pkg/apb/unbind.go does not wait for APB to complete
# Bug

As it stands currently we don't execute APBs on unbind, however, if we were I think there is a potential issue where we could delete an APB sandbox before the APB has completed it's execution:

```
func Unbind(instance *ServiceInstance, parameters *Parameters, clusterConfig ClusterConfig, log *logging.Logger) error {
	log.Notice("                       UNBINDING                              ")
	log.Notice(fmt.Sprintf("ServiceInstance.ID: %s", instance.Spec.ID))
	log.Notice(fmt.Sprintf("ServiceInstance.Name: %v", instance.Spec.FQName))
	log.Notice(fmt.Sprintf("ServiceInstance.Image: %s", instance.Spec.Image))
	log.Notice(fmt.Sprintf("ServiceInstance.Description: %s", instance.Spec.Description))

	sm := NewServiceAccountManager(log)
	executionContext, err := ExecuteApb(
		"unbind", clusterConfig, instance.Spec,
		instance.Context, parameters, log,
	)
	defer sm.DestroyApbSandbox(executionContext, clusterConfig)

	if err != nil {
		log.Errorf("Problem executing apb [%s] unbind:", executionContext.PodName)
		return err
	}

	return nil
}
```

There is no logic in `ExecuteApb` to wait for complete and `DestroyApbSandbox` only checks if it is allowed to delete the namespace before doing so and a Pod that is `in progress` would fall through:

```
	if clusterConfig.KeepNamespaceOnError {
		if pod.Status.Phase == apicorev1.PodFailed || pod.Status.Phase == apicorev1.PodUnknown || getPodErr != nil {
			return false
		}
	}
```

djzager:
This change makes it so the broker can handle secrets that are created
by APBs when using the `asb_encode_binding` module.

- Update the broker so that it can handle secrets generated by the APB
  when `asb_encode_binding` module is used from the asb-modules.
- Update `executor::ExecuteApb` to wait for pod to complete, since the
  pod is no longer kept alive for credential extraction.
- Clean up some of the log messages and code format related to apb
  actions.

Fixes #544 
Fixes #553 
Implements the proposal #550 

Depends on the following PRs:
- [ansible-asb-modules#8](ansibleplaybookbundle/ansible-asb-modules#8) This is how the secret gets generated.
- [apb-base#7](ansibleplaybookbundle/apb-base#7) Remove scripts related to extracting credentials from the containers filesystem.
- [ansible-playbook-bundle#163](ansibleplaybookbundle/ansible-playbook-bundle#163) Bump the APB versions so freshly built APBs will pass version validation checks.

### DIFF
--- a/docs/proposals/prop-apb-gen-creds.md
+++ b/docs/proposals/prop-apb-gen-creds.md
@@ -23,20 +23,29 @@ successfully.
 
 ## Implementation Details
 
-- [ ] Update the [`asb_encode_binding`
-module](https://github.com/ansibleplaybookbundle/ansible-asb-modules/blob/master/library/asb_encode_binding.py)
-to create a kubernetes secret in the APB sandbox namespace.
-- [ ] Update
+- [x] Update the [`asb_encode_binding` module](https://github.com/ansibleplaybookbundle/ansible-asb-modules/blob/master/library/asb_encode_binding.py)
+  to create a kubernetes secret in the APB sandbox namespace.
+  [ansible-asb-modules#8](https://github.com/ansibleplaybookbundle/ansible-asb-modules/pull/8)
+- [x] Update
   [`apb-base`](https://github.com/ansibleplaybookbundle/apb-base/tree/master/files/usr/bin):
   we no longer need to run `bind-init` or `broker-bind-creds`.
-- [ ] Bump the APB version to `1.1`. This will prevent older broker's from grabbing
-  new APBs that it won't be able to handle and allow us to centrally locate our
-  backwards compatibility in the broker.
-- [ ] Update
+  [apb-base#7](https://github.com/ansibleplaybookbundle/apb-base/pull/7)
+- [x] Add `LABEL "com.redhat.apb.runtime"="2"` to apb-base since this is a
+  runtime change.
+  ~~Bump the APB version to `2.0`. This will prevent older broker's from grabbing~~
+  ~~new APBs that it won't be able to handle and allow us to centrally locate our~~
+  ~~backwards compatibility in the broker.~~
+  ~~[ansible-playbook-bundle#163](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/163)~~
+- [x] Update broker to evaluate APB runtime versions, support min/max runtime
+  versions, and assume runtime version is `1` if label doesn't exist.
+- [x] Update
   [`pkg/apb/ext_creds.go
-  ExtractCredentials`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/apb/ext_creds.go#L33):
-  If APB version is `1.0`, do it the way we are currently doing it.
-  If APB version is `1.1`, 1) watch pod and wait for it to complete 2) evaluate
-  success/failure of APB execution 3) read credentials from secret.
-- [ ] Bump [`MaxAPBVersion`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/version/apbversion.go#L27)
-  to `1.1`
+  ExtractCredentials`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/apb/ext_creds.go#L33)
+  **if runtime version `== 1`**: Do it the old way
+  **if runtime version `>= 2`**:
+  1) watch pod and wait for it to complete
+  2) evaluate success/failure of APB execution
+  3) read credentials from secret.
+- [x] ~~Bump [`MinAPBVersion`] and~~
+  ~~[`MaxAPBVersion`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/version/apbversion.go#L27)~~
+  ~~to `2.0`~~
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -116,7 +116,7 @@ func ExecuteApb(
 		Spec: v1.PodSpec{
 			Containers: []v1.Container{
 				{
-					Name:  "apb",
+					Name:  ApbContainerName,
 					Image: spec.Image,
 					Args: []string{
 						action,
@@ -155,12 +155,6 @@ func ExecuteApb(
 	_, err = k8scli.Client.CoreV1().Pods(executionContext.Namespace).Create(pod)
 	if err != nil {
 		log.Errorf("Failed to create pod - %v", err)
-		return executionContext, err
-	}
-
-	err = watchPod(executionContext.PodName, executionContext.Namespace, log)
-	if err != nil {
-		log.Errorf("APB Execution failed - %v", err)
 		return executionContext, err
 	}
 
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -153,6 +153,17 @@ func ExecuteApb(
 
 	log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, executionContext.Namespace))
 	_, err = k8scli.Client.CoreV1().Pods(executionContext.Namespace).Create(pod)
+	if err != nil {
+		log.Errorf("Failed to create pod - %v", err)
+		return executionContext, err
+	}
+
+	err = watchPod(executionContext.PodName, executionContext.Namespace, log)
+	if err != nil {
+		log.Errorf("APB Execution failed - %v", err)
+		return executionContext, err
+	}
+
 	return executionContext, err
 }
 
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -153,10 +153,6 @@ func ExecuteApb(
 
 	log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, executionContext.Namespace))
 	_, err = k8scli.Client.CoreV1().Pods(executionContext.Namespace).Create(pod)
-	if err != nil {
-		log.Errorf("Failed to create pod - %v", err)
-		return executionContext, err
-	}
 
 	return executionContext, err
 }
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -19,78 +19,31 @@ package apb
 import (
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
-	"strings"
-	"time"
 
-	"github.com/openshift/ansible-service-broker/pkg/runtime"
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 
 	logging "github.com/op/go-logging"
 )
 
 // ExtractCredentials - Extract credentials from pod in a certain namespace.
 func ExtractCredentials(
-	podname string, namespace string, log *logging.Logger,
+	podname string,
+	namespace string,
+	log *logging.Logger,
 ) (*ExtractedCredentials, error) {
-	log.Debug("Calling monitorOutput on " + podname)
-	bindOutput, err := monitorOutput(namespace, podname, log)
+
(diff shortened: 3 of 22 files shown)


rthallisey:
Note to reviewers: This will fail CI until all the deps are merged.

dymurray:
I'm a little concerned with bumping the APB spec version. The [versioning doc](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#apb-spec-versioning) explains how that version should be bumped. It is meant to track changes to the spec file itself and *not* meant to track dependency changes.

I've raised the concern with David but I cannot think of a better way to track dependency changes that will break broker functionality unless we strive to be backwards compatible. In that case I would like to see the broker support the old mechanism of execing in if the secret doesn't exist or simply read from the secret if we are using the newest asb_encode_binding.

jmrodri:
@dymurray I tend to agree with you that the version is for the spec. This PR isn't changing the spec definition.

jmrodri:
At one point I suggested we remain backwards compatible, but we seemed to have dropped that for some reason. @djzager do you remember why we stopped being compatible?

djzager:
The lesson learned here for me is that I should have included @dymurray on the original proposal. My bad.

djzager:
The reason we stopped the backwards compatibility train @jmrodri was @shawn-hurley wise intervention about 1-off if statements that change behavior and I think that remains true. I tend to agree with @dymurray's point(s), we'll have to do a meaningful rework of the `ExecuteApb` architecture and flow to handle backwards compatibility correctly. (Call me out if anything I just said was stupid).

jmrodri:
@djzager ok i'm not afraid of some if statements, unless of course they are EVERYWHERE. Then let's re-evaluate the ExecuteApb then to see how to best handle this new requirement. I seriously think we're going to piss people off if we don't allow old APBs to work. Thanks for the summary.

dymurray:
Based on comments in scrum I think this would be a good approach:

* Keep APB spec version at `1.0`.
* Add a new label to the `Dockerfile` for apb-base titled `apb-base-version` and set it to `1.0`.
* In the broker, add a new check during spec bootstrap to check apb-base version
* If apb-base version exists and is `1.0` then we can assume it has the newer `asb_encode_binding` module and will look for a secret for the bind creds
* if apb-base version does *not* exist then assume it doesn't have the newer `asb_encode_binding` module and revert to execing into the pod

djzager:
@dymurray @shawn-hurley @jmrodri @eriknelson 

A bunch of changes have been made for the sake of 1) introducing a new APB runtime version (now at version `2`) and 2) providing backwards compatibility between the 2 runtime versions, specifically with respect to secrets. I still need to update the proposal document to line up with what was actually done but I've included below my method for testing.

## Testing

### Setup
**NOTE** Started with catasb to bring up cluster.

* Build and tag the image:

```
$ make build-image
$ docker tag docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest \
   docker.io/djzager/origin-ansible-service-broker:latest
```

* Update the `broker-config` to include `djzager` org:

```diff
$ oc edit configmap broker-config
+      - type: "dockerhub"
+        name: "dz"
+        url: "docker.io"
+        org: "djzager"
+        tag: "latest"
+        white_list:
+          - ".*rhscl-postgresql-apb$"
```

* Update the broker deployment config to deploy the new broker

```diff
$ oc edit dc asb
-        image: docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest
+        image: docker.io/djzager/origin-ansible-service-broker:latest
```

## Test

1. Provision PostgreSQL into `oldproject`
1. Provision MediaWiki into `oldproject`
1. Bind PostgreSQL to MediaWiki in `oldproject`
1. Verify MediaWiki is installed correctly
1. Unbind PostgreSQL
1. Deprovision PostgreSQL && MediaWiki
1. Provision PostgreSQL V2 (from djzager) into `newproject`
1. Provision MediaWiki into `newproject`
1. Bind PostgreSQL to MediaWiki in `newproject`
1. Verify MediaWiki is installed correctly
1. Unbind PostgreSQL
1. Deprovision PostgreSQL && MediaWiki
1. :sunglasses:

docs/proposals/prop-apb-gen-creds.md
```diff
@@ -23,20 +23,22 @@ successfully.
...
-  If APB version is `1.0`, do it the way we are currently doing it.
-  If APB version is `1.1`, 1) watch pod and wait for it to complete 2) evaluate
-  success/failure of APB execution 3) read credentials from secret.
-- [ ] Bump [`MaxAPBVersion`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/version/apbversion.go#L27)
-  to `1.1`
+  1) watch pod and wait for it to complete
+  2) evaluate success/failure of APB execution
+  3) read credentials from secret.
+- [x] Bump [`MinAPBVersion`] and [`MaxAPBVersion`](https://github.com/openshift/ansible-service-broker/blob/8dda3277/pkg/version/apbversion.go#L27)
+  to `2.0`
```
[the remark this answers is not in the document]
djzager: You are correct.

pkg/registries/adapters/adapter.go
```diff
@@ -137,6 +139,18 @@ func imageToSpec(log *logging.Logger, req *http.Request, image string) (*apb.Spe
 		return nil, err
 	}
 
+	runtime := conf.Config.Label.Runtime
+	if runtime == "" {
+		log.Infof("No runtime label found, assuming 1")
```
rthallisey: Should we translate this a little more since we're logging it?
Something like: ```No runtime label found, using 'exec' to gather bind credentials```

rthallisey:
> Note to reviewers: This will fail CI until all the deps are merged.
>
Backwards compatibility for using exec is in this PR.  If the gate is green, it's testing using exec until the deps are merged.  Once this gets rebased and passes CI again, we can merged the deps and let CI run again to confirm that using a secret in the gate also works.

djzager:

pkg/apb/executor.go
```diff
@@ -153,6 +153,11 @@ func ExecuteApb(
 
 	log.Notice(fmt.Sprintf("Creating pod %q in the %s namespace", pod.Name, executionContext.Namespace))
 	_, err = k8scli.Client.CoreV1().Pods(executionContext.Namespace).Create(pod)
+	if err != nil {
+		log.Errorf("Failed to create pod - %v", err)
+		return executionContext, err
```
eriknelson: Redundant return statement. The error will get bubbled up and should get checked by executor callers, but this is probably still valuable for debug reasons, it just might result in some louder logs, which seems fine to me.

pkg/apb/ext_creds.go
```diff
@@ -24,29 +24,46 @@ import (
...
 	"time"
 
+	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/openshift/ansible-service-broker/pkg/runtime"
+	"github.com/openshift/ansible-service-broker/pkg/version"
 
 	logging "github.com/op/go-logging"
 )
 
+type extractCreds func(
```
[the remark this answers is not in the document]
eriknelson: +1, I like the way this worked out. Nice step towards the executor struct we talked about.

pkg/apb/executor.go
djzager: I'm going to remove the if block entirely since all of our calls to the ExecuteAPB look something like:

```
	executionContext, err := ExecuteApb(
...
	)
        defer runtime.Provider.DestroySandbox(...)
	if err != nil {
		log.Errorf("Problem executing apb [%s] $ACTION_NAME", executionContext.PodName)
		return executionContext.PodName, nil, err
	}
```

pkg/apb/ext_creds.go
```diff
@@ -24,29 +24,46 @@ import (
...
+	return extractCredsFunc(podname, namespace, log, k8s)
 }
 
-func monitorOutput(namespace string, podname string, log *logging.Logger) ([]byte, error) {
+// ExtractCredentialsAsFile - Extract credentials from running APB using exec
+func ExtractCredentialsAsFile(
+	podname string,
+	namespace string,
+	log *logging.Logger,
+	k8s *clients.KubernetesClient,
```
shawn-hurley: Why are we passing in the k8s client? I think the right thing to do here is getting it from the method.

djzager: Fair point.

djzager:
@shawn-hurley I've got the feeling that I missed that discussion about breaking up method declarations.

Before I start, let me just say this...I may be a little OCD (just ask me about scratches on my sunglasses) and may need to just get over myself. Personally, I would prefer **either** break them up in the way I had done it **or** shove it all on one line regardless of length. I say that because one dangling parameter + return list + `{` makes me a sad :panda_face: 

I'll do whatever is agreed on.

1. I need to read the actual go style guide.
1. We should probably document where we diverge from that style so newcomers to the project can use that as reference.

pkg/apb/ext_creds.go
djzager: I understand this to mean why pass it when you can just ask for another one with `k8s, err := clients.Kubernetes(log)`. Is that right?

shawn-hurley: yes, that is exactly what I was thinking

djzager: I hate it that I agree with you b/c then I have to change it. :sunglasses:

djzager: I'm going to wait to push this change until there are thoughts on the method declarations.

rthallisey: I'm fine with either.  I leave it up to the writer to choose as long as the function wouldn't normally fit on one line.

rthallisey: I don't think we need to block on this since we haven't really established a style here. And we would have to change it everywhere to match the style we go with.  We can discuss in a separate issue if we want to pursue it.

rthallisey:
> I'm fine with either.  I leave it up to the writer to choose as long as the function wouldn't normally fit
> on one line.
>
> I don't think we need to block on this since we haven't really established a style here. And we would
> have to change it everywhere to match the style we go with.  We can discuss in a separate issue if
> we want to pursue it.
>

scripts/broker-ci/setup-cluster.sh
```diff
@@ -13,6 +13,7 @@ function cluster-setup () {
 dockerhub_org: ansibleplaybookbundle
 broker_tag: latest
 broker_kind: ClusterServiceBroker
+apbtab: canary
 EOF
```
rthallisey: You can remove this. It doesn't get used by the gate.

PR #556, opened 2017-11-15
Use the rbac API when creating and deleting policy

rthallisey:
Use the rbac API to create Roles.

 - Switch from using a file to create RBAC policy to using the kubernetes API
 - Setup the runtime pkg to abstract resource calls from the broker

depends-on: https://github.com/openshift/ansible-service-broker/pull/561
depends-on: https://github.com/openshift/ansible-service-broker/pull/562
Partially-fixes: https://github.com/openshift/ansible-service-broker/issues/233


djzager:
If you could have a separate PR for the vendor stuff, that would be fantastic.

rthallisey:

rthallisey:
@djzager @shawn-hurley I made a bunch of changes since your reviews.  Going to leave the do-not-merge tag around until you get a chance to review it again.

pkg/apb/svc_acct.go
```diff
@@ -52,172 +50,8 @@ func (s *ServiceAccountManager) CreateApbSandbox(
...
-			return "", err
-		}
-		defer file.Close()
-	} else if err != nil {
-		// Bubble any non-expected errors
-		return "", err
-	}
-
-	return rFilePath, nil
+	p := runtime.Provider{Log: s.log}
```
shawn-hurley: Should the runtime provider just be an `apb package` variable? Also, should we remove this function and just call the provider from the execute APB? These are just thoughts, all around this looks good.

rthallisey: I agree that the runtime var should be an available variable and it needs to be wherever we make cluster calls. I haven't figured out yet how I'm going to approach providing it yet. You're ahead of the game :).

Agree that this function doesn't need to be called.  All the guts are now in the runtime pkg.

PR #558, opened 2017-11-16
adding ability to connnect over SSL w/o authentication.

### LINKED ISSUE #554: Broker should be able to connect to ETCD over TLS while being unauthenticated
# Bug:

**What happened**:
I want to be able to start up ETCD to serve over TLS but un-authenticated
That when I pass the `etcd_ca_file` it is used in the transport.
**How to reproduce it**:
Start ETCD without the trusted ca cert & client-cert-auth but keep the server.crt and server.key, advertise over https. The broker should be able to connect using the ca that signed the cert.

shawn-hurley:
Allows the broker to only connect over TLS while etcd authentication is off.
 - only return the default transport if etcd client cert, client key and ca file are empty
 - if client cert and client key are set, use them in the TLS config

fixes #554


shawn-hurley:
I am going to add some test cases around the etcd_client.

pkg/clients/etcd_test.go
```diff
@@ -0,0 +1,273 @@
...
+			} else {
+				if len(transport.TLSClientConfig.Certificates) != 0 {
+					t.Fatal("Should not have created a certificate")
+				}
+			}
+		})
+	}
+}
+
+func TestEtcdVersion(t *testing.T) {
```
djzager: :+1:

PR #561, opened 2017-11-16
Create a Kubernetes Client struct

rthallisey:
In the Clients pkg, each of the clients should have a Class to
reference.  This creates the Kubernetes class reference so we
can build specific Kubernetes functions into the client class.

 -  Add a Kubernetes class to the client pkg


rthallisey:

PR #564, opened 2017-11-19
Use the global var instance to save client objects

rthallisey:
Instead of using sync to make sure we initialize the clients once, use the ```initializeClients``` function to create the clients the first time and store them in the instances struct.

 - Global variable will save client objects
 - No need to pass around log to the clients pkg

### DIFF
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -69,10 +69,7 @@ func ExecuteApb(
 	}
 
 	secrets := GetSecrets(spec)
-	k8scli, err := clients.Kubernetes()
-	if err != nil {
-		return executionContext, err
-	}
+	k8scli := clients.Kubernetes()
 
 	labels := map[string]string{
 		"apb-fqname": spec.FQName,
--- a/pkg/apb/provision_or_update.go
+++ b/pkg/apb/provision_or_update.go
@@ -52,15 +52,11 @@ func provisionOrUpdate(
 		return "", nil, errors.New("No image field found on instance.Spec")
 	}
 
-	k8scli, err := clients.Kubernetes()
-	if err != nil {
-		log.Error("Something went wrong getting kubernetes client")
-		return "", nil, err
-	}
+	k8scli := clients.Kubernetes()
 
 	ns := instance.Context.Namespace
 	log.Info("Checking if namespace %s exists.", ns)
-	_, err = k8scli.Client.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
+	_, err := k8scli.Client.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
 	if err != nil {
 		return "", nil, fmt.Errorf("Project %s does not exist", ns)
 	}
--- a/pkg/apb/secrets.go
+++ b/pkg/apb/secrets.go
@@ -167,10 +167,7 @@ func paramInSecret(param ParameterDescriptor, secretKeys []string) bool {
 }
 
 func getSecretKeys(secretName, namespace string) ([]string, error) {
-	k8scli, err := clients.Kubernetes()
-	if err != nil {
-		return nil, err
-	}
+	k8scli := clients.Kubernetes()
 
 	secretData, err := k8scli.Client.CoreV1().Secrets(namespace).Get(secretName, meta_v1.GetOptions{})
 	if err != nil {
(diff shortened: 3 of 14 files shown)


rthallisey:
> What's the reasoning for this change?
>
We substitute using sync to guarantee one client creation with creating the clients during broker initialization.
We gain from this:
  - Removal of sync library
  - No need to pass around a log object
  - Simplifies client pkg and removes code

> It looks like all the calling code assumes the client instances have been set. 
>
Yes.  After ```initClients``` has run, you will always have a valid client because the config file in the pod will never change.  We make the same assumption with sync.  Where this could error is if someone adds code that does ```instance.Kubernetes = nil```. <- that code we shouldn't merge.

> We should have an assertion in the accessors that enforces this assumption if that's true. Code can
> get moved around, maybe something isn't initialized with an initialize call, and then things break quietly.
>
Even if code gets moved around, I don't think we should expect a client to work before we run ```initClients```.  After ```initClients```, we should assume the client is always there.

rthallisey:
pending further discusiion

maleck13:
As someone very interested in this project and has used the kubernetes client quite regularly. I have the following thoughts on how the kubernetes client is used currently. Which I hope it is ok to add to this PR. Pls feel free to ignore:

- Having a function referenced directly from a package that is used throughout the code base to get a concrete struct that returns a concrete struct makes writing unit tests tricky.
I find using dependency injection and interfaces a good way to avoid this:

```go
// for things that should only know bout the client
type ClientProvider interface{
    Kubernetes()kubernetes.Interface
}
// for things that should only know about the config
type ConfigProvider interface{
   Config()*rest.Config
}
// for both
type ClientConfigProvider interface{
    ClientProvider
    ConfigProvider
}

type provider struct{}

func(p *provider)Kubernetes()kubernetes.Interface{
...
return clientset.Clientset
}

func(p *provider)Config()*rest.Config{
...
}

// provision_update.go

func provisionOrUpdate(clientProvider clients.ClientProvider ...){}
```
Doing this will allow the use of the kubernetes fake clientset. See an example of how this is used in tests here:
https://github.com/feedhenry/mcp-standalone/blob/master/pkg/data/mobileApp_test.go#L34
Also if needed you could add an ```OpenShift()``` method.
Likely the client provider would need to be constructed at startup (panic on fail) and added to the handler and passed through?

Another option would be to just use the ```kubernetes.Interface``` type as the dependency passed through removing the provider layer while still allowing the fake clientset to be used in unit tests.

Just some food for thought.

eriknelson:
@maleck13 Thanks for your comment! Good food for thought, we're hoping to have a bigger conversation around this in a week or so. This is part of an effort to decouple the APB runtime platform (k8s or OpenShift, currently) from the broker layer. Without getting too much in the weeds, it's expected that a broker function on the level of `provisionOrUpdate` will create an `Executor` object that it will use to "run" the APB method. It should care how or where that APB is actually executed.

Inside the Executor runner, it will have a platform client interface most likely, so it won't be a concrete struct.

Here's the 2nd of 3 phases we're trying to roll this stuff out into the broker; there have been a lot of design conversations in this area over the last few weeks: https://trello.com/c/z919Hy5h/438-25-agnostic-runtime-phase-2

Re: dependency injection, that was the intention with config and log references from the beginning, but over time we've found injecting dependencies as arguments to be quite painful for something referenced as frequently as a config or a logger. What we ended up with was the logger having to get passed through every call stack, and the signatures continue to grow; it becomes tedious and brittle. @shawn-hurley is working to excise this from the codebase with package level accessors, similar to how we're doing it with the client accessors. Any insight into how you deal with this particular problem would be appreciated!

maleck13:
@eriknelson Thanks for this info, it is good to understand this. I will take a look through the trello linked.
Is there a proposal around this work I could read also?

On Dependency injection. I see where you are coming from. The way I have handled common dependencies such as logging or config, is to inject them into a constructor function and have a struct that holds a reference to it, then attach the "methods" to this struct:

```go

type ProvisionUpdateHandler struct{
    logger *log.logger
}

func (ph *ProvisionUpdateHandler)Handle(params ProvisionParams)error{
ph.logger.Debug("....")
..
}

func NewProvisionUpdateHandler(logger *log.logger)*ProvisionUpdateHandler{
     return &ProvisionUpdateHandler{logger:logger}
}

```
Normally I would then construct these dependencies in the ```main``` function or an equivalent making it easy to inject them into any pieces that needed them.

Example:
https://github.com/feedhenry/mcp-standalone/blob/master/cmd/mcp-api/main.go#L135

eriknelson:
@maleck13 Here's the proposal, I believe most of it should be up to date. https://github.com/openshift/ansible-service-broker/blob/master/docs/proposals/add-kubernetes-support.md

rthallisey:
Closing this for now. I think there needs to be further discussion on the topic before we can have a PR merged.

PR #568, opened 2017-11-21
Few fixes to the local deploy templates

rthallisey:
A bunch of fixes when running locally.

 - Deleting the asb service causes issues. The service-ca (asb-tls) gets deleting causing later prep_local_dev_env runs to fail and the catalog fails to connect to the broker.
 - Use the proper path when deleting ci resources
 - Use a deployment config for asb-etcd
 - Point the asb-etcd resources at the etcd app and service
 - Remove the BROKER_PORT variable since it was causing an issue with template rendering and isn't needed

depends-on <PR>

fixes <issue_number>


rthallisey:
CI is down right now. Going to merge without a green CI pass since it doesn't use this code path because it's only for running the broker locally.

PR #571, opened 2017-11-27
Adding ability to warn and filter out bad specs.

### LINKED ISSUE #388: Missing type field on parameter causes empty catalog request
**This form is for bug reports and feature requests. Major features will go through a [spec process](https://github.com/openshift/ansible-service-broker/blob/master/CONTRIBUTING.md).**

# Bug:
Omitting type from a parameter causes an empty response when requesting the catalog.  

**What happened**:
After uploading an APB with an error in it (in this case missing the type field was missing on a parameter), requests to the broker catalog came back empty.

* Defaults the type for the parameter to something (string?)
* I would expect for a bad APB to be missing from the response, but I would expect the other APBs to be returned when requesting the catalog.  Seems like bad APBs might get uploaded so we should handle that more gracefully.

**How to reproduce it**:
1. Create an APB with a single parameter with only the name field.
```yaml
name: hello-world-apb
plans:
  - name: default
    parameters:
      - name: test_param_string
#       type: string # commented out will result in empty catalog
```
2. Build, push the apb.
```
apb build
...
docker push ...
```
3. Boostrap the broker
```
apb bootstrap # or curl -X POST http://0.0.0.0:1338/v2/bootstrap
```
4. GET request the catatalog
```
apb list -o json # or curl -X GET http://0.0.0.0:1338/v2/catalog
```

shawn-hurley:
Will add debug messages when a type from an APB parameter is not defined, and remove from the services for catalog.
 - add error handling to SpecToService

fixes #388

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -516,7 +516,12 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 
 	services = make([]Service, len(specs))
 	for i, spec := range specs {
-		services[i] = SpecToService(spec)
+		ser, err := SpecToService(spec)
+		if err != nil {
+			a.log.Debugf("no adding spec to list of services due to error transforming to service - %v", err)
+		} else {
+			services[i] = ser
+		}
 	}
 
 	return &CatalogResponse{services}, nil
@@ -1321,7 +1326,11 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
 		return nil, err
 	}
 	apb.AddSecretsFor(&spec)
-	service := SpecToService(&spec)
+	service, err := SpecToService(&spec)
+	if err != nil {
+		a.log.Debugf("spec was not added due to issue with transformation to serivce - %v", err)
+		return nil, err
+	}
 	metrics.SpecsLoaded(apbPushRegName, 1)
 	return &CatalogResponse{Services: []Service{service}}, nil
 }
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -518,7 +518,7 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 	for i, spec := range specs {
 		ser, err := SpecToService(spec)
 		if err != nil {
-			a.log.Debugf("no adding spec to list of services due to error transforming to service - %v", err)
+			a.log.Errorf("not adding spec to list of services due to error transforming to service - %v", err)
 		} else {
 			services[i] = ser
 		}
--- a/pkg/broker/util.go
+++ b/pkg/broker/util.go
@@ -35,7 +35,11 @@ type formItem struct {
 
 // SpecToService converts an apb Spec into a Service usable by the service
 // catalog.
-func SpecToService(spec *apb.Spec) Service {
+func SpecToService(spec *apb.Spec) (Service, error) {
+	plans, err := toBrokerPlans(spec.Plans)
+	if err != nil {
+		return Service{}, err
+	}
 	retSvc := Service{
 		ID:            spec.ID,
 		Name:          spec.FQName,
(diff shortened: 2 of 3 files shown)


pkg/broker/broker.go
```diff
@@ -516,7 +516,12 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 
 	services = make([]Service, len(specs))
 	for i, spec := range specs {
-		services[i] = SpecToService(spec)
+		ser, err := SpecToService(spec)
+		if err != nil {
+			a.log.Debugf("no adding spec to list of services due to error transforming to service - %v", err)
```
djzager: not?

shawn-hurley: 🤦‍♂️

jmrodri: debugf for an error?

djzager: :sunglasses:

pkg/broker/broker.go
```diff
@@ -1321,7 +1326,11 @@ func (a AnsibleBroker) AddSpec(spec apb.Spec) (*CatalogResponse, error) {
 		return nil, err
 	}
 	apb.AddSecretsFor(&spec)
-	service := SpecToService(&spec)
+	service, err := SpecToService(&spec)
+	if err != nil {
+		a.log.Debugf("spec was not added due to issue with transformation to serivce - %v", err)
```
jmrodri: do we intend to log an error as DEBUG? seems odd.

pkg/broker/util.go
```diff
@@ -187,35 +195,44 @@ func createUIFormItem(pd apb.ParameterDescriptor, paramIndex int) (interface{},
...
+		return []schema.PrimitiveType{schema.BooleanType}, nil
 	case "number":
-		return []schema.PrimitiveType{schema.NumberType}
+		return []schema.PrimitiveType{schema.NumberType}, nil
 	case "nil", "null":
-		return []schema.PrimitiveType{schema.NullType}
+		return []schema.PrimitiveType{schema.NullType}, nil
 	}
-	return []schema.PrimitiveType{schema.UnspecifiedType}
+	return nil, fmt.Errorf("Could not find the parameter type for: %v", paramType)
```
jmrodri: so what sparked this patch originally?

pkg/broker/util.go
```diff
@@ -250,23 +267,26 @@ func parametersToSchema(plan apb.Plan) Schema {
...
+func extractProperties(params []apb.ParameterDescriptor) (map[string]*schema.Schema, error) {
 	properties := make(map[string]*schema.Schema)
 
 	var patternRegex *regexp.Regexp
-	var err error
-
 	for _, pd := range params {
 		k := pd.Name
 
+		t, err := getType(pd.Type)
```
jmrodri: Why not just check to see if the type is UnspecifiedType? I'm not sure I like the error on the `getType`. Is it really an error on getType? It looked for one, didn't fine a good one, said it was unspecified.

```
t := getType(pd.Type)
if t == []schema.PrimitiveType{schema.UnspecifiedType} {
    return properties
}
```

pkg/broker/util.go
shawn-hurley: #388 sparked this patch. 

The basics of the issue are that the UnspecifiedType does not marshall into json, therefore the only things I can think to do are:
1. either catch this by converting all `specs` to `services` during load time of the specs 

2. When `/catalog` is called and we convert `specs` to `services` we catch that error/unspecified type and do not add it to the list of `services` but do not blow up the entire response. 

3. blow up the entire response w/ a 500. 


I choose to go down this route because if we can not gracefully handle the Unspecified type in the response then I think we should treat that as an error.

pkg/broker/broker.go
shawn-hurley: I think that warning might be better. This is an APB gave us bad data which we have not treated as an error before I don't think it should be errorf IMO

pkg/broker/util.go
shawn-hurley: I think it is reasonable that an ununderstandable type is an error. Especially since we can not use that type for it's intended purpose. If you feel strongly though, I can change the initial error here instead of in `getType` function.

PR #572, opened 2017-11-30
Initial proposal for dealing with network isolation SDNs

shawn-hurley:
This PR will define how we can deal with joining networks that are isolated to allow APBs to talk to the resulting pods that it creates over the network.

 - Implement interface and initial implementation for `redhat/ovs-multitenant` network plugin

### DIFF
--- a/docs/proposals/Network-Isolation.md
+++ b/docs/proposals/Network-Isolation.md
@@ -5,13 +5,16 @@ This [redhat/ovs-multitenant network plugin](https://docs.openshift.com/containe
 
 > The ovs-multitenant plug-in provides OpenShift Container Platform project level isolation for pods and services. Each project receives a unique Virtual Network ID (VNID) that identifies traffic from pods assigned to the project. Pods from different projects cannot send packets to or receive packets from pods and services of a different project.
 
-The ASB creates a transient namespace while running an APB and grants the correct access to the target namespace. This network plugin will cause APBs that assume they can reach, over the network the pod in the target namespace, to fail. 
+The ASB creates a transient namespace while running an APB and grants the correct access to the target namespace. This network plugin will cause some APBs to fail, if the APB assumes it can access pods in the target namespace over the network.
 
 ## Problem Description
 The transient namespace does not have network access to the target namespace, leaving the APB pod unable to perform all the tasks that it should be able to perform.
 
-## <Implementation Details>
-There are ways to manage this [network](https://docs.openshift.com/container-platform/3.6/admin_guide/managing_networking.html). We should be able to add the transient namespace to the same network as the target namespace. One of the big things is we want this to be easily expandable to [kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/networking) SDN's that could implement the same basic structure. This PR will not address each networking option but rather create a common structure for implementing more SDN's in the future.
+## Implementation Details
+With no network access to the target namespace, APBs will be limited in some of the more advanced tasks they can perform. We need a mechanism to give the APBs in the transient namespace access to the target namespace. We should be able to add the transient namespace to the same network as the target namespace. See the [managing networking](https://docs.openshift.com/container-platform/3.6/admin_guide/managing_networking.html) document for different ways of working with networking.
+
+One of the big things is we want this to be easily expandable to [kubernetes SDN's](https://kubernetes.io/docs/concepts/cluster-administration/networking) that could implement the same basic structure. This proposal will not address each networking option but rather create a common structure for implementing more SDN's in the future.
+
 
 ### Steps to take.
 1. Inside the runtime package, determine if adding networks is necessary.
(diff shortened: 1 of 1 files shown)


shawn-hurley:
[(5) Investigate and Enhancements to support network isolation in multi-tenet environment](https://trello.com/c/EET2clCJ/589-5-investigate-and-enhancements-to-support-network-isolation-in-multi-tenet-environment)

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,72 @@
+# Network Isolation
+
+## Introduction
+This [redhat/ovs-multitenant network plugin](https://docs.openshift.com/container-platform/3.6/architecture/additional_concepts/sdn.html#architecture-additional-concepts-sdn) will restrict namespaces from having network traffic flow between them. 
+
+> The ovs-multitenant plug-in provides OpenShift Container Platform project level isolation for pods and services. Each project receives a unique Virtual Network ID (VNID) that identifies traffic from pods assigned to the project. Pods from different projects cannot send packets to or receive packets from pods and services of a different project.
+
+The ASB creates a transient namespace while running an APB and grants the correct access to the target namespace. This network plugin will cause APBs that assume they can reach, over the network the pod in the target namespace, to fail.
```
jmrodri: Let's try to reword that last sentence.

> This network plugin will cause some APBs to fail, if the APB assumes it can access pods in the target namespace over the network.

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,72 @@
...
+This [redhat/ovs-multitenant network plugin](https://docs.openshift.com/container-platform/3.6/architecture/additional_concepts/sdn.html#architecture-additional-concepts-sdn) will restrict namespaces from having network traffic flow between them. 
+
+> The ovs-multitenant plug-in provides OpenShift Container Platform project level isolation for pods and services. Each project receives a unique Virtual Network ID (VNID) that identifies traffic from pods assigned to the project. Pods from different projects cannot send packets to or receive packets from pods and services of a different project.
+
+The ASB creates a transient namespace while running an APB and grants the correct access to the target namespace. This network plugin will cause APBs that assume they can reach, over the network the pod in the target namespace, to fail. 
+
+## Problem Description
+The transient namespace does not have network access to the target namespace, leaving the APB pod unable to perform all the tasks that it should be able to perform.
+
+## <Implementation Details>
```
jmrodri: why is this heading in angle brackets?

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,72 @@
...
+There are ways to manage this [network](https://docs.openshift.com/container-platform/3.6/admin_guide/managing_networking.html). We should be able to add the transient namespace to the same network as the target namespace. One of the big things is we want this to be easily expandable to [kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/networking) SDN's that could implement the same basic structure. This PR will not address each networking option but rather create a common structure for implementing more SDN's in the future.
```
jmrodri: Sorry rewriting the above paragraph to make it more clear in my mind:

> With no network access to the target namespace, APBs will be limited in some of the more advanced tasks they can perform. We need a mechanism to give the APBs in the transient namespace access to the target namespace. We should be able to add the transient namespace to the same network as the target namespace. See the [managing networking](https://docs.openshift.com/container-platform/3.6/admin_guide/managing_networking.html) document for different ways of working with networking.

> One of the big things is we want this to be easily expandable to [kubernetes](https://kubernetes.io/docs/concepts/cluster-administration/networking) SDN's that could implement the same basic structure. This proposal will not address each networking option but rather create a common structure for implementing more SDN's in the future.

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,72 @@
...
+            ...
+            if p.networkIsolation != nil {
+                if err := p.networkIsolation(...); err != nil {
+                    .... Log Statements ...
+                    return err
+                }
+            }
+    }
+    ```
+2. If we do need to add networks together, then during the apb sandbox creation, we should join the networks.
```
jmrodri: how would the `preSandboxCreation` hooks we talked about be affected by this?

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,72 @@
...
+    1. Examples of how to do this are in the `oadm` client in origin. [here](https://github.com/openshift/origin/blob/1f270ca122306656b228faa92bc71d2136e0f97a/pkg/oc/admin/network/project_options.go#L157) and the [update](https://github.com/openshift/origin/blob/master/pkg/network/netid.go#L73) to the annotation.
+3. If we can not join the networks together, and we should, then we should error out of the provision.
```
jmrodri: change #3 to read:

> If we should join the networks together but are not able to, then we should error out of the provision

docs/proposals/Network-Isolation.md
shawn-hurley: So, I think we can either make this a postSandboxCreation action that gets added once all the checks pass, or we can create a new function that is specific to joining networks similar to extract credentials is what I was thinking.

I think that if we went the first option this would be a good example of how to use the hooks from within our codebase.

docs/proposals/Network-Isolation.md
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    type provider struct {
+        ...
+        networkIsolation NetworkIsolation 
+
+    }
+
+    func NewRuntime() {
+         networkIsolation := shouldJoinNetworks(...)
```
rthallisey: Not blocking this since it's pseudo code, but I don't think we want to expose openshift specific calls outside of the coe interface.

How about something like: ```p.coe.shouldJoinNetworks``` or something more general: ```p.coe.initCoe```.

PR #573, opened 2017-12-04
docs and ci template updates for apb name changes

jmontleon:

 -
 -
 -

depends-on <PR>

fixes <issue_number>


templates/mediawiki.yaml
```diff
@@ -3,7 +3,7 @@ kind: ServiceInstance
 metadata:
   name: mediawiki123
 spec:
-  clusterServiceClassExternalName: dh-mediawiki123-apb
+  clusterServiceClassExternalName: dh-mediawiki-apb
```
djzager: If our ci is using the templates inside their relative projects...do we still need these templates? or the associated steps in `cleanup-ci.sh`?

It looks like we moved to the new CI, but the remnants hanging around may make life more confusing down the road.

rthallisey: I left this around so folks can run ```oc create -f templates/mediawiki.yaml``` and test things by hand.  ```cleanup-ci.sh``` does use them, but it doesn't have to.

PR #574, opened 2017-12-04
Identify the cluster in the NewRuntime call

rthallisey:
Contact the openshift version endpoint to verfiy what cluster
the broker is running on.  If the openshift endpoint isn't present, the
broker is running on kubernetes.

 - Identify which cluster the broker is running on

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -18,19 +18,16 @@ package app
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net"
 	"os"
 	"strings"
 	"time"
 
-	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/wait"
-	kubeversiontypes "k8s.io/apimachinery/pkg/version"
 	"k8s.io/apiserver/pkg/authentication/authenticatorfactory"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	genericoptions "k8s.io/apiserver/pkg/server/options"
@@ -170,31 +167,10 @@ func CreateApp() App {
 	}
 
 	// Initialize Runtime
+	app.log.Debug("Connecting to Cluster")
 	agnosticruntime.NewRuntime(app.log.Logger)
-
-	k8scli, err := clients.Kubernetes(app.log.Logger)
+	agnosticruntime.Provider.ValidateRuntime()
 	if err != nil {
-		app.log.Error(err.Error())
-		os.Exit(1)
-	}
-
-	restclient := k8scli.Client.CoreV1().RESTClient()
-	body, err := restclient.Get().AbsPath("/version").Do().Raw()
-
-	switch {
-	case err == nil:
-		var kubeServerInfo kubeversiontypes.Info
-		err = json.Unmarshal(body, &kubeServerInfo)
-		if err != nil && len(body) > 0 {
-			app.log.Error(err.Error())
-			os.Exit(1)
-		}
-		app.log.Info("Kubernetes version: %v", kubeServerInfo)
-	case kapierrors.IsNotFound(err) || kapierrors.IsUnauthorized(err) || kapierrors.IsForbidden(err):
-		app.log.Error("the server could not find the requested resource")
-		app.log.Error(err.Error())
-		os.Exit(1)
-	default:
 		app.log.Error(err.Error())
 		os.Exit(1)
 	}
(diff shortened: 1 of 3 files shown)


rthallisey:
[(13) Agnostic Runtime phase 3](https://trello.com/c/7grfQIbZ/439-13-agnostic-runtime-phase-3)

pkg/app/app.go
```diff
@@ -197,11 +191,17 @@ func CreateApp() App {
...
+		app.log.Error("the server could not find the requested resource")
+		app.log.Error(err.Error())
+		os.Exit(1)
 	default:
 		app.log.Error(err.Error())
 		os.Exit(1)
 	}
 
+	app.log.Debug("Connecting Dao")
+	app.dao, err = dao.NewDao(app.config.Dao, app.log.Logger)
```
shawn-hurley: Should we handle this error or should we remove the error?

pkg/runtime/runtime.go
```diff
@@ -52,7 +55,35 @@ type kubernetes struct{}
...
+	switch {
+	case err == nil:
+		var kubeServerInfo kubeversiontypes.Info
+		err = json.Unmarshal(body, &kubeServerInfo)
+		if err != nil && len(body) > 0 {
+			log.Error(err.Error())
+			panic(err.Error())
+		}
+		log.Info("OpenShift version: %v", kubeServerInfo)
+		cluster = openshift{}
```
shawn-hurley: Thinking about the future, can we make these a function that returns a `coe`?

rthallisey: I was thinking that the coe interface would stay private so that only functions in the runtime pkg have access to openshift.go and kubernetes.go.  That way, the broker only asks for things it understands like "I want a new sandbox to securely create this apb" instead of "I want a few rbac rules, a new service account, and a new namespace".

What did you have in mind for the future with ```coe```?

pkg/app/app.go
rthallisey: We should be handling it in case the etcd config file is not found.

pkg/runtime/runtime.go
shawn-hurley: well I am not saying to make it public, I am talking about making an unexported helper method that encapsulates the retrieval of an actual `coe` structure.

This is for something like the network joining only for openshift and only when a certain plugin is installed. This means that we would want to do extra stuff while creating a `coe` and it was just a preemptive thing. Once we need it we can always add it as well.

pkg/runtime/runtime.go
```diff
@@ -52,7 +55,35 @@ type kubernetes struct{}
...
+	case kapierrors.IsNotFound(err) || kapierrors.IsUnauthorized(err) || kapierrors.IsForbidden(err):
```
djzager: If we are accepting the scenario where we are not authorized to access the `/version/openshift` endpoint, should we not be ensuring we have permission to do work on a kubernetes cluster first?

I'm struggling to ask this question clearly. So I'm going to ask the same question again, worded differently, in the hopes I can communicate the idea. If we are not authorized / forbidden to access the `/version/openshift` **and** we are not authorized / forbidden to access `/version/kubernetes` (assuming there is a k8s version endpoint somewhere), should we fail?

pkg/runtime/runtime.go
rthallisey: Ok I made an assumption of what you meant based on how I was using NewRuntime.  If I understand what you mean, you would like something similar to what we have with the kubernetes client? I'll post a revision with that assumption.

pkg/runtime/runtime.go
rthallisey: Agreed @djzager.  In ```/pkg/app/app.go```, I added a [failure case](https://github.com/openshift/ansible-service-broker/pull/574/files?diff=split#diff-637a60a871f8b69622bdfd3d3b2fcad1R194) on kubernetes because if this validation fails then we aren't connected to a cluster.

```go
        case kapierrors.IsNotFound(err) || kapierrors.IsUnauthorized(err) || kapierrors.IsForbidden(err):
                app.log.Error("the server could not find the requested resource")
                app.log.Error(err.Error())
                os.Exit(1)
```

pkg/app/app.go
```diff
@@ -172,21 +172,15 @@ func CreateApp() App {
 	// Initialize Runtime
 	agnosticruntime.NewRuntime(app.log.Logger)
 
-	app.log.Debug("Connecting Dao")
-	app.dao, err = dao.NewDao(app.config.Dao, app.log.Logger)
-
 	k8scli, err := clients.Kubernetes(app.log.Logger)
```
djzager: I'm curious, based on the answer to my previous question, if this getting of a Kubernetes client, getting a k8s rest client, and checking the k8s version is something we should push down into the `NewRuntime` function. That way these checks could be collocated and maybe even simplify the error handling on app startup.

rthallisey: This validation could be behind the runtime pkg.  Theoretically, the runtime can be something other than k8s or openshift so we shouldn't assume.

shawn-hurley: isn't that, even more, reason to move the initialization of the client into just the runtime?

rthallisey:
> isn't that, even more, reason to move the initialization of the client into just the runtime?
>
@shawn-hurley's comment disappeared.

Everywhere that we grab the kubernetes client and use it should be a function in runtime.  There are numerous cases of this so we'll do that in different patches.  Feel free to write them if you see them.

PR #577, opened 2017-12-06
Notify the apb what cluster it's running on with extravars

rthallisey:
Send the variable 'cluster' as an extravar to the apb so it can identify which playbook to run based on the runtime.

 - Pass cluster=<kubernetes/openshift> as an extravar to the apb on creation

depends-on: https://github.com/ansibleplaybookbundle/postgresql-apb/pull/23
depends-on: https://github.com/ansibleplaybookbundle/mediawiki-apb/pull/17


pkg/runtime/runtime_test.go
```diff
@@ -16,3 +16,8 @@ func (f fakeProvider) DestroySandbox(podName string, namespace string, targets [
 	//TODO: Write tests for DestroySandbox using the fake kubernetes client
 	return
 }
+
+func (f fakeProvider) GetRuntime() string {
```
djzager: I would like to see tests implemented for this change before giving the :+1:

pkg/apb/executor.go
```diff
@@ -198,6 +198,8 @@ func createExtraVars(context *Context, parameters *Parameters) (string, error) {
 	if context != nil {
 		paramsCopy["namespace"] = context.Namespace
 	}
+
+	paramsCopy["cluster"] = runtime.Provider.GetRuntime()
```
djzager: I'm not sure if I believe this should be an extra variable or an environment variable that APBs should load and have default values for...

djzager: I'm curious the reason that you have this as an argument instead of an environment variable to the APB.

rthallisey: +1 for curiosity David. The reason I went with extravars is because it will overwrite the value of ```cluster``` right from the command line.  Then we don't have to 1) add a default environment var to apb-base, 2) I don't need to rely on the user setting CLUSTER=kubernetes to run the playbooks locally, and 3) I can default the var right in the playbook:
```yaml
vars:
     cluster: kubernetes
```
```
docker run docker.io/ansibleplaybookbundle/mediawiki-apb provision --extra-vars cluster=kubernetes

ansible-playbook playbooks/provision.yaml --extra-vars "cluster=kubernetes"
```

pkg/runtime/runtime_test.go
rthallisey: deal

rthallisey:
@djzager I added the test in. Let me know if you have any more concerns.

PR #581, opened 2017-12-06
Resolve runtime version from local openshift

djzager:
Update the local openshift adapter to read the runtime version off of
the image.

### DIFF
--- a/pkg/clients/openshift.go
+++ b/pkg/clients/openshift.go
@@ -52,7 +52,8 @@ type OpenshiftClient struct {
 }
 
 type imageLabel struct {
-	Spec string `json:"com.redhat.apb.spec"`
+	Spec    string `json:"com.redhat.apb.spec"`
+	Runtime string `json:"com.redhat.apb.runtime"`
 }
 
 type containerConfig struct {
@@ -79,6 +80,7 @@ type ImageList struct {
 type FQImage struct {
 	Name        string
 	DecodedSpec []byte
+	Runtime     string
 }
 
 // SubjectRulesReview is a resource you can create to determine which actions another user can perform in a namespace
@@ -372,6 +374,7 @@ func (o OpenshiftClient) ConvertRegistryImagesToSpecs(log *logging.Logger, image
 				}
 				fqImage.Name = imageName
 				fqImage.DecodedSpec = decodedSpec
+				fqImage.Runtime = image.Metadata.ContainerConfig.Labels.Runtime
 				fqList = append(fqList, fqImage)
 			}
 		}
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -40,6 +40,7 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/broker"
 	"github.com/openshift/ansible-service-broker/pkg/clients"
 	"github.com/openshift/ansible-service-broker/pkg/config"
+	"github.com/openshift/ansible-service-broker/pkg/version"
 	"github.com/pborman/uuid"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/kubernetes/pkg/api/v1"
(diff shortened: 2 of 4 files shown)


djzager:
It is worth noting that `apb push` without using the openshift internal registry does not work at this time:

```
[2017-12-06T21:14:36.277Z] [DEBUG] handler::apbAddSpec
[2017-12-06T21:14:36.277Z] [DEBUG] Successfully decoded pushed spec:
[2017-12-06T21:14:36.277Z] [DEBUG] version: 1.0
name: hello-world-apb
description: deploys hello-world web application
bindable: False
async: optional
metadata:
  displayName: Hello World Mine (APB)
  longDescription: A sample APB which deploys a containerized Hello World web application
  dependencies: ['docker.io/ansibleplaybookbundle/hello-world:latest']
  providerDisplayName: "Red Hat, Inc."
plans:
  - name: default
    description: A sample APB which deploys Hello World
    free: True
    metadata:
      displayName: Default
      longDescription: This plan deploys a Python web application displaying Hello World
      cost: $0.00
    parameters: []

[2017-12-06T21:14:36.278Z] [DEBUG] Unmarshalled into apb.Spec:
[2017-12-06T21:14:36.278Z] [DEBUG] {ID: Runtime:0 Version:1.0 FQName:hello-world-apb Image: Tags:[] Bindable:false Description:deploys hello-world web application Metadata:map[displayName:Hello World Mine (APB) longDescription:A sample APB which deploys a containerized Hello World web application dependencies:[docker.io/ansibleplaybookbundle/hello-world:latest] providerDisplayName:Red Hat, Inc.] Async:optional Plans:[{ID: Name:default Description:A sample APB which deploys Hello World Metadata:map[displayName:Default longDescription:This plan deploys a Python web application displaying Hello World cost:$0.00] Free:true Bindable:false Parameters:[] BindParameters:[] UpdatesTo:[]}]}
```
I'm not certain the best way to address this.

djzager:

rthallisey:
> It is worth noting that apb push without using the openshift internal registry does not work at this time:
>
Can you explain this further?  I think I remember you mentioning it on a call.

djzager:
> Can you explain this further? I think I remember you mentioning it on a call.

@rthallisey when we perform an `apb push` that doesn't use the internal openshift registry, we simply communicate with the broker 1) the name of our image and 2) the contents of our apb.yml. There is no `com.redhat.apb.runtime` label to inspect.

We **could** update the `apb push` command to also send along the runtime version **OR** we could come up with a whole new way that we would like for it to work in the kubernetes case with no internal registry. Either way, this PR doesn't fix those issues.

rthallisey:
We don't have to pass along a runtime since every new apb is going to use the secret for bind creds. Then we would have to default the runtime to 2 in the broker.

djzager:
That doesn't seem unreasonable. We can't really trust that people won't try and push a v1 runtime image but if we warn on the push that we are assuming the runtime version...then I could be down with that change.

rthallisey:
If you pull the latest apb-ansible-module you will always get runtime 2. So if you are on a runtime 1 label and you rebuild your container, it should update to the runtime 2 ansible module.  A warning on the apb tool side would be good.

> We can't really trust that people won't try and push a v1 runtime image 
>
Would you have to install a specific package version of the asb module?

djzager:
The `ansible-asb-modules` do not have the label, the [apb-base image has the label](https://github.com/ansibleplaybookbundle/apb-base/blob/master/Dockerfile-canary#L3). The apb tool does not do any inspection of the `FROM:` line of the Dockerfile before building it and I don't know if the apb tool uses the `--no-cache` option when building the image (this would make it possible for you to have a cached version of apb-base and use that).

This is a development endpoint, so it is __okay__ if it doesn't work because you messed up, but there are a number of ways that we could __think__ that we are looking at a runtime 2 when it is not. I just don't want that fact to go unnoticed if I'm going to update the broker to handle `apb push`.

PR #582, opened 2017-12-07
Remove logging from function and structs.

### LINKED ISSUE #270: Warn about unknown keys found in config.
Got bit by this while debugging the new filter feature, I used "whitelist" instead of the expected "white_list". We should at least warn about unknown keys so users aren't expecting something like this to take when it's an unexpected key.

shawn-hurley:
Removes logging from the function signatures and structures
 - create a utility/logging package to manage the logging.
 - adds `log` var to each package that needs it that is package accessible.
 - removes log from functions and structs

Most of the fixes are in util/logging/logger.go. All the other changes are setting up the logger per package and removing the log from structures and functions

fixes #270

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -20,11 +20,11 @@ import (
 	"encoding/json"
 
 	"github.com/openshift/ansible-service-broker/pkg/config"
-	utillogging "github.com/openshift/ansible-service-broker/pkg/util/logging"
+	logutil "github.com/openshift/ansible-service-broker/pkg/util/logging"
 	"github.com/pborman/uuid"
 )
 
-var log = utillogging.NewLog()
+var log = logutil.NewLog()
 
 // Parameters - generic string to object or value parameter
 type Parameters map[string]interface{}
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -19,10 +19,12 @@ package apb
 import (
 	"encoding/json"
 
-	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/config"
+	utillogging "github.com/openshift/ansible-service-broker/pkg/util/logging"
 	"github.com/pborman/uuid"
 )
+
+var log = utillogging.NewLog()
 
 // Parameters - generic string to object or value parameter
 type Parameters map[string]interface{}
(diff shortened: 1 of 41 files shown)


pkg/apb/types.go
```diff
@@ -19,11 +19,13 @@ package apb
 import (
 	"encoding/json"
 
-	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/config"
+	utillogging "github.com/openshift/ansible-service-broker/pkg/util/logging"
```
jmrodri: After reading it a few times, how about 'logutil` instead of `utillogging`? :)

pkg/app/app.go
```diff
@@ -153,84 +153,86 @@ func CreateApp() App {
 		os.Stderr.WriteString(err.Error() + "\n")
 		os.Exit(1)
 	}
-	fmt.Printf("%#v", app.config)
-
-	if app.log, err = NewLog(app.config); err != nil {
+	c := utillogging.LogConfig{
```
jmrodri: logutil instead of utillogging

pkg/util/logging/logger.go
```diff
@@ -33,47 +32,23 @@ type LogConfig struct {
...
-	config := LogConfig{
-		LogFile: c.GetString("log.logfile"),
-		Stdout:  c.GetBool("log.stdout"),
-		Level:   c.GetString("log.level"),
-		Color:   c.GetBool("log.color"),
-	}
+var backends []logging.Backend
+var logLevel logging.Level
 
+//InitilizeLog - initilize the logging utility
```
jmrodri: space after the //

pkg/util/logging/logger.go
```diff
@@ -33,47 +32,23 @@ type LogConfig struct {
...
+func InitilizeLog(config LogConfig) error {
```
jmrodri: TYPO: InitilizeLog -> InitializeLog

pkg/auth/auth.go
```diff
@@ -24,10 +24,12 @@ import (
 	"path"
 	"strings"
 
-	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/config"
+	utillogging "github.com/openshift/ansible-service-broker/pkg/util/logging"
 )
 
+var log = utillogging.NewLog()
```
jmrodri: Ah so each package does this call. I was searching for where the hell the logger was set. Found it.

pkg/util/logging/logger.go
```diff
@@ -22,7 +22,6 @@ import (
 	"os"
 
 	logging "github.com/op/go-logging"
-	"github.com/openshift/ansible-service-broker/pkg/config"
 )
```
jmrodri: The package of this file is `package app` yet it lives in `util/logging` which is the right package?

rthallisey:
👍 pending Zeus' comment

PR #586, opened 2017-12-11
Bug 1510486 - Return 403 for disallowed actions

shawn-hurley:
Updating the status code for when a user does not have access to do an apb action

fixes bug 1510486


jmrodri:
Consider shortening the commit message. The problem is it will get cut off. How about:

```
Bug 1510486 - Return 403 for disallowed actions
```

PR #593, opened 2017-12-14
Proposal for fixing bug 1512042

shawn-hurley:
Proposal for fixing bug 1512042

### DIFF
--- a/docs/proposals/proposal-bug-1512042
+++ b/docs/proposals/proposal-bug-1512042
@@ -1,0 +1,23 @@
+# Proposal for [Bug 1512042](https://bugzilla.redhat.com/show_bug.cgi?id=1512042)
+
+## Introduction
+The issue is that when a custom namespace (not openshift) is used for holding the images the broker does not currently give access to that namespace when creating the service account for the sandbox. This cause's the resulting transient namespace to get stuck in an error pulling the image. 
+
+# Problem Description
+The problem is that we want to be able to control how the privileges get escalated into that namespace to see/pull the images. This means that we will need a robust solution, that is not a simple bug fix.
+
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the customer namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. 
+
+To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
+
+There is also a third option that we should document, which is to allow everyone to access your images in that custom namespace. This option will work even with the default because the user will always have access then.
+
+### Open Questions
+1. I think this check should be in the handler. This is an authorization check in my mind and should be treated as such.
+2. The runtime package should be used to determine the check, something like `CheckAccessToImage`.
+3. We need to figure out how to handle if the adapter is any other adapter then local_openshift. 
+    1. Add a field on the spec to state the registry/adapter that it is using?
+    2. We will also need to track the namespace that it should check.
+    3. Need to be able to not run this check if the not available.
+    4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user) 
--- a/docs/proposals/proposal-bug-1512042
+++ b/docs/proposals/proposal-bug-1512042
@@ -20,4 +20,61 @@ There is also a third option that we should document, which is to allow everyone
     1. Add a field on the spec to state the registry/adapter that it is using?
     2. We will also need to track the namespace that it should check.
     3. Need to be able to not run this check if the not available.
-    4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user) 
+    4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user, gcr adapter could use their Cloud IAM). Or a calller could want to do their own authorization check using our built in auth model (basic auth example).
+
+### Possible Solutions
+1. Create new fields on the spec. These fields will keep track of the adapters/registries that added the spec. Add a function in the runtime package that will be called by the handler for authorization. This function will take the spec, the originating user structure, and some undefined structures as a variadic. Example:
+```go
+type RuntimeAuthorization func(apb.Spec, apb.UserInfo, interface{}...)
+
+var RuntimeAuthorization RuntimeAuthorization
+```
+2.  Create new fields on the spec sor that it can track the adapters/registries that added this specific spec. Add a interface Authorization for the runtime package. This will have a single method, Authorize that will take a UserInfo struct. This then can be chained, and the runtime package should manage this as this is supposed to be exposed so everyone can add one. We can helper methods to add providers to the authorization on the fly.
+```go
+type UserInfo {
+    .... openservicebroker userinfo
+}
+type Authorization interface {
+        Authorize(UserInfo) bool
+} 
+
+var authz Authorization
+
+func AddAuthorization(a Authorization) {
+    authz = chainAuthz(authz, a) 
+}
+
+func chainAuthz(a, b Authorization){
+    return AuthorizationFunc(func(u UserInfo){
+        if a(u) {
+            return b(u)
+        }
+        return false
+    })
+}
+type AuthorizationFunc func(UserInfo)
+
+func (a AuthorizationFunc) Authorize(u UserInfo) {
+    a(u)
+}
+```
+Using it would be:
+```go
+func addProviderAuth(provider auth.Provider) runtime.Authorization {
+    return runtime.AuthorizationFunc(func(u UserInfo) bool {
+        principal, err := provider.GetPrincipal(r)
+        if principal != nil {
+            log.Debug("We found one. HOORAY!")
+            // we found our principal, stop looking
+            return true
+        }
+        if err != nil {
+            log.Errorf("Error - %v", err)
+        }
+        return false
+    })
+}
+.... 
+runtime.AddAuthorization(addProviderAuth(provider))
+...
+```
--- a/docs/proposals/proposal-bug-1512042.md
+++ b/docs/proposals/proposal-bug-1512042.md
@@ -7,19 +7,19 @@ The issue is that when a custom namespace (not openshift) is used for holding th
 The problem is that we want to be able to control how the privileges get escalated into that namespace to see/pull the images. This means that we will need a robust solution, that is not a simple bug fix.
 
 ## Implementation Details
-The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the customer namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. 
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the custom namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. There will only be two options for the configuration value `CheckAccessToNamespace` and `AutoGrantAccessToNamespace`.
 
 To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
 
-There is also a third option that we should document, which is to allow everyone to access your images in that custom namespace. This option will work even with the default because the user will always have access then.
+There is also a third option that we should document, which is to allow everyone to access the images in that custom namespace. This is purelly an RBAC policy setting and will allow all the users to access the images. 
 
 ### Open Questions
-1. I think this check should be in the handler. This is an authorization check in my mind and should be treated as such.
+1. I think this check should be in the `pkg/handler/handler.go` file. This is an authorization check in my mind and should be treated as such. This check is only needed when an APB is being launched and should follow the placement of the current role authorization check.
 2. The runtime package should be used to determine the check, something like `CheckAccessToImage`.
 3. We need to figure out how to handle if the adapter is any other adapter then local_openshift. 
     1. Add a field on the spec to state the registry/adapter that it is using?
     2. We will also need to track the namespace that it should check.
-    3. Need to be able to not run this check if the not available.
+    3. Need the ability to turn off the check when certain specs from other adapters are the ones being used.
     4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user, gcr adapter could use their Cloud IAM). Or a calller could want to do their own authorization check using our built in auth model (basic auth example).
 
 ### Possible Solutions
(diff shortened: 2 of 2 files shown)


eriknelson:
This is silly, but can you suffix the file with `.md` for consistency and so it renders?

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
+# Proposal for [Bug 1512042](https://bugzilla.redhat.com/show_bug.cgi?id=1512042)
+
+## Introduction
+The issue is that when a custom namespace (not openshift) is used for holding the images the broker does not currently give access to that namespace when creating the service account for the sandbox. This cause's the resulting transient namespace to get stuck in an error pulling the image. 
+
+# Problem Description
+The problem is that we want to be able to control how the privileges get escalated into that namespace to see/pull the images. This means that we will need a robust solution, that is not a simple bug fix.
+
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the customer namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace.
```
eriknelson: `customer namespace` -> `custom namespace` I think?

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the customer namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. 
+
+To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
+
+There is also a third option that we should document, which is to allow everyone to access your images in that custom namespace. This option will work even with the default because the user will always have access then.
+
+### Open Questions
+1. I think this check should be in the handler. This is an authorization check in my mind and should be treated as such.
+2. The runtime package should be used to determine the check, something like `CheckAccessToImage`.
```
eriknelson: So this will basically be auto passed in the case of "all users can access images"?

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+3. We need to figure out how to handle if the adapter is any other adapter then local_openshift. 
+    1. Add a field on the spec to state the registry/adapter that it is using?
```
eriknelson: Probably useful metadata to have on a spec for the broker anyway.

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+    2. We will also need to track the namespace that it should check.
```
eriknelson: Which namespace is this, the one where the Image resides? This data would come from the local adapter?

Feels a little weird to put this on a `Spec` when it only applies to one specific adapter and only for Openshift.

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+    3. Need to be able to not run this check if the not available.
```
eriknelson: Not sure what this means. Are you thinking in the case of a `k8s` cluster?

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+    4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user, gcr adapter could use their Cloud IAM). Or a calller could want to do their own authorization check using our built in auth model (basic auth example).
```
eriknelson: `calller -> caller`

eriknelson: Seems like the concept is generically useful across adapters. Would be nice to think though how adapter implementers may be able to optionally implement access rights.

docs/proposals/proposal-bug-1512042
```diff
@@ -0,0 +1,80 @@
...
+
+### Possible Solutions
+1. Create new fields on the spec. These fields will keep track of the adapters/registries that added the spec. Add a function in the runtime package that will be called by the handler for authorization. This function will take the spec, the originating user structure, and some undefined structures as a variadic. Example:
+```go
+type RuntimeAuthorization func(apb.Spec, apb.UserInfo, interface{}...)
+
+var RuntimeAuthorization RuntimeAuthorization
+```
+2.  Create new fields on the spec sor that it can track the adapters/registries that added this specific spec. Add a interface Authorization for the runtime package. This will have a single method, Authorize that will take a UserInfo struct. This then can be chained, and the runtime package should manage this as this is supposed to be exposed so everyone can add one. We can helper methods to add providers to the authorization on the fly.
```
eriknelson: `sor -> so`

eriknelson: Think the name of `Authorization` needs to be more specific if it's going to live in the runtime package. `runtime.Authorization` could mean anything.

eriknelson: `can helper -> can have helper`

eriknelson: +1 to the chain idea.

djzager: nit: Add a interface Authorization --> Add an interface

docs/proposals/proposal-bug-1512042
shawn-hurley: Correct, if we are suppose to auto grant then this function would return true.

docs/proposals/proposal-bug-1512042
shawn-hurley: I don't think that this needs to apply only to a single adapter. I think that the amazon ECR and google GCR adapters could use this behavior to add checks to a registry. Also, why would we want this check for docker hub private repos? Right now you can specify whether or not the broker has access to those images with a username and password, but what if I want to stop the user who is calling me from accessing. I think that maybe you are right this needs to be a more general structure for keeping this metadata, but I think it is useful.

docs/proposals/proposal-bug-1512042
shawn-hurley: With the work here: https://trello.com/c/mFYD8lRf I think that the correct thing to do is give them hooks so that they can add their own checks like how we are giving them the ability to create and add their own adapters or add code to pre and post sandbox creation. Is that what you meant or is this not what you were asking?

docs/proposals/proposal-bug-1512042
shawn-hurley: @eriknelson can you explain a little more to me why `runtime.Authorization` does not work. In my mind, we are saying that the only authorization that will be done is by the managed authorization checks at runtime. This means that the handler calls a single function and all the authorization happens in that chain. What other authorizations could be happing and what ideas do you have to make it more specific?

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
+# Proposal for [Bug 1512042](https://bugzilla.redhat.com/show_bug.cgi?id=1512042)
+
+## Introduction
+The issue is that when a custom namespace (not openshift) is used for holding the images the broker does not currently give access to that namespace when creating the service account for the sandbox. This cause's the resulting transient namespace to get stuck in an error pulling the image.
```
jmrodri: Restate the above as:

> When using a custom namespace for holding the images, the broker does not give access to that namespace when creating the service account for the sandbox. This causes the resulting transient namespace to get stuck in an error pulling the image.

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
+# Proposal for [Bug 1512042](https://bugzilla.redhat.com/show_bug.cgi?id=1512042)
+
+## Introduction
+The issue is that when a custom namespace (not openshift) is used for holding the images the broker does not currently give access to that namespace when creating the service account for the sandbox. This cause's the resulting transient namespace to get stuck in an error pulling the image. 
+
+# Problem Description
+The problem is that we want to be able to control how the privileges get escalated into that namespace to see/pull the images. This means that we will need a robust solution, that is not a simple bug fix.
```
jmrodri: Remove: "The problem is that". Change we -> We.

Change that -> this

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the custom namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. There will only be two options for the configuration value `CheckAccessToNamespace` and `AutoGrantAccessToNamespace`.
+
+To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
+
+There is also a third option that we should document, which is to allow everyone to access the images in that custom namespace. This is purelly an RBAC policy setting and will allow all the users to access the images.
```
jmrodri: TYPO: purelly -> purely

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the custom namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. There will only be two options for the configuration value `CheckAccessToNamespace` and `AutoGrantAccessToNamespace`.
+
+To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
+
+There is also a third option that we should document, which is to allow everyone to access the images in that custom namespace. This is purelly an RBAC policy setting and will allow all the users to access the images. 
+
+### Open Questions
+1. I think this check should be in the `pkg/handler/handler.go` file. This is an authorization check in my mind and should be treated as such. This check is only needed when an APB is being launched and should follow the placement of the current role authorization check.
```
jmrodri: My auth stuff has a filter that can do this before the actual handler call, not sure if that's too early or not.

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+2. The runtime package should be used to determine the check, something like `CheckAccessToImage`.
+3. We need to figure out how to handle if the adapter is any other adapter then local_openshift.
```
jmrodri: TYPO: then -> than

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+
+To grant access to sandbox's service account we will grant access to pull images from the custom namespace. 
+
+There is also a third option that we should document, which is to allow everyone to access the images in that custom namespace. This is purelly an RBAC policy setting and will allow all the users to access the images. 
+
+### Open Questions
+1. I think this check should be in the `pkg/handler/handler.go` file. This is an authorization check in my mind and should be treated as such. This check is only needed when an APB is being launched and should follow the placement of the current role authorization check.
+2. The runtime package should be used to determine the check, something like `CheckAccessToImage`.
+3. We need to figure out how to handle if the adapter is any other adapter then local_openshift. 
+    1. Add a field on the spec to state the registry/adapter that it is using?
```
jmrodri: Doesn't the image name have the registry encoded in it? i.e. `dh-` etc?

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+    2. We will also need to track the namespace that it should check.
+    3. Need the ability to turn off the check when certain specs from other adapters are the ones being used.
```
jmrodri: Not quite sure I understand item 3.

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+    4. Need this to be usable by other adapters if they want (ecr adapter could use this to look at IAM access of the originating origin user, gcr adapter could use their Cloud IAM). Or a calller could want to do their own authorization check using our built in auth model (basic auth example).
```
jmrodri: TYPO: calller -> caller.

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
+# Proposal for [Bug 1512042](https://bugzilla.redhat.com/show_bug.cgi?id=1512042)
+
+## Introduction
+The issue is that when a custom namespace (not openshift) is used for holding the images the broker does not currently give access to that namespace when creating the service account for the sandbox. This cause's the resulting transient namespace to get stuck in an error pulling the image. 
+
+# Problem Description
+The problem is that we want to be able to control how the privileges get escalated into that namespace to see/pull the images. This means that we will need a robust solution, that is not a simple bug fix.
+
+## Implementation Details
+The proposed changes are adding a configuration option to the local_registry. This configuration option will be similar to the auto_escalate feature we already have. The option will allow for checking if the originating user has access to the custom namespace to pull images. This will be the default, and you can specify to auto grant access to the namespace. There will only be two options for the configuration value `CheckAccessToNamespace` and `AutoGrantAccessToNamespace`.
```
jmrodri: later you mention other adapters using this. Why would we add a config option only to the `local_registry`?

docs/proposals/proposal-bug-1512042.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Possible Solutions
+1. Create new fields on the spec. These fields will keep track of the adapters/registries that added the spec. Add a function in the runtime package that will be called by the handler for authorization. This function will take the spec, the originating user structure, and some undefined structures as a variadic. Example:
+```go
+type RuntimeAuthorization func(apb.Spec, apb.UserInfo, interface{}...)
+
+var RuntimeAuthorization RuntimeAuthorization
+```
+2.  Create new fields on the spec so that it can track the adapters/registries that added this specific spec. Add an interface Authorization for the runtime package. This will have a single method, Authorize that will take a UserInfo struct. This then can be chained, and the runtime package should manage this as this is supposed to be exposed so everyone can add one. We can have helper methods to add providers to the authorization on the fly.
```
jmrodri: When I ported over the auth stuff, I left out the `Permission` interface because we didn't need it yet.

```
type Permission interface {
   CanAccess(interface{}, SubResource, Access) bool
}
```
Access and SubResource were basically enums. SubResource was the list of objects that we knew about: entitlements, consumers, subscriptions, etc. In broker land that would be namespace, registry, etc. 

Access was NONE, READONLY, CREATE, ALL.

We had a Verify filter that would check to see if the entity had access to a resource.

shawn-hurley:
Closing this PR. This solution is complex and does not need to be done to complete this bug.

PR #596, opened 2017-12-14
Execute into a pod with API for runtime V1

shawn-hurley:
moves runtime v1 credential extraction to the remote API
 - update to vendor so that the remote api can be used 
 - update ext from pod to use the remote execution


shawn-hurley:
[Integrate remote execution for v1 (backwards compat) credential extraction.](https://trello.com/c/HjGNBAtG/607-integrate-remote-execution-for-v1-backwards-compat-credential-extraction)

pkg/apb/ext_creds.go
```diff
@@ -43,54 +56,85 @@ func ExtractCredentials(podname string, ns string, runtime int) (*ExtractedCrede
...
-			log.Warning("[%s] Retry attempt %d: Failed to exec into the container", podname, r)
-		} else if err.Error() == credsNotAvailable.Error() {
-			log.Info(string(output))
-			log.Warning("[%s] Retry attempt %d: Bind credentials not available yet", podname, r)
-		} else {
-			log.Info(string(output))
-			log.Warning("[%s] Retry attempt %d: Failed to exec into the container", podname, r)
+		default:
+			log.Infof("command output: %v - err: %v", stdoutBuffer.String(), stderrBuffer.String())
+			log.Warningf("retry attempt: %v pod: %v in namespace: %v failed to exec into the container", r, podname, namespace)
```
eriknelson: I noticed this got bumped up to a `Warning` level a while back in the old code with previous iterations of the credential extractions. I don't think this should be of a warning level since the "tried to get credentials but they weren't ready yet" branch of execution is expected to get hit a few times during a normal run of an APB.

If we agree this isn't a warning, probably drop v2 back down to `Info` as well, either in this PR or a small follow up.

pkg/apb/ext_creds.go
```diff
@@ -17,16 +17,29 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	// GatherCredentialsCommand - Command used when execing for bind credentials
+	// moving this constant here because eventuall Extrating creds will
```
djzager: nit: eventually

djzager: Thank you for putting the reason for the move here.

shawn-hurley:
@djzager @eriknelson I could not find the `Warningf` log in watch pod, am I missing where the log would be now?

eriknelson:
@shawn-hurley it's possible it got removed, I just remember it being I think in the old `ext_creds` or something. Wouldn't worry about it now.

pkg/apb/ext_creds.go
```diff
@@ -43,54 +55,82 @@ func ExtractCredentials(podname string, ns string, runtime int) (*ExtractedCrede
...
+	clientConfig := k8scli.ClientConfig
+	clientConfig.GroupVersion = &v1.SchemeGroupVersion
+	clientConfig.NegotiatedSerializer =
+		serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
 
-		// cannot exec container, pod is done
-		podFailed := strings.Contains(string(output), "current phase is Failed")
-		podCompleted := strings.Contains(string(output), "current phase is Succeeded") ||
-			strings.Contains(string(output), "cannot exec into a container in a completed pod")
+	// NOTE: kubectl exec simply sets the API path to /api when where is no
```
jmrodri: when where?

PR #599, opened 2017-12-15
allows parameter types to be case-insensitive

### LINKED ISSUE #592: Type fields in APB specs are case-sensitive
`apb push -o` an APB spec with a type of `String` rather than `string` will cause the catalog endpoint to return empty spec objects, since they end up erroring on this case statement. Of potential importance, this is the local openshift adapter:

https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/util.go#L200

This causes `apb list` to display an empty entry (it will look like a blank line).

Interestingly, an additional bootstrap will resolve the problem...somehow. `apb bootstrap && apb list` will display the expected entry, so something must be lowercasing the type.

mhrivnak:
The error message when failing to parse a spec is improved to include the name
of the spec.

fixes #592

### DIFF
--- a/pkg/broker/util.go
+++ b/pkg/broker/util.go
@@ -19,6 +19,7 @@ package broker
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"encoding/json"
 
@@ -196,7 +197,7 @@ func createUIFormItem(pd apb.ParameterDescriptor, paramIndex int) (interface{}, 
 
 // getType transforms an apb parameter type to a JSON Schema type
 func getType(paramType string) (schema.PrimitiveTypes, error) {
-	switch paramType {
+	switch strings.ToLower(paramType) {
 	case "string", "enum":
 		return []schema.PrimitiveType{schema.StringType}, nil
 	case "int":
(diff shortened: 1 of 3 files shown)


pkg/broker/util.go
```diff
@@ -196,7 +197,7 @@ func createUIFormItem(pd apb.ParameterDescriptor, paramIndex int) (interface{},
 
 // getType transforms an apb parameter type to a JSON Schema type
 func getType(paramType string) (schema.PrimitiveTypes, error) {
-	switch paramType {
+	switch strings.ToLower(paramType) {
```
shawn-hurley: Can we add a test case for this? `util_test.go:371` should be the test cases to add to?

shawn-hurley:
@mhrivnak I am re-running the Gate test after the apb base fix that we merged yesterday.

pkg/broker/util.go
mhrivnak: Do you have something in mind that's more than what I already added there?

shawn-hurley:
```go
...
{"number", schema.NumberType},
{"NUMBER", schema.NumberType},
{"nil", schema.NullType},
...
```

Add that to line 385ish in `pkg/broker/util_test.go`

Issue #590, opened 2017-12-13
We should move to using the openshift client-go


shawn-hurley:
# Feature: Use Openshift Client Go and Openshift API

The [openshift client-go] and [api] should be used instead of our [openshift client] that we have been using. This will require a change to our copied modules [here]() as well, but should then easily changed. 

This will give us access to all of the openshift APIs without building our own rest clients now.

jmrodri:
Fixed by PR https://github.com/automationbroker/bundle-lib/pull/105

PR #600, opened 2017-12-19
Openshift Multi-tenant Sandbox Hooks

shawn-hurley:
Implementation of #572 
 - Adding post sandbox create/destroy hooks
 - Adding actual join/isolate network hook implementation
 - Adding cluster specific `shouldJoinNetworks` method to retrieve the hooks to be used. 
- Using [openshift api](https://github.com/openshift/api) to get structures for openshift calls partial fix for #590


shawn-hurley:
[(5) Implement network isolation proposal for multi-tenet environment.](https://trello.com/c/LoNEWiCp/611-5-implement-network-isolation-proposal-for-multi-tenet-environment)

shawn-hurley:
This PR is related to https://github.com/openshift/openshift-ansible/pull/6536.

They do not block each other, but they are needed to work together.

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
...
+	netNamespace := &networkoapi.NetNamespace{}
+	err := o.networkClient.Get().Resource("netnamespaces").Name(nsName).Do().Into(netNamespace)
+	if err != nil {
+		return nil, err
+	}
+	return netNamespace, nil
+}
+
+// JoinNamespacesNetworks - Will take the net namespace to be added to a network,
+// and the namespace ID of that netwok.
```
shawn-hurley: make this more clear that the namespace id is the network that we are joining to.

pkg/runtime/openshift.go
```diff
@@ -1,5 +1,134 @@
...
+	ocli, err := clients.Openshift()
+	if err != nil {
+		log.Errorf("unable to get openshift client - %v", err)
+		// Defaulting if anything goes wrong to not join the networks.
+		return false, nil, nil
+	}
+	pluginName, err := ocli.GetClusterNetworkPlugin()
+	log.Debugf("plugin for the network - %v", pluginName)
+	if err != nil {
+		// Plugins could not be defined oc cluster up or other use cases.
```
shawn-hurley: make this statement more clear

pkg/runtime/runtime.go
```diff
@@ -172,6 +186,16 @@ func (p provider) CreateSandbox(podName string,
 	}
 
 	log.Info("Successfully created apb sandbox: [ %s ], with %s permissions in namespace %s", podName, apbRole, namespace)
+	log.Info("Running post create sandbox fuctions if defined.")
+	for i, f := range p.postSandboxCreate {
+		log.Debug("Running function: %v", i)
```
shawn-hurley: Debug -> Debugf

Also, make this more useful.

pkg/runtime/sandbox_hooks.go
```diff
@@ -0,0 +1,36 @@
...
+// after itself.
+// Parameters in order of adding to the function.
+// string - pod name is also the svc accounts name.
+// string - namespace of the transient namespace.
+// list of strings - target namespaces.
+// string - abp role.
+// return error.
+type PostSandboxCreate func(string, string, []string, string) error
+
+// AddPostCreateSandbox - Adds a post create sandbox to the runtime.
```
shawn-hurley: Adds a post create sandbox to the runtime -> Adds a post create sandbox function to the runtime.

pkg/runtime/sandbox_hooks.go
```diff
@@ -0,0 +1,36 @@
...
+// Once the sandbox is created all of the functions that have been defined here
+// will be executed.
+func (p *provider) AddPostCreateSandbox(f PostSandboxCreate) {
+	p.postSandboxCreate = append(p.postSandboxCreate, f)
+}
+
+// PostSandboxDestroy - The post sand box destroy function will be called
+// after the sandbox is destory. This could mean the namespace is kept around
```
shawn-hurley: destroy -> destroyed

pkg/runtime/sandbox_hooks.go
```diff
@@ -0,0 +1,36 @@
...
+// if the error and configuration conditions are met. This function should not
```
shawn-hurley: error/APB errors ....

jwmatthews:
@smarterclayton had some concerns with this PR, let's hold off on merging until we learn more about the concerns and can address.

jwmatthews:
Adding some additional background to help others understand the context of this PR.

Background of the issue:
  With APBs there will be 2 namespaces at play.
  "target" namespace, which is the namespace the deployed services will run in
  "transient" namespace, the temporary namespace the broker creates for the APB to run
      only the APB code runs in here, then the namespace is destroyed.


A typical workflow is: 
 1. 'target' namespace exists 
 2. User runs an action that triggers an APB action
 3. Broker creates a new namespace, the 'transient' namespace
 4. Broker runs the APB in the 'transient' namespace
   a.. The actual ansible logic for the APB runs in the 'transient' namespace
        OCP Resources are manipulated in the 'target' namespace
 5. Broker destroys the 'transient' namespace

In a multi-tenant setup, the namespace networks are isolated.  This means the APB logic is unable to talk to the deployed services.
 
The goal of the PR referenced is to allow the transient namespace to communicate to the target namespace.

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
...
+	netNamespace := &networkoapi.NetNamespace{}
+	err := o.networkClient.Get().Resource("netnamespaces").Name(nsName).Do().Into(netNamespace)
+	if err != nil {
+		return nil, err
+	}
+	return netNamespace, nil
+}
+
+// JoinNamespacesNetworks - Will take the net namespace to be added to a network,
+// and the namespace ID of the netwok that is being added to.
```
knobunc: netwok -> network

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
...
+	result := &networkoapi.NetNamespace{}
+	err := o.networkClient.Put().Resource("netnamespaces").Name(netns.Name).Body(netns).Do().Into(result)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+// IsolateNamespacesNetworks - Will take the net namespace to be added to a network,
+// and the namespace ID of the netwok that is being added to..
```
knobunc: netwok -> network

danwinship:
Summing up the bluejeans conversation:
- The only obvious alternative approach would be to have the APB deploy a helper pod in the target namespace and then use "oc exec" to run commands through that. But that would probably require lots of additional complexity in every APB that needed these features, whereas the proposal here puts all of the extra work in the infrastructure and makes things easier for the individual APBs.
- If the cluster is using the multitenant plugin, then joining the networks together is the right answer. I would have suggested just using `oc adm pod-network ...` but using the API directly works as well. Also, there's no need to re-isolate them afterward since "isolating" just means giving the transient NetNamespace a unique NetID again, but there's no point in doing that since you're about to delete it anyway.
- If the cluster is using the `redhat/openshift-ovs-networkpolicy` plugin, or basically any third-party plugin, then you can't "join networks", and you'll instead have to create a NetworkPolicy in the target namespace, allowing access from the transient namespace. (Which will require putting some sort of label on the transient Namespace object first so that you can match it with a LabelSelector from the NetworkPolicy.) (If the cluster is using the multitenant plugin, you can still do this anyway; it will just be a no-op.)
- Looking at the `"default"` ClusterNetwork record is the best currently-available way to figure out what network plugin is in use, but it's only guaranteed to be right if one of the openshift plugins is being used; otherwise, it's possible that it will contain stale data. (Our "how to change network plugins" documentation tells people to delete the ClusterNetwork object when switching to a third-party plugin, but they might not.). So the safe/right way to set up cross-namespace network access is:
  - always create a NetworkPolicy
  - if the default ClusterNetwork claims that multitenant is in use, then also do the join-projects hack, but it's possible that the ClusterNetwork record is wrong, so don't actually error out if the master ignores the request to join the networks. (But maybe warn about it?)

pkg/clients/openshift.go
```diff
@@ -41,13 +42,20 @@ import (
 	rbac "k8s.io/kubernetes/pkg/apis/rbac"
 )
 
+const (
+	// ChangePodNetworkAnnotation - Annotation used for changing a pod
+	// network, used to join networks together.
+	ChangePodNetworkAnnotation = "pod.network.openshift.io/multitenant.change-network"
```
danwinship: ah, right... I was going to say "that constant should be available in `networkoapi`", but it's not, because we had decided that this is "internal API", not public API. So maybe that's an argument that you should use `oc adm pod-network` instead? (But the API is still guaranteed to keep existing, since we have to preserve compatibility between old/new clients/servers, so it's safe for you to depend on it if you'd rather keep the code this way.)

shawn-hurley:
@danwinship 1 clarification

> Also, there's no need to re-isolate them afterward since "isolating" just means giving the transient NetNamespace a unique NetID again, but there's no point in doing that since you're about to delete it anyway.

we could potentially keep the transient namespace alive so that a cluster admin could diagnose issues if an APB fails. In that case, we should re-isolate the network correct?

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
 	}
 	return imageList, nil
 }
+
+// GetClusterNetworkPlugin - Get cluster network
+func (o OpenshiftClient) GetClusterNetworkPlugin() (string, error) {
+	net := &networkoapi.ClusterNetwork{}
+	err := o.networkClient.Get().Resource("clusternetworks").Name("default").Do().Into(net)
```
danwinship: `Name(networkoapi.ClusterNetworkDefault)`

pkg/runtime/openshift.go
```diff
@@ -1,5 +1,134 @@
...
+
+	//  wait for some time, to determine if the change was applied correctly.
+	backoff := wait.Backoff{
+		Steps:    15,
+		Duration: 500 * time.Millisecond,
+		Factor:   1.1,
+	}
+	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+		return didAnnotationUpdate("join", netns.NetName)
+	})
```
danwinship: as mentioned above, this could fail if the ClusterNetwork object is stale and mistakenly claims that they're using multitenant when really they're using a third-party plugin now

pkg/runtime/openshift.go
```diff
@@ -1,5 +1,134 @@
...
+		// or a pure k8s cluster. Therefore making this a notice.
+		log.Noticef("unable to retrieve the network plugin, defaulting to not joining networks - %v", err)
+		// Defaulting to not join the networks.
+		return false, nil, nil
+	}
+
+	// Case insensitive check here because want to prepare if things change.
+	if strings.ToLower(pluginName) == "redhat/openshift-ovs-multitenant" {
+		log.Debugf("stating that the pluginname is multitenant - %v", pluginName)
+		return true, addPodNetworks, isolatePodNetworks
```
danwinship: as mentioned above, there's really no need to re-isolate the transient network on cleanup

danwinship:
> we could potentially keep the transient namespace alive so that a cluster admin could diagnose issues if an APB fails. In that case, we should re-isolate the network correct?

You wouldn't want to keep things in the same state as when the APB ran? I guess there are arguments both ways.

Hm... I was about to say "leaving the networks joined won't hurt anything", but I just realized that's not actually the case: joining networks breaks multicast and EgressNetworkPolicy.

For multicast you can make things work by having the setup process set the "multicast-enabled" plugin on the transient NetNamespace if it's set on the target NetNamespace (https://docs.openshift.org/latest/admin_guide/managing_networking.html#admin-guide-networking-multicast). (Ideally you should *not* just set it unconditionally; the nodes will log warnings if you join two namespaces and one of them has the multicast annotation but the other doesn't.)

But for EgressNetworkPolicy, there's no way to make things work; if the target namespace has any EgressNetworkPolicies defined, then joining another namespace to it will break all cluster-external traffic from that namespace.

EgressNetworkPolicy isn't that widely used though, and we're thinking about deprecating it, so maybe you can just punt on that and say that APBs can't be deployed into a namespace that has any EgressNetworkPolicies defined in it, if the cluster is using the multitenant plugin.

shawn-hurley:
> You wouldn't want to keep things in the same state as when the APB ran? I guess there are arguments both ways.

We are currently removing the role bindings and such even if we are keeping the namespace. I figured that we would want to do the same and if the cluster admin would want to re-add the networks together then they could. @jwmatthews Do you have any preference here. I don't think it hurts to leave them attached, just thought the right thing was to remove our failed run from the cluster as much as possible? 

> EgressNetworkPolicy isn't that widely used though, and we're thinking about deprecating it, so maybe you can just punt on that and say that APBs can't be deployed into a namespace that has any EgressNetworkPolicies defined in it, if the cluster is using the multitenant plugin.

Sorry, I have not read about this policy yet, do you think that a large documentation warning on this would be sufficient? Should the broker be able to look up some value and error if a setting/condition is met?

pkg/clients/openshift.go
shawn-hurley: I think that we are attempting to remove shelling out in our code, therefore I think on a whole we would prefer to keep using the API endpoints. 

@jmrodri @eriknelson can you guys weigh in as far as broker code base is concerned? It sounds like the question is if we have to duplicate some constants and some logic around checking if an annotation has been updated vs shelling out to the oc command.

@danwinship Is that a fair characterization of the concern here? Do you see more code duplication then the above mentioned?

enj:
@danwinship do you see any issues in this scenario:

1. Transient namespace and target namespace have joined networks
2. ASB is running an AWS APB in the transient namespace, which causes it to copy AWS secrets into the transient namespace, and then use them on the network to provision AWS resources
3. The user who requested this from the ASB has a chance to poke at the joined network in an attempt to steal AWS secrets

So my guess is that if the APB does the right thing and verifies all TLS certs, it is not vulnerable to a MITM attack from the user.  I am just not sure of the scope of attacks that are possible here, and how the ASB should mitigate them.

cc @simo5 @mrogers950 @liggitt in case they have comments.

danwinship:
> Sorry, I have not read about [EgressNetworkPolicy] yet, do you think that a large documentation warning on this would be sufficient? Should the broker be able to look up some value and error if a setting/condition is met?

It can List() the networkapi.EgressNetworkPolicy objects in the target namespace, and if there are any, it should probably error out (before having tried to join the networks). Note that this (and the multicast concern) only applies to multitenant; there's no problem with the networkpolicy plugin or with third-party plugins.

> @danwinship do you see any issues in this scenario:

If the transient APB namespace has access to things that it doesn't want the target namespace to get, then yes, it will need to make sure that it's protecting those things from other pods that are able to talk to it over the network, rather than assuming that everyone who can talk to it is trusted.

Related: if you deploy an APB into the default namespace, then its transient namespace will get joined to "default", which means that *all* pods in all namespaces will be able to connect to the pods in the transient namespace. So if that's something that ever happens (deploying into "default" or any other global namespace) then APBs can *never* assume that anyone connecting to them is trusted.

(And again, these problems apply only to multitenant, with its unavoidably-symmetric permissions; in a NetworkPolicy-based network plugin, you can give the transient namespace permission to connect to the target namespace without giving anyone permission to connect to the transient namespace.)

shawn-hurley:
> If the transient APB namespace has access to things that it doesn't want the target namespace to get, then yes, it will need to make sure that it's protecting those things from other pods that are able to talk to it over the network, rather than assuming that everyone who can talk to it is trusted.

The APB does not have a service or even open any ports for anything to communicate to it. This should eliminate the above issue correct? 

Here is the pod definition to make sure that is the correct assumption: 
https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/executor.go#L107

shawn-hurley:
@enj @danwinship @knobunc Can you guys double check this PR. I believe that I have updated it to match the comments and process that was agreed upon. Thanks!

danwinship:
- You removed the "re-isolate" part at the end, but I think you should put it back. In particular, joining the temp namespace to the target namespace could cause disruption of multicast and EgressNetworkPolicy, so you want that to be undisrupted when you finish installing.
- addPodNetworks() still always returns an error if the join fails, but there's the problem that the user might be using a third-party plugin and ClusterNetwork has stale info, in which case joining would time out (because the controller that watches for the "join" annotation wouldn't be running). Maybe the right fix here is just to document it though: "if you're using a third-party network plugin, and get such-and-such error message, then the fix is to run `oc delete clusternetwork default` to delete the stale config object that claims you're running the multitenant plugin".
- The new NetworkPolicy-related code looks correct. (I assume that that value you're sticking in `apb-pod-name` is unique-ish?)

shawn-hurley:
> addPodNetworks() still always returns an error if the join fails, but there's the problem that the user might be using a third-party plugin and ClusterNetwork has stale info, in which case joining would time out (because the controller that watches for the "join" annotation wouldn't be running). Maybe the right fix here is just to document it though: "if you're using a third-party network plugin, and get such-and-such error message, then the fix is to run oc delete clusternetwork default to delete the stale config object that claims you're running the multitenant plugin".

We are logging the error but the way in which the post hook works is it will not cause the rest of the APB execution to quit. I think that this one we do not need to change and can add that documentation regarding the message. 

> The new NetworkPolicy-related code looks correct. (I assume that that value you're sticking in apb-pod-name is unique-ish?)

It is a UUID so it does have some uniqueness guarantee, as far as I understand?

pkg/runtime/runtime.go
```diff
@@ -171,7 +186,44 @@ func (p provider) CreateSandbox(podName string,
 		}
 	}
 
+	// Must create a Network policy to allow for comunication from the APB pod to the target namespace.
+	networkPolicy := &networkingv1.NetworkPolicy{
```
liggitt: is there a "deny all ingress" network policy created in the apb pod namespace that will prevent the user from initiating communication in the other direction?

shawn-hurley: Correct me if I am wrong, but if the network policy does not exist to allow the traffic the base state of the plugin `ovs-networkpolicy` is to disallow traffic.

If you are using `ovs-multitenant` plugin I don't think that you can.

@danwinship are these statements correct?

danwinship:
> We are logging the error but the way in which the post hook works is it will not cause the rest of the APB execution to quit.

OK

> It is a UUID so it does have some uniqueness guarantee, as far as I understand?

Yeah, that's good. Not predictable, and an attacker can't trick you into using a value of their choice.

pkg/runtime/runtime.go
danwinship: The base state of the networkpolicy plugin is to allow all traffic. But you said before there's nothing listening to any ports in the temp namespace, so this shouldn't matter?

But if you want to be paranoid, you could add a policy:

    kind: NetworkPolicy
    apiVersion: extensions/v1beta1
    metadata:
      name: deny-all
    spec:
      podSelector:
    ingress: []

shawn-hurley: I think that if we ever open a port for the APB  to listen on then we would want to control the network with `NetworkPolicy` as @danwinship said. I don't think it is a concern as there is nothing exposed to the network in the apb pod namespace. 

@liggitt and I think @enj you had the same concern, does the above sound acceptable to you?

liggitt: yes

jmontleon:
I tested this (albeit before changes to support the tech preview network-policy plugin) and I was able to update an APB in a multi-tenant environment.

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
...
+	if netns.Annotations == nil {
+		netns.Annotations = make(map[string]string)
+	}
+	netns.Annotations[ChangePodNetworkAnnotation] = fmt.Sprintf("%s:%s", "join", targetNS)
+	result := &networkoapi.NetNamespace{}
+	err := o.networkClient.Put().Resource("netnamespaces").Name(netns.Name).Body(netns).Do().Into(result)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
```
rthallisey: Why return ```result``` if it's not used by the caller?

pkg/clients/openshift.go
```diff
@@ -404,3 +420,57 @@ func (o OpenshiftClient) ListRegistryImages() (images []string, err error) {
...
+	if netns.Annotations == nil {
+		netns.Annotations = make(map[string]string)
+	}
+	netns.Annotations[ChangePodNetworkAnnotation] = fmt.Sprintf("%s:%s", "isolate", targetNS)
+	result := &networkoapi.NetNamespace{}
+	err := o.networkClient.Put().Resource("netnamespaces").Name(netns.Name).Body(netns).Do().Into(result)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
```
rthallisey: same as above.

pkg/clients/openshift.go
shawn-hurley: no good reason. eventually, you may need it and you can ignore it when you call the method. 

do you have strong feelings on this one?

rthallisey: It doesn't matter to me if you think it will be used eventually I'm fine with it.

I would double check though that ```result``` isn't going to be the same as the```netns``` param.  I would have to do some testing to verify, but the I don't think ```NetName``` and ```NetId``` will change.

pkg/runtime/kubernetes.go
```diff
@@ -3,3 +3,7 @@ package runtime
 func (k kubernetes) getRuntime() string {
 	return "kubernetes"
 }
+
+func (k kubernetes) shouldJoinNetworks() (bool, PostSandboxCreate, PostSandboxDestroy) {
+	return false, nil, nil
```
eriknelson: Confirming for my own edification -- this is strictly an Openshift feature?

shawn-hurley: The openshift plugin for the multitenant environment is only for openshift. 

The network policy stuff is for both k8s and openshift so that is apart of the actual create sandbox and destroy sandbox methods.

pkg/runtime/sandbox_hooks.go
```diff
@@ -0,0 +1,36 @@
...
+
+// AddPostCreateSandbox - Adds a post create sandbox function to the runtime.
+// Once the sandbox is created all of the functions that have been added here
+// will be executed.
+func (p *provider) AddPostCreateSandbox(f PostSandboxCreate) {
+	p.postSandboxCreate = append(p.postSandboxCreate, f)
+}
+
+// PostSandboxDestroy - The post sand box destroy function will be called
+// after the sandbox is destoryed. This could mean the namespace is kept around
```
eriknelson: destoryed -> destroyed

pkg/runtime/kubernetes.go
eriknelson: Thanks.

PR #601, opened 2017-12-19
Adding local openshift adapter

matzew:
The PR adds the local openshift adapter to the template.

Problem is, with the `master` version of the `templates/deploy-ansible-service-broker.template.yaml`, and the `scripts/run_latest_build.sh` files, it is not possible to perform an `apb push` (with latest apb-tools).

The call succeeds, but the apb is not added.

On IRC @shawn-hurley told me that it's because I dont have the local openshift adapter. 

So I am adding it here

 - adding local openshift adapter to the template

depends-on <PR>

nope

fixes <issue_number>


shawn-hurley:
Thanks for the PR!

PR #603, opened 2017-12-27
Remove the different job msg types to avoid duplication of code

maleck13:

There were 3 different JobMsg types being used however they were all essentially the same barring one field and all consumers of the type were using the interface: WorkMsg 

 - remove the provisionmsg, deprovisionmsg and updatemsg type and replace them with a single jobmsg type

### DIFF
--- a/pkg/broker/deprovision_job.go
+++ b/pkg/broker/deprovision_job.go
@@ -17,8 +17,6 @@
 package broker
 
 import (
-	"encoding/json"
-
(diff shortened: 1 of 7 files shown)


maleck13:
While looking into https://github.com/openshift/ansible-service-broker/issues/475 I noticed the different job msg types were all the same and that the specific type was not being used for anything so seemed to make sense to merge them into a single type

PR #607, opened 2018-01-02
Bug 1512042 - Allowing error messages to make it from apb to user.

shawn-hurley:
* Make it so that job state holds errors from apb package
* Make it so de/provision/update/_job can get the error and handle correctly.
* make sure that the broker/handler handles the error and returns
* Warn when about permissions for specs that may not be pullable.

Fixes bug 1512042 by adding more explicit warning logs and handing errors back to the user.

Bug 1512042

### DIFF
--- a/pkg/apb/watch_pod.go
+++ b/pkg/apb/watch_pod.go
@@ -52,7 +52,7 @@ func watchPod(podName string, namespace string) error {
 
 		switch podStatus.Phase {
 		case apiv1.PodFailed:
-			if errorPullingImage(podStatus.Conditions) {
+			if errorPullingImage(podStatus.ContainerStatuses) {
 				return ErrorPodPullErr
 			}
 			return fmt.Errorf("Pod [ %s ] failed - %v", podName, podStatus.Message)
@@ -69,11 +69,22 @@ func watchPod(podName string, namespace string) error {
 	return fmt.Errorf("Timed out while watching pod %s for completion", podName)
 }
 
-func errorPullingImage(conds []apiv1.PodCondition) bool {
-	for _, cond := range conds {
-		if cond.Reason == "ErrImgPull" {
-			return true
-		}
+func errorPullingImage(conds []apiv1.ContainerStatus) bool {
+	if len(conds) < 1 {
+		log.Warningf("unable to get container status for APB pod")
+		return false
 	}
+	// We should expect only a single container for our APB pod.
+	// If this assumption changes then we may need to update this code.
+	// Basis for the image strings is here:
+	// https://github.com/kubernetes/kubernetes/blob/886e04f1fffbb04faf8a9f9ee141143b2684ae68/pkg/kubelet/images/types.go#L27
+	status := conds[0].State.Waiting
+
+	if status.Reason == "ErrImagePull" {
+		return true
+	} else if status.Reason == "ImagePullBackOff" {
+		return true
+	}
+
 	return false
 }
--- a/pkg/apb/watch_pod.go
+++ b/pkg/apb/watch_pod.go
@@ -25,6 +25,11 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 )
 
+var (
+	// ErrorPodPullErr - Error indicating we could not pull the image.
+	ErrorPodPullErr = fmt.Errorf("Unable to pull APB image from it's registry. Please contact your cluster admin")
+)
+
 func watchPod(podName string, namespace string) error {
 	log.Debugf(
 		"Watching pod [ %s ] in namespace [ %s ] for completion",
@@ -47,6 +52,9 @@ func watchPod(podName string, namespace string) error {
 
 		switch podStatus.Phase {
 		case apiv1.PodFailed:
+			if errorPullingImage(podStatus.Conditions) {
+				return ErrorPodPullErr
+			}
 			return fmt.Errorf("Pod [ %s ] failed - %v", podName, podStatus.Message)
 		case apiv1.PodSucceeded:
 			log.Debugf("Pod [ %s ] completed", podName)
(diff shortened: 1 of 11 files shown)


pkg/apb/watch_pod.go
```diff
@@ -47,6 +52,9 @@ func watchPod(podName string, namespace string) error {
 
 		switch podStatus.Phase {
 		case apiv1.PodFailed:
+			if errorPullingImage(podStatus.Conditions) {
```
djzager: :+1:

pkg/broker/deprovision_job.go
```diff
@@ -55,8 +55,26 @@ func (p *DeprovisionJob) Run(token string, msgBuffer chan<- JobMsg) {
 	if err != nil {
 		log.Error("broker::Deprovision error occurred.")
 		log.Errorf("%s", err.Error())
-		msgBuffer <- JobMsg{InstanceUUID: p.serviceInstance.ID.String(), PodName: podName,
-			JobToken: token, SpecID: p.serviceInstance.Spec.ID, Error: err.Error()}
+		// Because we know the error we should return that error.
```
djzager: I'm not certain I understand why we are returning the specific error message only if `apb.ErrorPodPullErr`. What's the reason for adding this logic?

pkg/registries/adapters/local_openshift_adapter.go
```diff
@@ -114,6 +114,13 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*apb.Spec, err
 			namespace = nsList[1]
 		}
 		for _, ns := range r.Config.Namespaces {
+			// logging to warn users about the potential bug if
+			// the svc-acct does not have access to the namespace.
+			if ns != "openshift" {
+				r.Log.Warningf("You may not be able to load provision images from the namespace: %v.\n"+
```
djzager: :+1:

pkg/apb/watch_pod.go
```diff
@@ -60,3 +68,12 @@ func watchPod(podName string, namespace string) error {
 
 	return fmt.Errorf("Timed out while watching pod %s for completion", podName)
 }
+
+func errorPullingImage(conds []apiv1.PodCondition) bool {
+	for _, cond := range conds {
+		if cond.Reason == "ErrImgPull" {
```
rthallisey: I didn't find "ErrImgPull" anywhere in vendor, only "ErrImagePull". Is "ErrImgPull" the error we want or is it coming from something we don't have vendored? https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L33

pkg/broker/deprovision_job.go
shawn-hurley: Want to give a specific error about not being able to get the image.

If some other error occurs, I don't want to display that message to the user, that is why I want to do this check here.

pkg/apb/watch_pod.go
shawn-hurley: We are using the pod's [status](https://godoc.org/k8s.io/api/core/v1#PodStatus) field. The [conditions](https://godoc.org/k8s.io/api/core/v1#PodStatus) slice is where I am looking for this value. It appears that the reason is a string and this is set to ErrImgPull which you can see when you get the pod w/ yaml when it can not pull the image. 

Is there a better way to get this? I would prefer a better way.

rthallisey: I think we can do something like https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1114-L1147

shawn-hurley: I will re-do this section based on that. It looks much more correct!

pkg/broker/deprovision_job.go
djzager: Thinking more about this. Why don't use just log the error and return a generic error message in `watch_pod` like:

```
log.Errorf("Pod [ %s ] failed - %v", podName, podStatus.Message)
return fmt.Error("Error occurred during APB execution. Please contact administrator if it presists.")
```

pkg/apb/watch_pod.go
```diff
@@ -60,3 +68,23 @@ func watchPod(podName string, namespace string) error {
...
+		log.Warningf("unable to get container status for APB pod")
+		return false
+	}
+	// We should expect only a single container for our APB pod.
+	// If this assumption changes then we may need to update this code.
+	// Basis for the image strings is here:
+	// https://github.com/kubernetes/kubernetes/blob/886e04f1fffbb04faf8a9f9ee141143b2684ae68/pkg/kubelet/images/types.go#L27
+	status := conds[0].State.Waiting
+
+	if status.Reason == "ErrImagePull" {
```
rthallisey: Might as well use ```images.ErrImagePull.Error()``` and ```images.ErrImagePullBackOff.Error()``` since it's where the string comes from.

pkg/broker/deprovision_job.go
shawn-hurley: Because it is not just `watch_pod` that we are worried about, it is the error that apb.(provision/update/deprovision) returns that we are dealing with here.

pkg/apb/watch_pod.go
shawn-hurley: I would prefer not to use something that is not in the kubernetes client-go or kubernetes api. 

The images errors that are referenced in the core kuberentes package, I think that eventually, we could remove this dependency.

rthallisey: wfm

pkg/broker/deprovision_job.go
djzager: Let's hope that is the biggest oversight I have this year.

PR #609, opened 2018-01-04
Bug 1526887 - Handle case when whitelist/blacklist set to ""

shawn-hurley:
Handles case when the yaml looks like
```yaml
registry:
...
    white_list:
       - ""
....
```
 - Test case for the use case
 - If block to ignore regex strings that are zero value or `""`

Bug 1526887

### DIFF
--- a/pkg/registries/filter.go
+++ b/pkg/registries/filter.go
@@ -17,7 +17,6 @@
 package registries
 
 import (
-	"fmt"
 	"regexp"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


pkg/registries/filter.go
```diff
@@ -62,6 +61,10 @@ func compileRegexp(regexStrs []string) ([]*regexp.Regexp, []failedRegexp) {
 	failedRegexps := []failedRegexp{}
 
 	for _, str := range regexStrs {
+		if str == "" {
+			log.Debugf("ignoring empty string that was set as regex string")
```
rthallisey: How about: "Ignoring empty whitelist or blacklist regex."

PR #615, opened 2018-01-05
Bug 1472226 - Add additional field validations for JSON Schema.

cfchase:
Updated apb spec to allow for more validations from JSON Schema. 
https://datatracker.ietf.org/doc/draft-handrews-json-schema-validation/

Validations follow JSON schema keywords, and uses snake_case instead of camelCase for consistency when writing APBs

Added apb_field (JSON schema field):
- max_length (maxLength)
- min_length (minLength)
- multiple_of (multipleOf)
- minimum (minimum)
- maximum (maximum)
- exclusive_minimum (exclusiveMinimum)
- exclusive_maximum (exclusiveMaximum)

Allow maxlength to be used for backwards compatibility as DeprecatedMaxlength.
Allowed `type: integer` since it's used in JSON schema

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -32,6 +32,9 @@ type Parameters map[string]interface{}
 // SpecManifest - Spec ID to Spec manifest
 type SpecManifest map[string]*Spec
 
+// NilableNumber - Number that could be nil (e.g. when omitted from json/yaml)
+type NilableNumber float64
+
 // ParameterDescriptor - a parameter to be used by the service catalog to get data.
 type ParameterDescriptor struct {
 	Name        string      `json:"name"`
@@ -47,11 +50,11 @@ type ParameterDescriptor struct {
 	Pattern             string `json:"pattern,omitempty"`
 
 	// number validators
-	MultipleOf       float64     `json:"multipleOf,omitempty" yaml:"multiple_of,omitempty"`
-	Maximum          interface{} `json:"maximum,omitempty"`
-	ExclusiveMaximum interface{} `json:"exclusiveMaximum,omitempty" yaml:"exclusive_maximum,omitempty"`
-	Minimum          interface{} `json:"minimum,omitempty"`
-	ExclusiveMinimum interface{} `json:"exclusiveMinimum,omitempty" yaml:"exclusive_minimum,omitempty"`
+	MultipleOf       float64        `json:"multipleOf,omitempty" yaml:"multiple_of,omitempty"`
+	Maximum          *NilableNumber `json:"maximum,omitempty"`
+	ExclusiveMaximum *NilableNumber `json:"exclusiveMaximum,omitempty" yaml:"exclusive_maximum,omitempty"`
+	Minimum          *NilableNumber `json:"minimum,omitempty"`
+	ExclusiveMinimum *NilableNumber `json:"exclusiveMinimum,omitempty" yaml:"exclusive_minimum,omitempty"`
 
 	Enum         []string `json:"enum,omitempty"`
 	Required     bool     `json:"required"`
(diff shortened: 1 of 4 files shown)


pkg/broker/util.go
```diff
@@ -290,33 +289,87 @@ func extractProperties(params []apb.ParameterDescriptor) (map[string]*schema.Sch
...
+
+	// since 0 is a valid value for maximum, minimum, exclusiveMaximum, and exclusiveMinimum,
+	// we have to allow for empty.
+	if max, ok := pd.Maximum.(float64); ok {
+		prop.Maximum = schema.Number{Val: max, Initialized: true}
+	}
+	if min, ok := pd.Minimum.(float64); ok {
+		prop.Minimum = schema.Number{Val: min, Initialized: true}
+	}
+
```
cfchase: In the JSON schema [validations documentation](https://datatracker.ietf.org/doc/draft-handrews-json-schema-validation/), `exclusiveMinimum` and `exclusiveMaximum` MUST be a number.  However go-jsschema and the serialized values it sends uses boolean values for `exclusiveMaximum` and `exclusiveMinimum` are boolean and use the numbers in `maximum` and `minimum`.  

I followed with the docs and  used `exclusive_maximum`/`exclusive_minimum` as a number for the APB spec, but if there are strong opinions that I should use boolean `exclusive_maximum`/`exclusive_minimum` instead, I'd like to hear reasoning.  I was divided myself.

dymurray: I agree with following the docs. `exclusive_maximum` representing the max number seems intuitive to me.

pkg/apb/types.go
```diff
@@ -34,18 +34,30 @@ type SpecManifest map[string]*Spec
...
+
+	// string validators
+	DeprecatedMaxlength int    `json:"maxlength,omitempty" yaml:"maxlength,omitempty"` // backwards compatibility
+	MaxLength           int    `json:"maxLength,omitempty" yaml:"max_length,omitempty"`
+	MinLength           int    `json:"minLength,omitempty" yaml:"min_length,omitempty"`
+	Pattern             string `json:"pattern,omitempty"`
+
+	// number validators
+	MultipleOf       float64     `json:"multipleOf,omitempty" yaml:"multiple_of,omitempty"`
+	Maximum          interface{} `json:"maximum,omitempty"`
```
shawn-hurley: I don't like the loss of type validation if someone is creating a ParameterDescription by hand. If we can make this not exported I would be much less worried about it. 

This is another way around this. Using a type declaration for float64 and dereferencing the pointer when using that value. Example: 

```go
type nilableNumber float64

type ParameterDescription struct {
.....
	Maximum          *nilableNumber `json:"maximum,omitempty"`
.....
}
```

https://play.golang.org/p/qpeFB6b8GXp

cfchase:
@shawn-hurley thanks for the tip.  I switched to the nilableNumber in https://github.com/openshift/ansible-service-broker/pull/615/commits/7450587c1271c57fd9b1f45e45f33f3981d7cf92.

Let me know if this addresses your concerns.

pkg/broker/util.go
```diff
@@ -290,33 +289,87 @@ func extractProperties(params []apb.ParameterDescriptor) (map[string]*schema.Sch
...
+	// omitempty. Setting maxlength to 0 is NOT the same as omitting it.
+	// 0 is a worthless value for DeprecatedMaxlength so we will not set it
+
+	// maxlength
+	if pd.DeprecatedMaxlength > 0 {
+		prop.MaxLength = schema.Integer{Val: pd.DeprecatedMaxlength, Initialized: true}
+	}
+
+	// max_length overrides maxlength
+	if pd.MaxLength > 0 {
```
shawn-hurley: I am pretty sure that you are going to need to dereference the pointer here.

pkg/apb/types.go
```diff
@@ -32,20 +32,34 @@ type Parameters map[string]interface{}
 // SpecManifest - Spec ID to Spec manifest
 type SpecManifest map[string]*Spec
 
+type nilableNumber float64
```
shawn-hurley: I think you should make this exportable so that it can be used when creating a ParameterDescriptor outside of the APB package.

PR #616, opened 2018-01-05
Use the router prefix for apb tool endpoints

rthallisey:
Switch to using the same routing prefix for the apb tool endpoints. The apb tool has all the information needed to use these endpoints.  

partially: https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/issues/180

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -191,9 +191,9 @@ func NewHandler(b broker.Broker, brokerConfig *config.Config, prefix string,
 		createVarHandler(h.lastoperation)).Methods("GET")
 
 	if brokerConfig.GetBool("broker.dev_broker") {
-		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbAddSpec)).Methods("POST")
-		h.router.HandleFunc("/apb/spec/{spec_id}", createVarHandler(h.apbRemoveSpec)).Methods("DELETE")
-		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbRemoveSpecs)).Methods("DELETE")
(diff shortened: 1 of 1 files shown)


pkg/handler/handler.go
```diff
@@ -191,9 +191,9 @@ func NewHandler(b broker.Broker, brokerConfig *config.Config, prefix string,
 		createVarHandler(h.lastoperation)).Methods("GET")
 
 	if brokerConfig.GetBool("broker.dev_broker") {
-		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbAddSpec)).Methods("POST")
-		h.router.HandleFunc("/apb/spec/{spec_id}", createVarHandler(h.apbRemoveSpec)).Methods("DELETE")
-		h.router.HandleFunc("/apb/spec", createVarHandler(h.apbRemoveSpecs)).Methods("DELETE")
+		s.HandleFunc("/v2/apb", createVarHandler(h.apbAddSpec)).Methods("POST")
+		s.HandleFunc("/v2/apb/{spec_id}", createVarHandler(h.apbRemoveSpec)).Methods("DELETE")
+		s.HandleFunc("/v2/apb", createVarHandler(h.apbRemoveSpecs)).Methods("DELETE")
```
shawn-hurley: are we intentionally droping the `/spec` and including this in `.../v2/...` now?

rthallisey: I was thinking we could do that going forward.  Do you think we should keep it as is?

shawn-hurley: no real opinion either way. This will need follow up PRs for the APB tool I think if we do make the change. Other than that it LGTM

rthallisey:
[[3] APB Tooling improvements](https://trello.com/c/qTXN9XVZ/613-3-apb-tooling-improvements)

PR #620, opened 2018-01-08
Add 3.8/3.9 releasers to tito

djzager:
Simply update releasers for tito.


jwmatthews:
We probably don't need a 3.8 section, think that is an unpublished release...but also no harm in putting it in for now.

PR #626, opened 2018-01-10
Bug 588 - Run latest build with a default public ip.

### LINKED ISSUE #588: Getting started guide doesn't work on macOS
> # Bug:

**What happened**:
I tried the "Getting Started with the Ansible Service Broker" document here on my Mac: 
https://github.com/openshift/ansible-service-broker#getting-started-with-the-ansible-service-broker

Being able to start a local OpenShift cluster with ASB enabled.

**How to reproduce it**:
Try this on a Mac:
```
wget https://raw.githubusercontent.com/openshift/ansible-service-broker/master/scripts/run_latest_build.sh
chmod +x run_latest_build.sh
./run_latest_build.sh
```

#### More details:

* Command `ip` doesn't exist on a Mac: https://github.com/openshift/ansible-service-broker/blob/master/scripts/run_latest_build.sh#L36
I modified the file manually to solve this problem.

shawn-hurley:
This PR will default the `PUBLIC_IP` to the localhost. 

This value will allow for macOS to have this value because it can not find the docker interface `docker0`.

The issue that used to exist where the service catalog could not contact the broker is no longer a concern
 
fixes #588

### DIFF
--- a/scripts/run_latest_build.sh
+++ b/scripts/run_latest_build.sh
@@ -34,7 +34,8 @@
 #
 
 DOCKER_IP="$(ip addr show docker0 | grep -Po 'inet \K[\d.]+')"
-PUBLIC_IP=${DOCKER_IP:-"127.0.0.1"}
+DOCKER_IP=${DOCKER_IP:-"127.0.0.1"}
+PUBLIC_IP=${PUBLIC_IP:-$DOCKER_IP}
 HOSTNAME=${PUBLIC_IP}.nip.io
 ROUTING_SUFFIX="${HOSTNAME}"
 ORIGIN_IMAGE=${ORIGIN_IMAGE:-"docker.io/openshift/origin"}
--- a/scripts/run_latest_build.sh
+++ b/scripts/run_latest_build.sh
@@ -34,7 +34,7 @@
 #
 
 DOCKER_IP="$(ip addr show docker0 | grep -Po 'inet \K[\d.]+')"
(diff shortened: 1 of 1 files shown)


scripts/run_latest_build.sh
```diff
@@ -34,7 +34,7 @@
 #
 
 DOCKER_IP="$(ip addr show docker0 | grep -Po 'inet \K[\d.]+')"
-PUBLIC_IP=${PUBLIC_IP:-$DOCKER_IP}
```
mhrivnak: It looks like there used to be some hope that ``PUBLIC_IP`` had already been set. Is that still true? Or is it safe to assume that it is never set until this moment, or at least we are happy *always* overwriting any prior value?

shawn-hurley: I followed the getting started docs and saw nothing about setting the PUBLIC_IP. Basically, the mac has to have a default that will be set to fix the bug. If there is some better way, checking if PUBLIC_IP is already set, then checking DOCKER_IP and then a defaulting I wouldn't mind doing that. I would need some direction though for that :)

mhrivnak: Gotcha, ok. I have no idea if it's important, but just noticed that it was a change in behavior. If you're happy with the behavior, I'm happy. :)

That said, IF it was valuable to use a previously-set PUBLIC_IP, I think you could add a line above this one, and revert this line's change:

```
DOCKER_IP=${DOCKER_IP:-127.0.0.1}
```

PR #635, opened 2018-01-12
Bug 1526949 - Set registry user/pass if auth_type is not defined

dymurray:
Does not wipe out user/pass credentials just because the user didn't set auth_type. Later down the line we can explicity require auth_type if desired but it's not intuitive to set that value for long time users. This bit QE and the code was not in master for awhile so we weren't seeing this.

 - Accept registry user/pass if auth_type is not set

### DIFF
--- a/pkg/registries/registry.go
+++ b/pkg/registries/registry.go
@@ -413,8 +413,9 @@ func retrieveRegistryAuth(reg Config, asbNamespace string) (Config, error) {
 		}
 		return reg, nil
 	case "":
-		username = ""
-		password = ""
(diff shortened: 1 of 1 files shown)


pkg/registries/registry.go
```diff
@@ -413,8 +413,9 @@ func retrieveRegistryAuth(reg Config, asbNamespace string) (Config, error) {
 		}
 		return reg, nil
 	case "":
-		username = ""
-		password = ""
+		// Assuming that the user has either no credentials or defined them in the config
+		username = reg.User
+		password = reg.Pass
```
jmrodri: Okay, this looks good.

PR #636, opened 2018-01-12
Bug 1526949 - Do not ignore credentials if auth_type is not set

dymurray:
QE was not setting `auth_type` in the config which was causing nil credentials. This PR assumes the user defined creds in the config if `auth_type` is not set.

 - Set user/pass instead of nil if `auth_type` is not defined

### DIFF
--- a/pkg/app/config.go
+++ b/pkg/app/config.go
@@ -100,8 +100,9 @@ func CreateConfig(configFile string) (Config, error) {
 			username = config.Registry[regCount].User
 			password = config.Registry[regCount].Pass
 		case "":
-			username = ""
-			password = ""
+			// Assuming that the user has either no credentials or defined them in the config
+			username = reg.User
+			password = reg.Pass
 		default:
 			return Config{}, fmt.Errorf("Unrecognized registry AuthType: %s", reg.AuthType)
 		}


pkg/app/config.go
```diff
@@ -100,8 +100,9 @@ func CreateConfig(configFile string) (Config, error) {
 			username = config.Registry[regCount].User
 			password = config.Registry[regCount].Pass
 		case "":
-			username = ""
-			password = ""
+			// Assuming that the user has either no credentials or defined them in the config
+			username = reg.User
+			password = reg.Pass
```
jmrodri: Why isn't this like the code in `case "config"`?

```
username = config.Registry[regCount].User
password = config.Registry[regCount].Pass
```
Feels like `case ""` should be the same as `case "config"` without the validation of the credentials.

jmrodri:
Also, do we really need to fix this in release-1.0? seems like we can document that the configmap needs to have `"auth_type"` defined.

dymurray:
+1 @jmrodri lets only fix this in master.

dymurray:
https://github.com/openshift/ansible-service-broker/pull/635

PR #637, opened 2018-01-12
Add auth_type params to our template

dymurray:
Adds `auth_type` to the default registry config.

 - Add auth_type and auth_name to deploy template

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -273,8 +273,8 @@ objects:
     name: ${REGISTRY_SECRET_NAME}
     namespace: ansible-service-broker
   data:
-    username: ${DOCKERHUB_USER}
-    password: ${DOCKERHUB_PASS}
+    username: ${REGISTRY_USER}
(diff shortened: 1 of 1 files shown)


jmrodri:
@dymurray was there a bugzilla for this PR?

dymurray:
@jmrodri no BZ for this PR. It came from Tommy Hughes hoping to use this template with the RHCC registry. It can wait to merge if needed.

eriknelson:
I've rerun the build a couple times and it looks like it's timing out on a provision during integration testing. I can't tell if this is legitimate or not, but is the purpose of this change just to parameterize some values? I see two new keys in the registry config that were not there before. Is that causing problems somewhere with the broker?

rthallisey:
let's rebase and pickup the latest changes. Maybe we'll get a different result.

eriknelson:
Restarted the build since we've been seeing a lot of timeouts. @dymurray, this still belongs in master as-is, correct?

dymurray:
I'm closing this PR since the template is mainly used for upstream integration with Dockerhub. I am going to add documentation on using this template with RHCC instead.

PR #642, opened 2018-01-15
Bug 1534715 - unbind checks for existence of binding before trying to delete it

### LINKED ISSUE #640: multiple DELETEs to the same binding result in multiple APB runs
**What happened**:
I initiated an unbind with the service catalog. It made multiple DELETE requests to the same endpoint:

```
172.17.0.2 - - [12/Jan/2018:21:17:47 +0000] "DELETE /ansible-service-broker/v2/service_instances/76593dc4-39d0-4561-a031-e0b4376e5532/service_bindings/7a477324-eb4c-4fb7-9aa6-64a52b2a688d?plan_id=7f4a5e35e4af2beb70076e72fab0b7ff&service_id=1dda1477cace09730bd8ed7a6505607e HTTP/1.1" 200 3
```
The second one returns an HTTP 500 response code:
```
172.17.0.2 - - [12/Jan/2018:21:17:57 +0000] "DELETE /ansible-service-broker/v2/service_instances/76593dc4-39d0-4561-a031-e0b4376e5532/service_bindings/7a477324-eb4c-4fb7-9aa6-64a52b2a688d?plan_id=7f4a5e35e4af2beb70076e72fab0b7ff&service_id=1dda1477cace09730bd8ed7a6505607e HTTP/1.1" 500 82
```
The broker logs make it clear that a second APB container was launched, and I was able to see both pods and both namespaces.

A second DELETE to the same endpoint should not cause a second APB container to launch.

**How to reproduce it**:
1. Enable async binding.
2. Create a binding through the openshift console
3. Delete that binding throug hthe openshift console
4. Watch it run two APBs

mhrivnak:
Unbind was not checking this previously. When doing an async unbind, this meant
multiple APBs got launched. Additionally, all but the first DELETE requests
were geting a 500 response code due to incorrect handling of "err" in code.

fixes #640

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -995,6 +995,15 @@ func (a AnsibleBroker) Unbind(
 		return nil, errors.New(errMsg)
 	}
 
+	// Check if the binding exists
+	_, err := a.dao.GetBindInstance(bindingUUID.String())
+	if err != nil {
+		if client.IsKeyNotFound(err) {
+			return nil, ErrorNotFound
+		}
+		return nil, err
+	}
+
 	params := make(apb.Parameters)
 	provExtCreds, err := a.dao.GetExtractedCredentials(instance.ID.String())
 	if err != nil && !client.IsKeyNotFound(err) {
@@ -1054,6 +1063,7 @@ func (a AnsibleBroker) Unbind(
 			log.Debug("Skipping unbind apb execution")
 			err = nil
 		} else {
+			log.Debug("Launching apb for unbind in blocking mode")
 			err = apb.Unbind(&serviceInstance, &params)
 		}
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/broker/broker.go
```diff
@@ -995,6 +995,16 @@ func (a AnsibleBroker) Unbind(
 		return nil, errors.New(errMsg)
 	}
 
+	// Check if the binding exists
+	_, err := a.dao.GetBindInstance(bindingUUID.String())
+	if err != nil {
+		if client.IsKeyNotFound(err) {
+			return nil, ErrorNotFound
+		} else {
```
jmrodri: remove the else block

```golang
    return nil, ErrorNotFound
}
return nil, err
```

PR #653, opened 2018-01-18
Bug 1536088 - fixes panic when bind can't find ServiceInstance

### LINKED ISSUE #652: bind PUT causes broker panic and 502 when ServiceInstance not found
**What happened**:
I tried to ```PUT /v2/service_instances/$INSTANCE_ID/service_bindings/$BINDING_ID?accepts_incomplete=true``` where ```$INSTANCE_ID``` does not exist.

The broker panics and returns 502.

```
[2018-01-17T21:47:16.399Z] [DEBUG] - Unable to find originating origin header
[2018-01-17T21:47:16.4Z] [ERROR] - Could not find a service instance in dao - 100: Key not found (/service_instance/43e76ef3-7e2a-43ee-b1fe-845a315bf557) [285]
[2018-01-17T21:47:16.4Z] [DEBUG] - Auto Escalate has been set to true, we are escalating permissions
[2018-01-17T21:47:16.4Z] [DEBUG] - Injecting PlanID as parameter: { _apb_plan_id: dev }
[2018-01-17T21:47:16.4Z] [DEBUG] - Injecting ServiceClassID as parameter: { _apb_service_class_id: dh-postgresql-apb-s964m }
[2018-01-17T21:47:16.4Z] [DEBUG] - Injecting ServiceInstanceID as parameter: { _apb_service_instance_id:  }
[2018-01-17T21:47:16.402Z] [INFO] - ASYNC binding in progress
[2018-01-17T21:47:16.402Z] [DEBUG] - bindjob: binding job started, calling apb.Bind
[2018-01-17T21:47:16.402Z] [NOTICE] - ============================================================
[2018-01-17T21:47:16.402Z] [NOTICE] -                        BINDING    
[2018-01-17T21:47:16.402Z] [NOTICE] - ============================================================
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1511412]
goroutine 326 [running]:
github.com/openshift/ansible-service-broker/pkg/apb.Bind(0xc4206dc840, 0xc420168108, 0x0, 0x0, 0x0, 0x0, 0x0)
    /home/mhrivnak/golang/src/github.com/openshift/ansible-service-broker/pkg/apb/bind.go:32 +0x152
github.com/openshift/ansible-service-broker/pkg/broker.(*BindingJob).Run(0xc420732270, 0xc4207322a0, 0x24, 0xc420244900)
    /home/mhrivnak/golang/src/github.com/openshift/ansible-service-broker/pkg/broker/binding_job.go:53 +0x94
created by github.com/openshift/ansible-service-broker/pkg/broker.(*WorkEngine).StartNewJob
    /home/mhrivnak/golang/src/github.com/openshift/ansible-service-broker/pkg/broker/work_engine.go:64 +0x115
```

400 per the OSB API spec

**How to reproduce it**:
Configure the broker to run APBs for bind and unbind, and to auto-escalate, then run the below script.

There is likely a similar failure even when APBs are not run for bind.

```
#!/usr/bin/env bash

INSTANCE_ID=$(uuidgen)
BINDING_ID=$(uuidgen)
PLAN_UUID="7f4a5e35e4af2beb70076e72fab0b7ff"
SERVICE_UUID="dh-postgresql-apb-s964m"

req="{
  \"plan_id\": \"$PLAN_UUID\",
  \"service_id\": \"$SERVICE_UUID\",
  \"context\": \"blog-project\",
  \"app_guid\":\"\",
  \"bind_resource\":{},
  \"parameters\": {}
}"

curl \
    -k \
    -i \
    -X PUT \
    -H "Authorization: bearer $(oc whoami -t)" \
    -H "Content-type: application/json" \
    -H "Accept: application/json" \
    -H "X-Broker-API-Originating-Identity: " \
    -d "$req" \
    "https://asb-1338-ansible-service-broker.172.17.0.1.nip.io/ansible-service-broker/v2/service_instances/$INSTANCE_ID/service_bindings/$BINDING_ID?accepts_incomplete=true"
```

mhrivnak:
fixes #652

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -537,6 +537,7 @@ func (h handler) bind(w http.ResponseWriter, r *http.Request, params map[string]
 		default:
 			writeResponse(w, http.StatusInternalServerError, broker.ErrorResponse{Description: err.Error()})
 		}
+		return
 	}
 
 	if !h.brokerConfig.GetBool("broker.auto_escalate") {
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -35,10 +35,11 @@ import (
 )
 
 type MockBroker struct {
-	Name      string
-	Verify    map[string]bool
-	Err       error
-	Operation string
+	Name                  string
+	Verify                map[string]bool
+	Err                   error
+	Operation             string
+	getServiceInstanceErr error
 }
 
 type MockUserServiceAdapter struct {
@@ -106,7 +107,7 @@ func (m MockBroker) LastOperation(uuid.UUID, *broker.LastOperationRequest) (*bro
 }
 
 func (m MockBroker) GetServiceInstance(uuid.UUID) (apb.ServiceInstance, error) {
-	return apb.ServiceInstance{}, nil
+	return apb.ServiceInstance{}, m.getServiceInstanceErr
 }
 
 func (m MockBroker) GetBindInstance(uuid.UUID) (apb.BindInstance, error) {
@@ -365,7 +366,8 @@ func TestBindAlreadyProvisioned(t *testing.T) {
 }
 
 func TestBindNotFound(t *testing.T) {
-	testhandler, w, r, params := buildBindHandler(uuid.New(), broker.ErrorNotFound)
+	testhandler, w, r, params := buildBindHandler(uuid.New(), nil)
+	testhandler.broker = MockBroker{Name: "testbroker", Err: nil, getServiceInstanceErr: broker.ErrorNotFound}
 	testhandler.bind(w, r, params)
 	ft.AssertEqual(t, w.Code, 400, "should've been a bad request for error not found")
 	ft.AssertError(t, w.Body, "not found")


pkg/handler/handler.go
```diff
@@ -533,10 +533,11 @@ func (h handler) bind(w http.ResponseWriter, r *http.Request, params map[string]
 	if err != nil {
 		switch err {
 		case broker.ErrorNotFound:
-			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: "ServiceInstance not found"})
```
rthallisey: Do we want to keep ```err.Error()``` since there are multiple errors we could return back from ```GetServiceInstance```?

mhrivnak: Hm. I'm assuming that if ``GetServiceInstance`` returns a ``ErrorNotFound``, it means the service instance was not found. Any other reason for returning that error would be surprising, I think. Different failure reasons should result in different error values, right?

I changed the message because from a client perspective, they just saw description "not found", without any hint as to *what* wasn't found. So in any case, at some point in the stack I think we need to get specific about what wasn't found, so the error in the HTTP response can be more helpful. A different way to do that would be to have ``GetServiceInstance`` return an error that already has a specific message.

rthallisey: > Hm. I'm assuming that if GetServiceInstance returns a ErrorNotFound, it means the service instance was not found.
>
I agree, I think the the right assumption. Because we're either returning ErrorNotFound or the error from ```getObject(...)```, which I think are going to be the same.

> A different way to do that would be to have GetServiceInstance return an error that already has a specific message.
>
I think this would be beneficial because there are multiple places in the handler that call ```broker.GetServiceInstance```. We could pass back the error message from that function so it improves the error message for all callers.

jmrodri: I would have just left this as `err.Error()`. Almost all the other cases of `ErrorNotFound` we just return `err.Error()`

```bash
$ grep -A2 ErrorNotFound pkg/handler/handler.go
		case broker.ErrorNotFound: // return 404
			writeResponse(w, http.StatusNotFound, broker.ErrorResponse{Description: err.Error()})
		default: // return 422
--
		case broker.ErrorNotFound:
		default:
--
		case broker.ErrorNotFound:
		default:
--
		case broker.ErrorNotFound:
			writeResponse(w, http.StatusGone, broker.DeprovisionResponse{})
			return
--
		case broker.ErrorNotFound:
			writeResponse(w, http.StatusGone, broker.DeprovisionResponse{})
			return
--
		case broker.ErrorNotFound:
		default:
--
		case broker.ErrorNotFound:
			writeResponse(w, http.StatusNotFound, broker.ErrorResponse{Description: err.Error()})
		default:
--
		case broker.ErrorNotFound:
		default:
--
		case broker.ErrorNotFound:
		default:
--
		case broker.ErrorNotFound:
			writeResponse(w, http.StatusGone, nil)
		default:
--
		case broker.ErrorNotFound:
			writeResponse(w, http.StatusGone, nil)
		default:
--
		case broker.ErrorNotFound: // return 404
			log.Debugf("Binding not found.")
			writeResponse(w, http.StatusNotFound, broker.ErrorResponse{Description: err.Error()})
```

shawn-hurley:
@jmrodri @rthallisey it makes sense to deal with `err.Error()` stuff in a different PR.

pkg/handler/handler.go
mhrivnak: It looks like there are a lot of cases already in ``handler.go`` where a handler function is setting a specific error message when writing a response, based on inspecting the return values (error included) from some other function. I did a search for ``Description: "`` and came up with 29 hits in that file.

Beyond just the ``Description``, the handler functions are always deciding what http response code to use based on details of an error or other return values, and that's arguably more impactful than the human-intended text description.

So my take is that this behavior is already pervasive within ``handler.go``. Inspecting errors from the broker and using the details to compose new and descriptive HTTP responses is the established precedent. 

That said, I very much see your point about there being an opportunity to tighten up the error handling flow and better-define that part of the API between the broker and handlers. The errors are being used today both to decide what response code to use, and to sometimes convey a human-readable error message for either devs or http client users (but it's not always clear which). Those two roles are in conflict with each other in some cases, as we're seeing here, so it would be great to better-define that. However that is the direction that I think would require a lot of change (for example it gets more complex to have a generic ``ErrorNotFound`` as we do now if we expect it to specify an object type). Maybe it makes sense to tackle that as an enhancement once we're out of bug-fix mode? I'd be happy to help out with it.

For now, I think the options are to keep the PR as-is, or go back to just using the description provided from the error. I don't love the latter, because it's a worse user experience, but I'm ok either way. Improving the error message coming out of ``broker.go`` would be a much bigger project.

mhrivnak:
Ok, I reverted the error message change, so we can save that for another day.

PR #654, opened 2018-01-18
Bug 1506978 - Apply proxy settings to running APBs

eriknelson:
* Broker should inspect its own enviornment, and in the presence of
a proxy configuration, should apply those settings to executing APBs.

### DIFF
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


eriknelson:
Currently trying to test this in the example locked down environment, but I confirmed the settings are being applied to running APBs as expected when the various env vars are set.

pkg/apb/executor.go
```diff
@@ -154,6 +140,34 @@ func ExecuteApb(action string,
...
+// GetProxyConfig - Returns a ProxyConfig based on the presence of a proxy
+// configuration in the broker's environment. HTTP_PROXY, HTTPS_PROXY, and
+// NO_PROXY are the relevant environment variables. If no proxy is found,
+// GetProxyConfig will return nil.
+func GetProxyConfig() *ProxyConfig {
+	httpProxy, httpProxyPresent := os.LookupEnv(httpProxyEnvVar)
+	httpsProxy, httpsProxyPresent := os.LookupEnv(httpsProxyEnvVar)
+	noProxy, noProxyPresent := os.LookupEnv(noProxyEnvVar)
+
+	// TODO: Probably some more permutations of these that should be validated?
```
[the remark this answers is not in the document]
eriknelson: Yep. I can file an issue after this merges and follow up with a patch + tests.

pkg/apb/executor.go
```diff
@@ -120,24 +123,7 @@ func ExecuteApb(action string,
...
-						v1.EnvVar{
-							Name: "POD_NAMESPACE",
-							ValueFrom: &v1.EnvVarSource{
-								FieldRef: &v1.ObjectFieldSelector{
-									FieldPath: "metadata.namespace",
-								},
-							},
-						},
-					},
+					Env:             createPodEnv(executionContext),
```
jmrodri: Nice cleanup!

pkg/apb/executor.go
```diff
@@ -154,6 +140,34 @@ func ExecuteApb(action string,
...
+		return nil
+	}
+
+	return &ProxyConfig{
+		HttpProxy:  httpProxy,
+		HttpsProxy: httpsProxy,
+		NoProxy:    noProxy,
+	}
+}
+
```
jmrodri: 👍

pkg/apb/executor.go
```diff
@@ -201,6 +215,52 @@ func createExtraVars(context *Context, parameters *Parameters) (string, error) {
 	return string(extraVars), err
 }
 
+func createPodEnv(executionContext ExecutionContext) []v1.EnvVar {
```
jmrodri: EXCELLENT!

pkg/apb/executor.go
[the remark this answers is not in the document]
eriknelson: #663

PR #668, opened 2018-01-19
Update test bash scripts

jmrodri:
During async bind, I had a set of bash scripts: `bind.sh`, `unbind.sh`, `getbind.sh`, and `getinstance.sh` which worked with the broker's route. They were duplicates of the original bash scripts in the `test` folder. 

The older scripts were also outdated, so I brought them up to date with my newer scripts. They now also take in parameters for some of the arguments. They are not perfect but get the job done.

### DIFF
--- a/test/bind.sh
+++ b/test/bind.sh
@@ -1,23 +1,39 @@
 #!/bin/bash -e
 
-. shared.sh
+INSTANCE_ID=$1
+BINDING_ID=$2
+PLAN_UUID="7f4a5e35e4af2beb70076e72fab0b7ff"
+SERVICE_UUID="dh-postgresql-apb-s964m"
 
-# mlab with params
-instanceUUID="8c9adf85-9221-4776-aa18-aae7b7acc436"
+if [ "$INSTANCE_ID" = "" ]
+then
+  echo "Usage: $0 <instance uuid> <binding uuid>"
+  exit
+fi
+
+if [ "$BINDING_ID" = "" ]
+then
+  echo "Usage: $0 <instance uuid> <binding uuid>"
+  exit
+fi
+
 req="{
-  \"plan_id\": \"$planUUID\",
-  \"service_id\": \"$serviceUUID\",
+  \"plan_id\": \"$PLAN_UUID\",
+  \"service_id\": \"$SERVICE_UUID\",
+  \"context\": \"blog-project\",
   \"app_guid\":\"\",
   \"bind_resource\":{},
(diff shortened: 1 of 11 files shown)


test/bind.sh
```diff
@@ -1,23 +1,39 @@
 #!/bin/bash -e
 
-. shared.sh
+INSTANCE_ID=$1
+BINDING_ID=$2
+PLAN_UUID="7f4a5e35e4af2beb70076e72fab0b7ff"
```
shawn-hurley: This looks like it is hardcoded I assume there is a registry config that we must use to get this to work. I think it might be worth documenting in the README.

test/bind.sh
```diff
@@ -1,23 +1,39 @@
...
 curl \
-  -X PUT \
-  -H 'X-Broker-API-Version: 2.9' \
-  -H 'Content-Type: application/json' \
-  -d "$req" \
-  -v \
-  http://localhost:1338/v2/service_instances/$instanceUUID/service_bindings/$bindingUUID
+    -k \
+    -X PUT \
+    -H "Authorization: bearer $(oc whoami -t)" \
```
shawn-hurley: Also, make sure the user knows they must be logged into the cluster and have the right permissions.

test/bind.sh
```diff
@@ -1,23 +1,39 @@
...
+    -H "Content-type: application/json" \
+    -H "Accept: application/json" \
+    -H "X-Broker-API-Originating-Identity: " \
```
shawn-hurley: I think you're going to need to set auto_escalate=true for these scripts to work, I would also document that.

test/bind.sh
djzager: And you can't be a `system:admin` for this.

test/bind.sh
jmrodri: @shawn-hurley heh it was hardcoded because on a few of the calls we don't really care apparently. Clearly this is NOT going to work with provision. I'll look into it.

jmrodri:
@shawn-hurley was not supposed to replace asbcli. I guess it could since I haven't used asbcli in ages. Does it even work anymore? Just some generally useful scripts to do curl against the broker's api.

test/bind.sh
jmrodri: added to the readme

test/bind.sh
jmrodri: added to the readme

PR #669, opened 2018-01-19
Bug 1536659 - bind PUT returns http code 202 when operation runs async

### LINKED ISSUE #667: async bind returns 201 instead of 202
**What happened**:
I used the broker API to create a binding by making this request:

```
PUT /ansible-service-broker/v2/service_instances/c4aa708d-8c64-4c37-a131-557a4a155148/service_bindings/745ddeb7-395e-412e-b4ed-8a0b52b11ac0?accepts_incomplete=true 
```

The response code was 201, but should have been 202 according to the OSB API spec proposed changes around async bind/unbind.

```
HTTP/1.1 201 Created
Date: Fri, 19 Jan 2018 19:42:53 GMT
Content-Length: 58
Set-Cookie: d05351dd24520dce41f310fcfc112523=1591f023a03f3453b1d2364f14c83799; path=/; HttpOnly; Secure

{
  "operation": "338c289e-8792-4ab4-95fa-4fd9b4752cf0"
}
```

202 response code

**How to reproduce it**:
1. configure the broker to run APBs on bind/unbind and auto-escalate
2. Run this script, supplying a valid instance ID as the only argument

```
#!/usr/bin/env bash

INSTANCE_ID=$1
BINDING_ID=$(uuidgen)
PLAN_UUID="7f4a5e35e4af2beb70076e72fab0b7ff"
SERVICE_UUID="dh-postgresql-apb-s964m"

req="{
  \"plan_id\": \"$PLAN_UUID\",
  \"service_id\": \"$SERVICE_UUID\",
  \"context\": \"blog-project\",
  \"app_guid\":\"\",
  \"bind_resource\":{},
  \"parameters\": {}
}"

curl \
    -k \
    -i \
    -X PUT \
    -H "Authorization: bearer $(oc whoami -t)" \
    -H "Content-type: application/json" \
    -H "Accept: application/json" \
    -H "X-Broker-API-Originating-Identity: " \
    -d "$req" \
    "https://asb-1338-ansible-service-broker.172.17.0.1.nip.io/ansible-service-broker/v2/service_instances/$INSTANCE_ID/service_bindings/$BINDING_ID?accepts_incomplete=true"
```

mhrivnak:
Previously the http response incorrectly had code 201.

Once async logic was removed from buildBindRequest, it looked an awful lot like
a simple constructor. That's why I moved it next to the type definition and
renamed it NewBindRequest to match golang convention.

fixes #667

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -79,7 +79,7 @@ type Broker interface {
 	Provision(uuid.UUID, *ProvisionRequest, bool) (*ProvisionResponse, error)
 	Update(uuid.UUID, *UpdateRequest, bool) (*UpdateResponse, error)
 	Deprovision(apb.ServiceInstance, string, bool, bool) (*DeprovisionResponse, error)
-	Bind(apb.ServiceInstance, uuid.UUID, *BindRequest, bool) (*BindResponse, error)
+	Bind(apb.ServiceInstance, uuid.UUID, *BindRequest, bool) (*BindResponse, bool, error)
 	Unbind(apb.ServiceInstance, apb.BindInstance, string, bool, bool) (*UnbindResponse, error)
 	LastOperation(uuid.UUID, *LastOperationRequest) (*LastOperationResponse, error)
 	Recover() (string, error)
@@ -804,12 +804,14 @@ func (a AnsibleBroker) GetBind(instance apb.ServiceInstance, bindingUUID uuid.UU
 	}
 
 	log.Debug("broker.GetBind: we got the bind credentials")
-	return a.buildBindResponse(provExtCreds, bindExtCreds, false, "")
+	return NewBindResponse(provExtCreds, bindExtCreds)
 }
 
-// Bind - will create a binding between a service.
+// Bind - will create a binding between a service. Parameter "async" declares
+// whether the caller is willing to have the operation run asynchronously. The
+// returned bool will be true if the operation actually ran asynchronously.
 func (a AnsibleBroker) Bind(instance apb.ServiceInstance, bindingUUID uuid.UUID, req *BindRequest, async bool,
-) (*BindResponse, error) {
+) (*BindResponse, bool, error) {
 	// binding_id is the id of the binding.
 	// the instanceUUID is the previously provisioned service id.
 	//
@@ -826,17 +828,17 @@ func (a AnsibleBroker) Bind(instance apb.ServiceInstance, bindingUUID uuid.UUID,
 		errMsg :=
 			"PlanID from bind request is blank. " +
 				"Bind requests must specify PlanIDs"
-		return nil, errors.New(errMsg)
+		return nil, false, errors.New(errMsg)
 	}
 
 	planName, err := a.dao.GetPlanName(req.PlanID)
 	if err != nil {
 		// etcd return not found i.e. code 100
 		if client.IsKeyNotFound(err) {
-			return nil, ErrorNotFound
+			return nil, false, ErrorNotFound
 		}
 		// otherwise unknown error bubble it up
-		return nil, err
+		return nil, false, err
 	}
 
 	log.Debugf(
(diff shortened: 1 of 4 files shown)


pkg/broker/broker.go
```diff
@@ -79,7 +79,7 @@ type Broker interface {
 	Provision(uuid.UUID, *ProvisionRequest, bool) (*ProvisionResponse, error)
 	Update(uuid.UUID, *UpdateRequest, bool) (*UpdateResponse, error)
 	Deprovision(apb.ServiceInstance, string, bool, bool) (*DeprovisionResponse, error)
-	Bind(apb.ServiceInstance, uuid.UUID, *BindRequest, bool) (*BindResponse, error)
+	Bind(apb.ServiceInstance, uuid.UUID, *BindRequest, bool) (*BindResponse, bool, error)
```
jmrodri: Why are we returning a bool?

pkg/broker/broker.go
```diff
@@ -817,12 +817,14 @@ func (a AnsibleBroker) GetBind(instance apb.ServiceInstance, bindingUUID uuid.UU
 	}
 
 	log.Debug("broker.GetBind: we got the bind credentials")
-	return a.buildBindResponse(provExtCreds, bindExtCreds, false, "")
+	return NewBindResponse(provExtCreds, bindExtCreds)
```
jmrodri: I like this better.

pkg/handler/handler.go
```diff
@@ -560,7 +560,7 @@ func (h handler) bind(w http.ResponseWriter, r *http.Request, params map[string]
 	}
 
 	// process binding request
-	resp, err := h.broker.Bind(serviceInstance, bindingUUID, req, async)
+	resp, ranAsync, err := h.broker.Bind(serviceInstance, bindingUUID, req, async)
```
jmrodri: I don't think we need to return this boolean. We know if we are async based on the async. `Provision` returns similar values without returning a `ranAsync`. I'd prefer to model it similar to `Provision`.

```golang
    if err != nil {
        switch err {
        case broker.ErrorDuplicate:
            writeResponse(w, http.StatusConflict, broker.ProvisionResponse{})
        case broker.ErrorProvisionInProgress:
            writeResponse(w, http.StatusAccepted, resp)
        case broker.ErrorAlreadyProvisioned:
            writeResponse(w, http.StatusOK, resp)
        case broker.ErrorNotFound:
        default:
        }
    } else if async {
        writeDefaultResponse(w, http.StatusAccepted, resp, err)
    } else {
        writeDefaultResponse(w, http.StatusCreated, resp, err)
    }
```

pkg/handler/handler_test.go
```diff
@@ -93,9 +93,9 @@ func (m MockBroker) Deprovision(apb.ServiceInstance, string, bool, bool) (*broke
 	m.called("deprovision", true)
 	return nil, m.Err
 }
-func (m MockBroker) Bind(apb.ServiceInstance, uuid.UUID, *broker.BindRequest, bool) (*broker.BindResponse, error) {
+func (m MockBroker) Bind(apb.ServiceInstance, uuid.UUID, *broker.BindRequest, bool) (*broker.BindResponse, bool, error) {
```
jmrodri: -1 to returning `bool`.

pkg/handler/handler.go
mhrivnak: When ``async == true`` for the argument being passed into the broker's Bind function, there are at least two cases where there will not be asynchronous execution.

1. if the binding already exists
2. if the broker is configured for ``launch_apb_on_bind = false``

In theory there could be other reasons for doing it synchronously even if ``async == true``.

Put another way, when ``async == true``, all that really means is that the client specified ``accepts_incomplete=true``. That is not a guarantee that asynchronous execution will happen. The handler must allow the broker to decide whether async execution is appropriate, and must then compose an http response based on what the broker decided. And the only way for the handler to know what the broker decided is if the broker tells it, hence the ``bool``.

A different but perfectly valid (I think?) way to handle this would be to add an attribute to the response object and communicate through that. But since the response object gets directly serialized into an HTTP response body, that might be overloading its definition.

We could add a private field "async" to the response, and then a method that returns whether the attribute was async or not, and I assume that would allow the field to be omitted from serialization to JSON? Something like:

```
type BindResponse struct {
	Credentials     map[string]interface{} `json:"credentials,omitempty"`
	SyslogDrainURL  string                 `json:"syslog_drain_url,omitempty"`
	RouteServiceURL string                 `json:"route_service_url,omitempty"`
	VolumeMounts    []interface{}          `json:"volume_mounts,omitempty"`
	Operation       string                 `json:"operation,omitempty"`
        ranAsync        bool
}

// returns true if the operation executed asynchronously, otherwise false
func (b BindResponse) RanAsync() (bool)
        return b.ranAsync
```

Even if that's doable, it's probably simpler, and definitely just as effective, to return a separate value.

jmrodri: @mhrivnak very good explanation to why you did it differently than provision. Okay that makes sense now. Sorry I missed that.

jmrodri:
@mhrivnak lint check failed. Please run `make check` for all the checks or `make lint` to verify the linter only.

mhrivnak:
@jmrodri thanks for pointing that out. I fixed one of the two lint errors, but the other appears to be failing on master and is unrelated to this PR:

```
pkg/broker/util.go:136:2: can probably use "var formDefinition []interface{}" instead
```

pkg/broker/broker.go
```diff
@@ -945,62 +947,36 @@ func (a AnsibleBroker) Bind(instance apb.ServiceInstance, bindingUUID uuid.UUID,
...
-	}
-
-	if bCreds != nil {
-		log.Debugf("bind creds: %v", bCreds.Credentials)
-		return &BindResponse{Credentials: bCreds.Credentials}, nil
-	}
-
-	log.Debugf("provision bind creds: %v", pCreds.Credentials)
-	return &BindResponse{Credentials: pCreds.Credentials}, nil
+	resp, err := NewBindResponse(provExtCreds, bindExtCreds)
```
eriknelson: +1, no dependency on the state of the broker, no reason to have it as a method.

jmrodri:
@mhrivnak needs rebase

Issue #504, opened 2017-10-19
SetState (JobState) from within StartNewJob?


eriknelson:
Was surprised to find the Provision routine in broker.go is actually creating and setting job state, instead of `engine.StartNewJob`. I got bit tonight by this because I thought the initial job state setter was taken care of transparently, but evidently not:

https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/broker.go#L673

I'm thinking that call ^ needs to get moved into whatever function actually initiates the job?

EDIT: Trying to articulate my complaint a little better; the current design is prone to error. An author that kicks off a new job **must** remember to set the initial state also in etcd, and it's likely experienced devs will forget to do so 6mo down the road (like this bonehead), and people new to the project won't know to do it at all. Think we can improve the dev ergonomics there.

eriknelson:
Possibly related to #608? Need to investigate.

PR #671, opened 2018-01-20
Bug 1536629 - Send job msg immediately as job starts. To set initial JobState 

maleck13:
 - removes the need to explicitly set state in broker after starting new job

fixes 504

@eriknelson I believe this change will stop the need to set the state in the broker and fix https://github.com/openshift/ansible-service-broker/issues/504

### DIFF
--- a/pkg/broker/binding_job.go
+++ b/pkg/broker/binding_job.go
@@ -21,6 +21,7 @@
 package broker
 
 import (
+	"github.com/apex/log"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 	"github.com/openshift/ansible-service-broker/pkg/metrics"
 	"github.com/pborman/uuid"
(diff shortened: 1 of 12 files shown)


rthallisey:
@maleck13 Does this need to go before https://github.com/openshift/ansible-service-broker/pull/619?

maleck13:
@rthallisey yeah I think that makes sense. It is a small and distinct piece of work so can prob be merged faster then rebased onto the larger PR

pkg/broker/binding_job.go
```diff
@@ -21,6 +21,7 @@
 package broker
 
 import (
+	"github.com/apex/log"
```
rthallisey: Looks like github.com/apex snuck in there again

maleck13: Sigh I think it must be my goimports that auto runs. :(

shawn-hurley:
I think this needs a Bugzilla to merge.

maleck13:
Is the Bugzilla  something I need to do?

jmrodri:
@maleck13 I already had a bug relating to job state. Reusing that one.

PR #673, opened 2018-01-22
fixes #665 - remove many of the TODOs

### LINKED ISSUE #665: Too many TODOs
[A lot of TODOs remain in the project](https://gist.githubusercontent.com/eriknelson/90b23355ec5c8bc2aeb2f913682b25db/raw/867847d72878e7a4f40d372368726d010de67118/broker.todo.txt), I count 47.

We should make an effort to cut these down so they're actually meaningful, and commit to only accepting them in new code if an associated issue is made (@shawn-hurley was good about calling me out 👍). Existing ones that are relevant should remain and be logged as issues. If not relevant, they should get pruned. A lot of them are vestiges of the early codebase. I would love to see some housekeeping.

jmrodri:
Reduced the *TODO* count from 60 to 38.

```
$ git grep TODO| grep -v vendor | wc -l                                                                 
60                                   
$ git grep TODO | grep -v vendor | wc -l
38
```

fixes #665

### DIFF
--- a/ansible-service-broker.spec
+++ b/ansible-service-broker.spec
@@ -1,6 +1,5 @@
 %if 0%{?fedora} || 0%{?rhel} >= 6
 %global with_devel 1
-# TODO: package new deps
 %global with_bundled 0
 %global with_debug 0
 %global with_check 0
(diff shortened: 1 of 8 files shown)


eriknelson:
Merging without a bug because this is all comment removal with no functional changes.

PR #677, opened 2018-01-23
Bug 1537367 - missing last_operation for bindings

jmrodri:
Missing last_operation for the async binds. The service_instance method would work because our jobs are the same. But OSB platforms will want to call the last_operation binding.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -184,6 +184,8 @@ func NewHandler(b broker.Broker, brokerConfig *config.Config, prefix string,
 		createVarHandler(h.unbind)).Methods("DELETE")
 	s.HandleFunc("/v2/service_instances/{instance_uuid}/last_operation",
 		createVarHandler(h.lastoperation)).Methods("GET")
+	s.HandleFunc("/v2/service_instances/{instance_uuid}/service_bindings/{binding_uuid}/last_operation",
+		createVarHandler(h.lastoperation)).Methods("GET")
 
 	if brokerConfig.GetBool("broker.dev_broker") {
 		s.HandleFunc("/v2/apb", createVarHandler(h.apbAddSpec)).Methods("POST")
@@ -692,9 +694,18 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
 		return
 	}
 
+	// we have a binding job
+	if strings.Index(r.URL.Path, "/service_bindings/") > 0 {
+		bindingUUID := uuid.Parse(params["binding_uuid"])
+		if bindingUUID == nil {
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: "invalid binding_uuid"})
+			return
+		}
+	}
+
 	req := broker.LastOperationRequest{}
 
-	// operation is rqeuired
+	// operation is required
 	if op := r.FormValue("operation"); op != "" {
 		req.Operation = op
 	} else {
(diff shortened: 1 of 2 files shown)


jmrodri:
@mhrivnak Have a looksie

rthallisey:
related: https://github.com/openshift/ansible-service-broker/issues/475

pkg/handler/handler.go
```diff
@@ -688,9 +690,18 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
...
+		if bindingUUID == nil {
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: "invalid binding_uuid"})
+			return
+		}
+	}
+
 	req := broker.LastOperationRequest{}
 
-	// operation is rqeuired
+	// operation is required
```
mhrivnak: Speaking of this comment, the code below actually makes it optional. <shrug> It just logs a warning but continues its lookup using just the service ID. I think it would be better to start enforcing what this comment says, and make it required, if not for the service last operation, at least for this case of a binding last operation.

If we let it continue without the ``operation``, I'm not sure how it would behave. Maybe it would just fail to lookup the state? Or would it grab some other ``JobState`` for the same service ID, but not necessarily the right one? Either way, it's better to not let it get that far.

jmrodri: Great point.

jmrodri: It returns an error

```
< HTTP/1.1 400 Bad Request                          
< Content-Type: application/json                    
< Date: Tue, 23 Jan 2018 16:33:31 GMT               
< Content-Length: 124                               
< Set-Cookie: d05351dd24520dce41f310fcfc112523=2808fe5497bbaeea73011198cb5cc014; path=/; HttpOnly; Secure                                                                                                          
<                                                   
{                                                   
  "description": "operation not supplied for a last_operation with instance_uuid [3aae62b1-f990-43de-89b6-25ad2c69e4e1]"                                                                                            
} 
```

jmrodri:
@mhrivnak addressed both of your comments.

pkg/handler/handler.go
```diff
@@ -692,13 +694,38 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
...
+
 	req := broker.LastOperationRequest{}
 
-	// operation is rqeuired
+	// operation is expected
 	if op := r.FormValue("operation"); op != "" {
 		req.Operation = op
 	} else {
-		log.Warning(fmt.Sprintf("operation not supplied, relying solely on the instance_uuid [%s]", instanceUUID))
+		errmsg := fmt.Sprintf("operation not supplied for a last_operation with instance_uuid [%s]", instanceUUID)
```
mhrivnak: Would it make sense to shuffle this around to something like this, to factor out the ``else``?

```
if op := r.FormValue("operation"); op == "" {
  // do error stuff
  return
}
req.Operation = op
...
```

pkg/handler/handler.go
```diff
@@ -692,13 +694,38 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
 		return
 	}
 
+	// we have a binding job
+	if strings.Index(r.URL.Path, "/service_bindings/") > 0 {
```
shawn-hurley: `strings.Contains` [might](https://golang.org/pkg/strings/#Contains) be more helpful when reading this? it is basically just a wrapper around what you are doing though so meh.

rthallisey: Using ```strings.Contains``` would make the code read a little better.

PR #679, opened 2018-01-23
One CI fix and a few improvements

rthallisey:
Fix:
  - ```git diff FETCH_HEAD...$TRAVIS_BRANCH``` was always returning true.  Instead, just use ```git diff $TRAVIS_BRANCH``` it will compare the diff from the current PR branch to the master branch.

Improvement:
  - Use the freshly built service-catalog images
  - Reduce service-catalog log-level to 2
  - Only gather logs from the service-catalog from the last 5 minutes.
  - List the docker images that travis has

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
         # Copied from https://github.com/facebook/react/pull/2000
         - |
             if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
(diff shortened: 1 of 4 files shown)


rthallisey:
@jmrodri slight amendment to the previous PR.

rthallisey:
Merging this with one ack since we need to unblock the gates.

PR #690, opened 2018-01-25
add a 3.10 releaser

jmrodri:
adding a 3.10 releaser, the brew tag doesn't exist yet but it will prevent you from building 3.10 into 3.9.


jmrodri:
@jmontleon I wonder if we should remove the other releases from master.

Issue #698, opened 2018-01-28
deprovision handler is missing fall-through case


eriknelson:
https://github.com/openshift/ansible-service-broker/blob/master/pkg/handler/handler.go#L443

We should have a default with something like an internal server error if this case falls through. This is what's done with most of the other action's handlers.

PR #700, opened 2018-01-29
Add fall through case to deprovision handler

PhilipGough:

Adds a default error to deprovision handler

- Return a 500 status code as a fallback

fixes https://github.com/openshift/ansible-service-broker/issues/698


pkg/handler/handler.go
```diff
@@ -443,6 +443,8 @@ func (h handler) deprovision(w http.ResponseWriter, r *http.Request, params map[
 		case broker.ErrorDeprovisionInProgress:
 			writeResponse(w, http.StatusAccepted, broker.DeprovisionResponse{})
 			return
+		default:
+			writeResponse(w, http.StatusInternalServerError, broker.ErrorResponse{Description: err.Error()})
```
rthallisey: I think you also need a return after you respond with the error.

eriknelson:
Thank you @PhilipGough!

PR #713, opened 2018-01-31
Bug 1538986 - Remove bad enum values from Update Request

jmontleon:
Fix a bug

 - Drop requested parameter updates with invalid enum values
 
Doesn't depend on anything else.

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -1200,6 +1200,21 @@ func (a AnsibleBroker) Update(instanceUUID uuid.UUID, req *UpdateRequest, async 
 
 	req.Parameters = a.validateRequestedUpdateParams(req.Parameters, toPlan, si)
 
+	if fromPlanName == toPlanName && len(req.Parameters) == 0 {
+		log.Warningf("Returning without running the APB. No changes were actually requested")
+
+		noopUpdateUuid := uuid.New()
+
+		a.dao.SetState(si.ID.String(), apb.JobState{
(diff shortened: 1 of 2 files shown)


shawn-hurley:
I think that this [line](https://github.com/jmontleon/ansible-service-broker/blob/a1080b3b239d7762cd0bbdba7c124128c9316f69/pkg/handler/handler.go#L359) also needs to get updated so that we return the 2xx when the new no parameters changes error occurs.

jmontleon:
@shawn-hurley I think this is no longer an issue with the most recent set of changes.

pkg/broker/broker.go
```diff
@@ -1200,6 +1200,19 @@ func (a AnsibleBroker) Update(instanceUUID uuid.UUID, req *UpdateRequest, async
...
+		log.Warningf("Returning without running the APB. No changes were actually requested")
+
+		a.dao.SetState(si.ID.String(), apb.JobState{
+			Token:   "noop-update",
+			State:   apb.StateSucceeded,
+			Podname: "noop-update",
+			Method:  apb.JobMethodUpdate,
+		})
+
+		return &UpdateResponse{"noop-update"}, nil
```
shawn-hurley: This might seem silly, but this would be constantly overwriting the same job I believe. I wonder if generating a UUID for the token even for the `noop` might make sense?

jmrodri:
@jmontleon needs a PR against release-1.1 branch to be included in 3.9 builds.

PR #716, opened 2018-02-01
Bug 1533425 - handle missing pod action error

jmrodri:
apb-base will now return a non-zero exit status meaning the pod will be marked as failed when an action that doesn't exist is used. For example, calling bind on an apb that does NOT have a bind.yaml. 

PR depends on 2 others to satisfy the bugzilla:

apb-base: https://github.com/ansibleplaybookbundle/apb-base/pull/17
postgresql-apb: https://github.com/ansibleplaybookbundle/postgresql-apb/pull/32

### DIFF
--- a/pkg/apb/watch_pod.go
+++ b/pkg/apb/watch_pod.go
@@ -59,17 +59,7 @@ func watchPod(podName string, namespace string) error {
 			}
 
 			// handle the return code from the pod
-			exitCode := getExitCode(podStatus.ContainerStatuses)
-			if exitCode == 8 {
-				log.Errorf("Pod [ %s ] failed - action not found.", podName)
-				return ErrorActionNotFound
-			} else if exitCode > 0 {
-				return fmt.Errorf("Pod [ %s ] failed with exit code [%d]", podName, exitCode)
-			} else if exitCode < 0 {
-				return fmt.Errorf("Pod [ %s ] failed. Unable to determine exit code [%d]", podName, exitCode)
-			}
-
-			return fmt.Errorf("Pod [ %s ] failed - %v", podName, podStatus.Message)
+			return translateExitStatus(podName, podStatus)
 		case apiv1.PodSucceeded:
 			log.Debugf("Pod [ %s ] completed", podName)
 			return nil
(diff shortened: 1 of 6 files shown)


pkg/apb/watch_pod.go
```diff
@@ -91,3 +105,16 @@ func errorPullingImage(conds []apiv1.ContainerStatus) bool {
 
 	return false
 }
+
+func getExitCode(conds []apiv1.ContainerStatus) int32 {
```
maleck13: Just a suggestion. But as we know something went wrong because the pod failed, would it be clearer to have something like ```translateExitStatus(podName,  []apiv1.ContainerStatus)error ``` so that line 62 would become: 

```
return translateExitStatus(podName, podSatus.ContainerStatuses)
```
And the logic around exit codes would be contained to the single function. It would also avoid using the exit codes -120 and -127

pkg/clients/kubernetes.go
```diff
@@ -184,12 +184,6 @@ func GetSecretData(secretName, namespace string) (map[string][]byte, error) {
...
 	}
-	secretData, err := k8scli.Client.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
-	if err != nil {
-		log.Errorf("Unable to load secret '%s' from namespace '%s'", secretName, namespace)
-		return make(map[string][]byte), nil
-	}
-	log.Debugf("Found secret with name %v\n", secretName)
 
-	return secretData.Data, nil
+	return k8scli.GetSecretData(secretName, namespace)
```
maleck13: As you mentioned in IRC it is strange to have 2 of these functions. Think its worth getting rid of the duplication as part of a separate issue?

jmrodri: @maleck13 yeah I think a separate PR will be better to fix this duplication. I can do that.

jmrodri:
I guess I need to create this PR against release-1.1 :)

pkg/apb/watch_pod.go
jmrodri: @maleck13 excellent suggestion, thanks.

pkg/apb/watch_pod.go
```diff
@@ -55,6 +57,18 @@ func watchPod(podName string, namespace string) error {
 			if errorPullingImage(podStatus.ContainerStatuses) {
 				return ErrorPodPullErr
 			}
+
+			// handle the return code from the pod
+			exitCode := getExitCode(podStatus.ContainerStatuses)
+			if exitCode == 8 {
+				log.Errorf("Pod [ %s ] failed - action not found.", podName)
```
rthallisey: There might be a way to make this more informative. The caller of watchPod is one of the actions (bind, undbind, provision, deprovision).  So the caller knows what action was performed.  Inside watchPod, we know the caller's action failed. We don't know what it is, but we do know the playbook for that action doesn't exist if the exitCode is 8.

A better way to do this might be to return something like ```Pod [ %s ] failed - the action's playbook was not found```. Then at the action level, we change ```log.Errorf("APB Execution failed - %v", err)``` to identify the action and bubble up the error.  For bind, this could look like  ```log.Errorf("Bind action failed - %v", err)```

What do folks think?

maleck13: This sounds reasonable to me. More context for the end user the better. I would actually suggest moving to https://github.com/pkg/errors for this kind of thing. It is very useful.
Has a method
```
errors.Wrap(err,"some additional context")
```
that returns an error. It also provides the benefit of providing stack traces when the errors are logged with ```%+v```

shawn-hurley: @maleck13 I think we should discuss moving to pkg/errors in a proposal I have created to #717 track this change. I agree that this package does help in debugging errors and is super helpful.

eriknelson: It sounds great, @shawn-hurley it does sound like a broader convention we should adopt and talk about how we want to get there.

jmrodri:
@maleck13 and @rthallisey I updated the PR based off both of your suggestions. Let me know what you think. I'm going to post another PR against release-1.1 for this exact same bug.

Issue #685, opened 2018-01-25
Give more respect to the whitelist param


rthallisey:
# Bug
The issue https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/issues/215 pointed out that regardless of whitelist value, we always expect the suffix ```-apb``` in order for the apb to be seen by the ansible-service-broker. The relevant code is [here](https://github.com/openshift/ansible-service-broker/blob/master/pkg/registries/registry.go#L165).

I think it's worth to consider giving more power to the whitelist param.  I think it will be less error prone if the user had full control even though it's likely they will use the suffix ```-apb```. 

We can take on this issue as part of the working going into https://github.com/openshift/ansible-service-broker/issues/684

eriknelson:
This dates all the way back to CAP actually; IIRC, when we were first investigating how to do discovery, there were some limitations with the dockerhub API where the process was to query a repo for a set of matching image names, then pull those images. Obviously we didn't want to pull all the images to check to see if they were an APB, so we made a decision and said "an atomicapp ends with -atomic" (this was when things were called atomicapps). That let us make sure we weren't doing throwaway work.

Now, obviously we've learned a lot. Particularly, there should be an API that let's us pull and inspect the metadata without downloading the image. If an APB label is on that image, we can trust that it's an APB and include it in the set of image names to be filtered by the white/black lists.

IMO, we should remove the function that hard enforced a "-apb" suffix, and update the adapters to inspect the metadata and only report ImageNames that contain the label. It may require some shifting around of the `Adapter` interface, but that's how I'd do it. Hopefully we can get away from imposing a name convention on users; that's kind of ugly.

rthallisey:
I'm still seeing images that don't have the *-apb suffix being processed when I have the whitelist set to only filter *-apb images.

rthallisey:
https://github.com/openshift/ansible-service-broker/pull/781

PR #719, opened 2018-02-01
No longer only search for apbs that end with -apb

rthallisey:
Give full respect to the whitelist config value by no longer filtering for '-apb'.

 - Remove hardcoded filter on the suffix '-apb'

fixes https://github.com/openshift/ansible-service-broker/issues/685

### DIFF
--- a/pkg/registries/registry.go
+++ b/pkg/registries/registry.go
@@ -111,8 +111,6 @@ func (r Registry) LoadSpecs() ([]*apb.Spec, int, error) {
 			r.config.Name, err)
 		return []*apb.Spec{}, 0, err
 	}
-	// Registry will throw out all images that do not end in -apb
-	imageNames = registryFilterImagesForAPBs(imageNames)
 	validNames, filteredNames := r.filter.Run(imageNames)
 
 	log.Debug("Filter applied against registry: %s", r.config.Name)
@@ -157,16 +155,6 @@ func (r Registry) LoadSpecs() ([]*apb.Spec, int, error) {
 	metrics.SpecsLoaded(r.RegistryName(), len(validatedSpecs))
 
 	return validatedSpecs, len(imageNames), nil
-}
-
-func registryFilterImagesForAPBs(imageNames []string) []string {
-	newNames := []string{}
-	for _, imagesName := range imageNames {
-		if strings.HasSuffix(strings.ToLower(imagesName), "-apb") {
-			newNames = append(newNames, imagesName)
-		}
-	}
-	return newNames
 }
 
 // Fail - will determine if the registry should cause a failure.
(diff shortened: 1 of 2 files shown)


rthallisey:
Turns out we do skip if we don't find the apb label.
```
        if conf.Config.Label.Spec == "" {
                log.Infof("Didn't find encoded Spec label. Assuming image is not APB and skiping")
                return nil, nil
        }
```

jmrodri:
What was the premise behind this change? Do we know if there is any performance impact with this?
Can we use the white list and just add another name to include? This seems to cause my spidey senses to tingle but I may be over reacting.

jmrodri:
Just using the ansibleplaybookbundle repo, I don't see any time difference in bootstrapping the broker.
With this change it took 5.674 seconds.
```
[2018-02-01T21:47:31.829Z] [INFO] - Attempting bootstrap...
[2018-02-01T21:47:37.503Z] [NOTICE] - Broker successfully bootstrapped on startup
```

Using the code in master, bootstrap took: 6.27 seconds
```
[2018-02-01T21:53:25.571Z] [INFO] - Attempting bootstrap...
[2018-02-01T21:53:31.841Z] [NOTICE] - Broker successfully bootstrapped on startup
```

I only did one run and captured the logs for each. Surprisingly this change came in faster by 0.596 seconds. :)

jmrodri:
I'd like for us to look at cleaning up our logging for bootstrapping to make it more organized, possibly for a later PR.

jmrodri:
Would it be bad to make this behavior configurable? that is disable the -apb check by default, but allow someone to turn it on with a flag? I mean this has served us well so far while removing it will solve one use case, I can see someone wanting to keep the original behavior.

maleck13:
Forgive the lack of knowledge here but does the white_list not provide them with this functionality?

rthallisey:
> Would it be bad to make this behavior configurable? that is disable the -apb check by default, but allow someone to turn it on with a flag? I mean this has served us well so far while removing it will solve one use case, I can see someone wanting to keep the original behavior.
>
It's already configurable with whitelist and the default behaviour is identical to the filter I'm removing. https://github.com/openshift/ansible-service-broker/blob/master/templates/deploy-ansible-service-broker.template.yaml#L311-L312.

What I'm doing with this patch is removing the redundancy because we are filtering in two places.  Whitelist should work as advertised and filter on the regex the user expects.  Right now, that's not true because we always filter on the suffix -apb in addition to what's in whitelist.

@jmrodri There's some additional information in the issue https://github.com/openshift/ansible-service-broker/issues/685 if I didn't address your concern. Let me know if you have any more comments/concerns.

jmrodri:
@rthallisey thanks for that clarification. This makes a lot of sense now.

PR #722, opened 2018-02-02
Initial prop for custom resource definitions

shawn-hurley:
Proposal for using custom resource definitions.

### DIFF
--- a/docs/proposals/custom_resources_definitions.md
+++ b/docs/proposals/custom_resources_definitions.md
@@ -1,0 +1,22 @@
+## Proposal: Move To Custom Resources Definitions For Persisting Data
+The slides that describe the problem in more detail can be found [here](https://docs.google.com/presentation/d/1UQU4BWlLGw70KQEwqNTZghxQVl0dTzyWLGtQ9z1DLx8/edit?usp=sharing).
+
+### Problem Description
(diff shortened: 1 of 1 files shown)


maleck13:
@shawn-hurley @eriknelson We use client generation for a CRD that we use with the CLI we are creating. In my experience it saves quite a lot of time (although the initial setup is a cost) as it generates a full client with a very similar API to the kubernetes go-client. It also generates all the fakes that you need to allow simple testing when using the client.
If you are interested here is our definition:
https://github.com/aerogear/mobile-cli/tree/master/pkg/apis/mobile
The client it generates lives here:
https://github.com/aerogear/mobile-cli/tree/master/pkg/client
And here is an example of the fakes that allow good unit tests
https://github.com/aerogear/mobile-cli/tree/master/pkg/client
Finally here is the make target to generate the clients
https://github.com/aerogear/mobile-cli/blob/master/scripts/generate.sh

shawn-hurley:
I am adding my two sense here because I realize I did not advocate for one way or the other in the proposal. 

I agree that for the first pass that we need to keep these for internal use only. I think soon after we may have to start dealing with cases where things change underneath us.

The client I think is going to start easy with basic CRUD and get much more involved. I wouldn't mind taking this approach first and reevaluating latter if the maintenance of it becomes a pain. 

I think that in the end, we will not want to create and manage listers/watchers and the rest when it comes to the client. A full client is actually pretty involved, and I think there might be a case for making a kube client look like other kube clients. 

No matter where the client lives I think it makes sense to keep it isolated so that if we need to it should be easily ripped out.

PR #723, opened 2018-02-02
Bug 1541265 - avoid broker panic, check for nil

jmrodri:
Ensure we check the extCred for nil.


eriknelson:
Thank you @jmrodri !

PR #724, opened 2018-02-02
Bug 1541265 - avoid broker panic, check for nil

jmrodri:
The apb.Update call can return nil for extCreds. This is a valid use
case, we need to check nil objects before trying to dereference them.


eriknelson:
Woops, thought I ack'd this but didnt.

PR #725, opened 2018-02-02
Dao Interface

### LINKED ISSUE #542: Proposal: using interfaces for DAO operations to allow simple unit tests

shawn-hurley:
Having an interface for the DAO will allow us to switch persistent layers. This will also allow for unit testing in other parts of the code base.

 - move etcd dao into dao/etcd package
 - everything should now use the DAO interface

depends-on <PR>
fix for PR #722 Work Item 1
This fixes #542

### DIFF
--- a/pkg/dao/dao.go
+++ b/pkg/dao/dao.go
@@ -17,407 +17,83 @@
 package dao
 
 import (
-	"context"
-	"fmt"
-	"strings"
-
-	"github.com/coreos/etcd/client"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
-	"github.com/openshift/ansible-service-broker/pkg/clients"
+	etcd "github.com/openshift/ansible-service-broker/pkg/dao/etcd"
 	logutil "github.com/openshift/ansible-service-broker/pkg/util/logging"
-	"github.com/pborman/uuid"
 )
 
 var log = logutil.NewLog()
 
-// Dao - object to interface with the data store.
-type Dao struct {
-	client client.Client
-	kapi   client.KeysAPI // Used to interact with kvp API over HTTP
-}
-
 // NewDao - Create a new Dao object
-func NewDao() (*Dao, error) {
-	dao := Dao{}
-
-	etcdClient, err := clients.Etcd()
-	if err != nil {
-		return nil, err
-	}
-	dao.client = etcdClient
-	dao.kapi = client.NewKeysAPI(dao.client)
-	return &dao, nil
+func NewDao() (Dao, error) {
+	return etcd.NewDao()
 }
 
-// SetRaw - Allows the setting of the value json string to the key in the kvp API.
-func (d *Dao) SetRaw(key string, val string) error {
-	_, err := d.kapi.Set(context.Background(), key, val /*opts*/, nil)
-	return err
-}
-
-// GetRaw - gets a specific json string for a key from the kvp API.
-func (d *Dao) GetRaw(key string) (string, error) {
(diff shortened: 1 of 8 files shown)


shawn-hurley:
@maleck13 Sorry I am not fully following your recommendation regarding interfaces. 

What I am seeing is `dao` is the consumer of `dao/etcd`(the provider) and therefore defines the interface. 

Is that different then what you are referring to?

maleck13:
Sorry @shawn-hurley  I will give a more concrete  example of what I mean.

The broker package depends on the DAO. So in my suggestion the broker package would now define the interface that it needs as it knows its requirements best and so would no longer depend on
``` "github.com/openshift/ansible-service-broker/pkg/dao" ```
as the interface is defined within its own package. Does that make what I mean clearer. As mentioned this is only a suggestion

**Comment by shawn-hurley** [comment]
Ah ok, that does make more sense now thank you.

I would like to hear more about that helps prevent circular dependencies. 
Also, is their an example in the standard library of using the interface like this? I am only asking because that might help me match use cases?

File: pkg/dao/dao.go [hunk 1]
```diff
@@ -17,407 +17,83 @@
...
-	for i, payload := range *payloads {
-		spec := &apb.Spec{}
-		apb.LoadJSON(payload, spec)
-		specs[i] = spec
-		log.Debug("Batch idx [ %d ] -> [ %s ]", i, spec.ID)
-	}
-
-	return specs, nil
+func NewDao() (Dao, error) {
+	return etcd.NewDao()
```
jmrodri: Is the idea we'll have multiples from here?

shawn-hurley: Yep, that was the initial plan/thought process

jmrodri:
@eriknelson @shawn-hurley as far as names, the only thing that is descriptive is DAL since it is a data access layer vs DAO which usually means data access object.

jmrodri:
@eriknelson @shawn-hurley correction I do have a name suggestion: AXON

http://www.dictionary.com/browse/axon

> the long threadlike part of a nerve cell along which impulses are conducted from the cell body to other cells.

jmrodri:
Are we holding off because of naming? or are we good?

maleck13:
@shawn-hurley hope this is an interesting discussion ( I think so) but not my intention to hijack the PR :)
**TLDR**
- It is not a hard black and white rule. 
- You could refactor a package consuming an interface to declare its own version easily if it became and issue
- It is really just a pattern I have found useful and to work well

**Standard Lib:**
I haven't gone hunting through the standard library to check, however i know the ```io``` package declares the ```Writer``` and ```Reader```  interface and consumes them heavily throughout that package. 
So in my opinion, the io package knows what it wants in order to consume a reader or a writer, so it is the package that declares the interfaces. However in this case, it is an exported interface as it is generic enough that it can be useful outside of the io package. If it was not useful outside of the io package I don't think it would be exported and would just be an internal interface. There are some of these in the io package also.

On the circular dependencies: (this really is less of a thing but a side benefit)
By declaring the interface within the consuming package you reduce the the chance of a circular dependency as you have removed an import of an external package.

PR #727, opened 2018-02-04
Initial pass at proposal to change subscribers

### LINKED ISSUE #638: Proposal: replace the current subscriber pattern
# Proposal

Currently we use a subscriber pattern where as a Job is started and is given a channel to send messages on, there is then a subscriber to this channel which uses the DAO to update state based on the msg sent.
While this works, I wanted to open a constructive discussion about potential alternatives that may simplify and improve on this.

**Some thoughts on this approach:**

- As there is only one subscriber to this channel, it causes a natural bottleneck. Once a Job has sent a message to the channel, no other work can be done nor can state be saved until after the subscriber has completed its work. Currently this work is all synchronous so only one Job state will be saved at any given time and all other Jobs of the same type will block waiting to send their messages through the channel until it is complete.
- There seems to be very little benefit in the current subscribers. The Jobs know the state they are in and also have access to the credentials sending this information through a channel to be saved seems overly complex.

**Alternatives**

- Pass an interface that exposes the required methods of the DAO to the Jobs allowing the jobs to make the required state changes. As these jobs are running within their own go routine and no longer using the channel, it would remove any potential to impact on other Jobs while simplifying the code.

- Add a set of Observers to each job that are passed the JobState as it changes, these Observers could ensure we keep the responsibilities decoupled.

- metrics
- logging
- state update

Sudo code:

```
NewProvisionJob(..., observers []ProvisionStateObserver){}

(p *ProvisionJob)Run func(...){
   js := JobState{}
   for _, ob := range p.observers{

      go ob(js)

   }

}

```

The above code may not be clear enough. If there is appetite for either of these approaches I am happy to create a more detailed proposal.

@jmrodri @rthallisey based on comments in other PRs created this issue to reflect some of my thoughts on the subscribers. Hope it makes sense.

maleck13:
Proposal based on conversations with the team about the current subscriber pattern and some of its weaknesses. 
This proposal is fairly complete in my opinion but keen to get feedback on the approaches outlined and to ensure I am fulfilling everyone's expectations around this change.

fixes #638

### DIFF
--- a/docs/proposals/subscriber-fan-in-fan-out.md
+++ b/docs/proposals/subscriber-fan-in-fan-out.md
@@ -1,13 +1,13 @@
-# Switch to using Fan in Fan out / Observer pattern for subscribers
+# Switch to using Fan-in & Fan-out / Observer pattern for subscribers
 
-This is not quite the same as a fan in fan out pattern but not a
+This is not quite the same as a fan-in / fan-out out pattern but not a
 million miles away either.
 
 ## Terminology
 
 **Observer:** suggestion to change the name from subscriber to observer
 as I think it more clearly conveys what they will be doing.
-For this doc I will continue to call them subscribers to avoid
+For this doc, I will continue to call them subscribers to avoid
 confusion.
 
 ## Problem Description
@@ -17,31 +17,31 @@ Our current subscriber pattern has some issues that have been outlined in the fo
 To surmise:
 
 - **Shared Channel with single subscriber**: There is a single buffered channel per topic (topics being broker
-actions such as provision, bind etc). All jobs of the given type write to
+actions such as provision, bind etc). All jobs, of the given type, write to
 this channel. This has the potential to cause a bottleneck, particularly
 as there is currently only a single subscriber reading from this channel and
 this subscriber does not read the next message until it has completed
-all of its work. Currently there is not too much happening in these
+all of its work. Currently, there is not too much happening in these
 subscribers, but if the work done in these subscribers were to increase
-or become computationally complex, it could begin to slow down throughput
+or become computationally complex, it could begin to slow down the throughput
 of messages from the jobs and potentially block them. The likely hood
 of this happening has been increased by the fact that we will soon allow
-fine grained status updates to sent by the APB developer via last operation
+fine-grained status updates to sent by the APB developer via last operation
 updates.
 
 - **Separation of concerns**: While adding more subscribers would alleviate
 the previous issue, the current pattern means only one subscriber can
 deal with a given message at any one time. This makes it difficult to have a
-separation of concerns.  Ideally we would be able to have each subscriber
+separation of concerns.  Ideally, we would be able to have each subscriber
 responsible for doing just one thing (recording metrics adding various
 logs, persisting state etc all based on the msg it had been passed). This
-would make it simple to add more functionality based around these messages
+would make it simple to add more functionality based on these messages
 along with making testing simple and isolated.
 
 
 ## Goal
 
-- Outline a design for passing messages, both user generated and system
+- Outline a design for passing messages, both user-generated and system
 generated from executing APB jobs to each of a set of registered subscribers
 that removes the outlined problems and improves the overall design of the broker.
 
@@ -49,18 +49,18 @@ that removes the outlined problems and improves the overall design of the broker
 
 This proposal outlines changes to the subscribers and allowing
 the addition of more subscribers allowing them all to act on the same message.
-It is a key principal that the order these subscribers are called in
-should not matter, their work should be independent of any other subscribers work.
+The order these subscribers are called in should not matter, their work
+should be independent of any other subscribers work.
 
 ## Proposal
 
 
 ### Change the subscriber interface
 
-Currently the subscribers implement the WorkSubscriber interface, we
+Currently, the subscribers implement the WorkSubscriber interface, we
 would change this interface and the current subscribers.
 Rather than passing the channel into the subscriber, we would rename and
-change this interface to accept a ```JobMsg```, value not pointer to stop
+change this interface to accept a ```JobMsg```, value not a pointer to stop
 unexpected mutation, and return an error.
 The current subscribers would handle state persistence and would be
 renamed to reflect their role.
@@ -75,9 +75,9 @@ type WorkObserver interface{
 
 #### Change how subscribers are referenced
 
-Currently when a subscriber is added, we take the existing channel or
+Currently, when a subscriber is added, we take the existing channel or
 create one if not present and hand that channel to the subscriber.
-Instead we would now have a registry of subscribers to topics:
+Instead, we would now have a registry of subscribers to topics:
 
 ```go
 type WorkEngine struct {
@@ -95,7 +95,7 @@ any blocking:
 
 #### 1) Keep the single channel but remove potential for blocking
 
-Currently the topic channel is handed to the single subscriber which
+Currently, the topic channel is handed to the single subscriber which
 performs all the work. As mentioned this means no more messages can be
 read from the channel until this unit of work is complete. A change to
 how the messages are handled could alleviate any potential issue. The
@@ -123,7 +123,7 @@ for{
 The stop channel here would be tied to os signals
 [example](https://gist.github.com/reiki4040/be3705f307d3cd136e85).
 
-In this case we would still have a single channel per topic, but as we
+In this case, we would still have a single channel per topic, but as we
 hand off the message immediately and don't block waiting for the
 subscribers to finish their work, there is little to no chance for the
 channel to become full and so block the running jobs.
@@ -132,7 +132,7 @@ channel to become full and so block the running jobs.
 
 The work engine will now be the one reading messages from the channel
 and passing it on to the Observers. To do this I propose we add a  public
-``Start(stop <-chan struct{})`` method, that would be called in it's own
+``Start(stop <-chan struct{})`` method, that would be called in its own
 go routine during ASB start up. I also propose adding a signal channel
 that will send a message to anything running  in the background
 (could also use context.Context). This would signal for the background
@@ -143,10 +143,10 @@ above and close them down once a message was received.
 #### 2) Create a channel per job
 
 Remove the single channel per topic and replace it with a channel per
-job. The work engine would be updated to create a channel per Job,
+job. The work engine would be updated to create a channel per Job
 when starting a new job. The work engine would be responsible for
 the lifecycle of these channels. As above, the subscribers would be
-registered in a map and called in a non blocking way when a message was received.
+registered in a map and called in a non-blocking way when a message was received.
 While more complex, this method would allow us to leverage more control
 (use a buffer to control how much could be sent while still
 not impacting on any other Jobs that running).
@@ -186,4 +186,4 @@ many thousands of go routines without issue.
 
 ## Considerations
 
-- perhaps want a config option to limit the number of in progress jobs
+- perhaps want a config option to limit the number of in-progress jobs
--- a/docs/proposals/subscriber-fan-in-fan-out.md
+++ b/docs/proposals/subscriber-fan-in-fan-out.md
@@ -1,0 +1,189 @@
+# Switch to using Fan in Fan out / Observer pattern for subscribers
+
+This is not quite the same as a fan in fan out pattern but not a
+million miles away either.
+
+## Terminology
+
+**Observer:** suggestion to change the name from subscriber to observer
+as I think it more clearly conveys what they will be doing.
+For this doc I will continue to call them subscribers to avoid
+confusion.
+
+## Problem Description
+
+Our current subscriber pattern has some issues that have been outlined in the following
+[github issue](https://github.com/openshift/ansible-service-broker/issues/638)
+To surmise:
+
+- **Shared Channel with single subscriber**: There is a single buffered channel per topic (topics being broker
+actions such as provision, bind etc). All jobs of the given type write to
+this channel. This has the potential to cause a bottleneck, particularly
+as there is currently only a single subscriber reading from this channel and
+this subscriber does not read the next message until it has completed
+all of its work. Currently there is not too much happening in these
+subscribers, but if the work done in these subscribers were to increase
+or become computationally complex, it could begin to slow down throughput
+of messages from the jobs and potentially block them. The likely hood
+of this happening has been increased by the fact that we will soon allow
+fine grained status updates to sent by the APB developer via last operation
+updates.
+
+- **Separation of concerns**: While adding more subscribers would alleviate
+the previous issue, the current pattern means only one subscriber can
+deal with a given message at any one time. This makes it difficult to have a
+separation of concerns.  Ideally we would be able to have each subscriber
+responsible for doing just one thing (recording metrics adding various
+logs, persisting state etc all based on the msg it had been passed). This
+would make it simple to add more functionality based around these messages
+along with making testing simple and isolated.
+
+
+## Goal
+
+- Outline a design for passing messages, both user generated and system
+generated from executing APB jobs to each of a set of registered subscribers
+that removes the outlined problems and improves the overall design of the broker.
+
+## Subscriber design principal
+
+This proposal outlines changes to the subscribers and allowing
+the addition of more subscribers allowing them all to act on the same message.
+It is a key principal that the order these subscribers are called in
+should not matter, their work should be independent of any other subscribers work.
+
+## Proposal
+
+
+### Change the subscriber interface
+
+Currently the subscribers implement the WorkSubscriber interface, we
+would change this interface and the current subscribers.
+Rather than passing the channel into the subscriber, we would rename and
+change this interface to accept a ```JobMsg```, value not pointer to stop
+unexpected mutation, and return an error.
+The current subscribers would handle state persistence and would be
+renamed to reflect their role.
+ 
+```go
+#Note I call it observer here see terminology
+type WorkObserver interface{
+   Notify(msg JobMsg)error
+}
+```
+
+
+#### Change how subscribers are referenced
+
+Currently when a subscriber is added, we take the existing channel or
+create one if not present and hand that channel to the subscriber.
+Instead we would now have a registry of subscribers to topics:
+
+```go
+type WorkEngine struct {
+	subscribers map[WorkTopic][]WorkSubscriber
+	...
+} 
+
+```
+
+
+
+### Channel changes
+There a couple of options available as the goal here is to avoid
+any blocking:
+
+#### 1) Keep the single channel but remove potential for blocking
+
+Currently the topic channel is handed to the single subscriber which
+performs all the work. As mentioned this means no more messages can be
+read from the channel until this unit of work is complete. A change to
+how the messages are handled could alleviate any potential issue. The
+reader would change to be the work engine which would start a read loop
+for each of the topics when the broker started and distribute incoming
+messages to any number of subscribers async:
+As long as the principal that subscribers shouldn't depend on each other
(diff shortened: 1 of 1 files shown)


maleck13:
@eriknelson @jmrodri initial proposal based on our conversations.

jmrodri:
@maleck13 would it be possible to draw a picture of what it will look like? Even a picture on paper would work. Otherwise, I'll look at it tomorrow and see if I can get a picture done.

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
+# Switch to using Fan in Fan out / Observer pattern for subscribers
```
jmrodri: Fan in Fan out -> Fan-in & Fan-out

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+
+This is not quite the same as a fan in fan out pattern but not a
```
jmrodri: fan in fan out -> fan-in / fan-out

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+million miles away either.
+
+## Terminology
+
+**Observer:** suggestion to change the name from subscriber to observer
+as I think it more clearly conveys what they will be doing.
+For this doc I will continue to call them subscribers to avoid
```
jmrodri: comma after doc "For this doc, I "

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+confusion.
+
+## Problem Description
+
+Our current subscriber pattern has some issues that have been outlined in the following
+[github issue](https://github.com/openshift/ansible-service-broker/issues/638)
+To surmise:
+
+- **Shared Channel with single subscriber**: There is a single buffered channel per topic (topics being broker
+actions such as provision, bind etc). All jobs of the given type write to
```
jmrodri: add commas "All jobs, of the given type, write to this channel"

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+this channel. This has the potential to cause a bottleneck, particularly
+as there is currently only a single subscriber reading from this channel and
+this subscriber does not read the next message until it has completed
+all of its work. Currently there is not too much happening in these
```
jmrodri: comma missing after Currently

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+subscribers, but if the work done in these subscribers were to increase
+or become computationally complex, it could begin to slow down throughput
```
jmrodri: change: slow down throughput -> slow down the throughput

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+of messages from the jobs and potentially block them. The likely hood
+of this happening has been increased by the fact that we will soon allow
+fine grained status updates to sent by the APB developer via last operation
```
jmrodri: TYPO: fine grained -> fine-grained

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+updates.
+
+- **Separation of concerns**: While adding more subscribers would alleviate
+the previous issue, the current pattern means only one subscriber can
+deal with a given message at any one time. This makes it difficult to have a
+separation of concerns.  Ideally we would be able to have each subscriber
```
jmrodri: comma after Ideally

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+responsible for doing just one thing (recording metrics adding various
+logs, persisting state etc all based on the msg it had been passed). This
+would make it simple to add more functionality based around these messages
```
jmrodri: replaced around with on. "based on these messages"

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+along with making testing simple and isolated.
+
+
+## Goal
+
+- Outline a design for passing messages, both user generated and system
```
jmrodri: TYPO: user generated -> user-generated

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+generated from executing APB jobs to each of a set of registered subscribers
+that removes the outlined problems and improves the overall design of the broker.
+
+## Subscriber design principal
+
+This proposal outlines changes to the subscribers and allowing
+the addition of more subscribers allowing them all to act on the same message.
+It is a key principal that the order these subscribers are called in
+should not matter, their work should be independent of any other subscribers work.
```
jmrodri: Let's reword that last sentence, remove "it is a key principal that "

The order subscribers are called should not matter, their work should be independent of any other subscribers.

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Proposal
+
+
+### Change the subscriber interface
+
+Currently the subscribers implement the WorkSubscriber interface, we
+would change this interface and the current subscribers.
+Rather than passing the channel into the subscriber, we would rename and
+change this interface to accept a ```JobMsg```, value not pointer to stop
```
jmrodri: missing an a. "not a pointer"

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+the addition of more subscribers allowing them all to act on the same message.
+It is a key principal that the order these subscribers are called in
+should not matter, their work should be independent of any other subscribers work.
+
+## Proposal
+
+
+### Change the subscriber interface
+
+Currently the subscribers implement the WorkSubscriber interface, we
```
jmrodri: comma after Currently

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+#Note I call it observer here see terminology
+type WorkObserver interface{
+   Notify(msg JobMsg)error
+}
+```
+
+
+#### Change how subscribers are referenced
+
+Currently when a subscriber is added, we take the existing channel or
```
jmrodri: comma after Currently

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+create one if not present and hand that channel to the subscriber.
+Instead we would now have a registry of subscribers to topics:
```
jmrodri: comma after Instead

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+
+### Channel changes
+There a couple of options available as the goal here is to avoid
+any blocking:
+
+#### 1) Keep the single channel but remove potential for blocking
+
+Currently the topic channel is handed to the single subscriber which
```
jmrodri: comma after Currently

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+   }
+}
+
+
+```
+
+The stop channel here would be tied to os signals
+[example](https://gist.github.com/reiki4040/be3705f307d3cd136e85).
+
+In this case we would still have a single channel per topic, but as we
```
jmrodri: comma after "In this case"

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+hand off the message immediately and don't block waiting for the
+subscribers to finish their work, there is little to no chance for the
+channel to become full and so block the running jobs.
+
+#### Add work engine start method
+
+The work engine will now be the one reading messages from the channel
+and passing it on to the Observers. To do this I propose we add a  public
+``Start(stop <-chan struct{})`` method, that would be called in it's own
```
jmrodri: TYPO: it's -> its

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+that will send a message to anything running  in the background
+(could also use context.Context). This would signal for the background
+loops etc to stop.
+This start method would start a loop for all of the topics as shown
+above and close them down once a message was received.
+
+#### 2) Create a channel per job
+
+Remove the single channel per topic and replace it with a channel per
+job. The work engine would be updated to create a channel per Job,
```
jmrodri: remove the comma after Job

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+when starting a new job. The work engine would be responsible for
+the lifecycle of these channels. As above, the subscribers would be
+registered in a map and called in a non blocking way when a message was received.
```
jmrodri: TYPO: non blocking -> nonblocking

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+
+Each job that was started would create approx 5 go routines and 1 channel.
+I don't have too much concern here as go is designed to be able to spawn
+many thousands of go routines without issue.
+
+
+
+## Considerations
+
+- perhaps want a config option to limit the number of in progress jobs
```
jmrodri: TYPO: in progress -> in-progress

maleck13:
@jmrodri Thanks for the quick response ! I will fix up the typos etc. Happy to do a picture, but I am mostly unavailable today and tomorrow due to training. 
I believe however it reasonably close to @eriknelson drawing
https://user-images.githubusercontent.com/410933/35408136-021411f2-01dc-11e8-8cd9-0464766d19da.jpg
Except the engine is also the distributor. Rough attempt here.
image

maleck13:
updated based on comments

rthallisey:
@jmrodri @eriknelson any more comments here?

eriknelson:
On my review stack!

eriknelson:
@maleck13 This is pedantic, but is there a semantic difference between the pattern language of `Observer - Subject` vs. `Publisher - Subscriber`? I'm entirely indifferent between the two, aside from the fact that there is precedence for `Subscriber`.

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+channel to become full and so block the running jobs.
+
+#### Add work engine start method
+
+The work engine will now be the one reading messages from the channel
+and passing it on to the Observers. To do this I propose we add a  public
+``Start(stop <-chan struct{})`` method, that would be called in its own
+go routine during ASB start up. I also propose adding a signal channel
+that will send a message to anything running  in the background
+(could also use context.Context). This would signal for the background
```
shawn-hurley: +1 to the use of context.

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+registered in a map and called in a non-blocking way when a message was received.
+While more complex, this method would allow us to leverage more control
+(use a buffer to control how much could be sent while still
+not impacting on any other Jobs that running).
+
+Sudo Code:
+
+```go 
+
+StartNewJob(token string, work Work, topic WorkTopic){
```
shawn-hurley: Just for my own clarification, the var `a` in the Sudo Code is for the work engine itself.

so from this, we see that a keeps track of all of the channels per a token, and keeps track of all of the subscribers for a given topic. Just want to make sure there is no other concerns that the engine is keeping track of.

Also, do we need to think about passing a context around here? I envision a use case where a subscriber says hey some really odd happened here and needs to send a message to other subscribers. I believe that is the use case for context?

maleck13: Yes those are the two things that would need to be tracked by the engine, If we were to use a channel per job.

On context, it is normally used to signal timeout and cancellations. I would not be against accepting a context as part of a Jobs definition to enable cancellation or potentially timeouts. 
Context can also carry values across API lines, but I am not such a fan of this approach.

maleck13:
@eriknelson Happy to leave them as subscribers, really just a suggestion as I like the term observer in the context of the Job. They are observing / watching the job. I agree though, I don't think there is really much difference and I can see that channel is essentially a message bus

docs/proposals/subscriber-fan-in-fan-out.md
shawn-hurley: I think adding the context for timeout is something that we should do.

I think that canceling a running subscriber if another subscriber fails is something we should consider. 

@eriknelson @jmrodri do you guys have thoughts on timeouts or cancellations?

maleck13:
I believe the actions on me here are to 
- Add a diagram showing the channel per job approach
- Remove the change to the name of the subscribers

Remaining is a discussion about passing a context in to Job to allow for timeout and cancellation. (should this be a separate proposal perhaps)?

Am I missing anything?

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+messages to any number of subscribers async:
+As long as the principal that subscribers shouldn't depend on each other
+is maintained there shouldn't be issues with this approach.
+```go 
+for{
+
+   select{
+       case msg := <-provisionChan:
+           for _,s := range e.subscribers[ProvisionTopic]{
+                go s(msg)
```
eriknelson: There's a bit of an inconsistency here I think. (I'm going to use `Subscriber` here for the sake of clarity, assume it is synonymous with `Observer`). If we are moving away from subscriber's taking a channel as an argument with `   Notify(msg JobMsg)error`, this should be `s.Notify(msg)`. Additionally, this is fire-and-forget; the engine doesn't have an opportunity to deal with errors that may come back. That may be desirable for this soln? The question I would ask is:

* Is it okay to just fire these off and ignore the outcome? If so, it doesn't make sense to return an error from `Notify`. How do we handle error scenarios in the subscribers then, if at all?
* If it's not okay, how do we expect the engine to handle errors that have been returned from subscribers?

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+This start method would start a loop for all of the topics as shown
+above and close them down once a message was received.
+
+#### 2) Create a channel per job
+
+Remove the single channel per topic and replace it with a channel per
+job. The work engine would be updated to create a channel per Job
+when starting a new job. The work engine would be responsible for
+the lifecycle of these channels. As above, the subscribers would be
+registered in a map and called in a non-blocking way when a message was received.
```
eriknelson: Is the expectation that there will be unique instances of registered subscriber types, per job? I.E., if I have an `EtcdSub`, and a `RabbitSub`, the work engine will instantiate unique ones for each job? Or each **topic**?

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+StartNewJob(token string, work Work, topic WorkTopic){
+  a.jobs[token] = make(chan JobMsg)
+  go func(){
+      //make sure we always close the channel  
+      defer close(e.jobs[token])
+      // start reading from the channel 
+      go func(){
+      // will auto stop once the channel is closed
+        for msg := range a.jobs[token]{
+            for _, s := range a.subscribers[topic]{
```
eriknelson: Can you confirm my understanding of this? It appears that there is a single instance of a given subscriber for each **topic** in this model, where there is only one topic per action. The subscribers listening on a given topic have their `Notify` methods triggered when a `JobMsg` comes in. It's spun off immediately into a go routine? Begs the question, why are subscribers structs at all, if the same instance is triggered for all jobs. Seems to me it should just be a stateless handler function.

docs/proposals/subscriber-fan-in-fan-out.md
```diff
@@ -0,0 +1,189 @@
...
+               // Note we could choose to run these sync but again if there was a slow
+               // subscriber it would block receiving on the channel.
+               go s(msg)
```
eriknelson: Same comment WRT `Notify` and what we should do about errors.

docs/proposals/subscriber-fan-in-fan-out.md
eriknelson: > I think that canceling a running subscriber if another subscriber fails is something we should consider.

My gut feeling is that they're intended to take a simple data structure that contains the result of some work and do something useful with it completely independent from other subscribers. I'm not sure I like the idea of them communicating to each other.

eriknelson:
We need a consensus on whether we'd like to pursue implementation 1 or 2. Is it fair to say 2 is preferred?

docs/proposals/subscriber-fan-in-fan-out.md
maleck13: I think as they are intended to be stateless functions, they would be registered at startup for each topic. I hadn't intended for them to be created per job just called per job. I will make this clearer in the proposal

docs/proposals/subscriber-fan-in-fan-out.md
maleck13: Your understanding here is correct and inline with what I was thinking. One clarification: in the current proposal  the subscribers would implement the ```WorkObserver / WorkSubscriber``` interface allowing us to have many different types to be Subscribers. That said, the ```WorkSubscriber``` could be declared as a func type
```
WorkSubscriber func(msg JobMsg)
```

docs/proposals/subscriber-fan-in-fan-out.md
maleck13: 👍 I should not have had the error return on the interface, I will  change this. As each WorkSubscriber is an independent function unaware of any of the other work that is happening, I think it makes sense for them to handle the error internally (similar to how the subscribers work now, marking the action as failed, cleaning up etc) as the only real options for the WorkEngine  on receiving an error would be to log it.

docs/proposals/subscriber-fan-in-fan-out.md
maleck13: I think the concept of Jobs that can be cancelled is probably a proposal of its own and outside the remit of this work. 
If we had jobs that could somehow be cancelled, then the Job itself would take the context, I don't think the subscribers would as they only react to messages passed in. In the case of cancellation, the likelihood is a message would be passed in with a state of cancelled rather than the subscriber watching the context itself.

docs/proposals/subscriber-fan-in-fan-out.md
maleck13: see above

maleck13:
Have updated with more details and removed the Observer name change.
@jmrodri Are you happy with this now? Do you need an additional drawing? 
@shawn-hurley I think we have answered your questions about context (anything outstanding now)?
@eriknelson  Made the changes you have suggested.

eriknelson:
@maleck13 I'm considering this accepted given the acks and discussion. Thanks again.

maleck13:
Great ! thanks guys

PR #728, opened 2018-02-05
Bug 1541335 - fix auth config parsing

jmrodri:
Added a test but it requires some directory to be installed on your desktop. I marked it as skip for now.


jmrodri:
TEST
-----
See plan from other PR
https://github.com/openshift/ansible-service-broker/pull/729#issuecomment-363230398

PR #729, opened 2018-02-05
Bug 1541335 - fix auth config parsing

jmrodri:


jmrodri:
TEST
------
Using `catasb` enable basic auth in your `my_vars.yml`. And set the broker label to be "bz1541335"
```
broker_enable_basic_auth: true
broker_tag: bz1541335
```
Then build an image of this branch:

```
make build-image ORG=YOURORG TAG=bz1541335
docker push docker.io/YOURORG/origin-ansible-service-broker:bz1541335
```
Then in `catasb` reset the environment
```
oc cluster down
cd local/linux/
./reset_environment.sh
```

Then use curl to get the catalog api using basic auth:

```
curl -kv "https://admin:admin@asb-1338-ansible-service-broker.172.17.0.1.nip.io/ansible-service-broker/v2/catalog"
```

PR #731, opened 2018-02-06
Make the endpoints headless for local brokers

rthallisey:
When using labels on the endpoint, we're associating the endpoint to an object that may not always be around, the locally running broker. This seems to cause the endpoint lose it's ip address. Instead, make the endpoint have no association to an app so it will always be around.

 - Keep the endpoint ip address around all the time

fixes https://github.com/openshift/ansible-service-broker/issues/660

### DIFF
--- a/scripts/kubernetes/resources.sh
+++ b/scripts/kubernetes/resources.sh
@@ -34,9 +34,6 @@ function cluster::process {
 apiVersion: v1
 kind: Endpoints
 metadata:
-  labels:
-    app: ansible-service-broker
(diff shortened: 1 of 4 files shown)


eriknelson:
This is working great for me, so I can confirm it fixes #660 and I am now much happier 😃 

@rthallisey what's up with CI, is that a valid failure? I've only just confirmed this allows me to work locally uninterupted.

eriknelson:
Ah, I see it was the vet problem from another PR. Merging.

PR #734, opened 2018-02-06
Bug 1540121 - Return 400 for bad update requests

### LINKED ISSUE #732: Correct behavior for unexpected parameters on update requests?
There are a couple branches of parameter validation where we are ignoring erroneous inputs:

* Users passing updated parameters that don't exist on a plan.
* Users passing enum values that aren't part of the acceptable set of values.

In both cases, we strip them from the request parameters. I think they should be Bad Request 400s, mainly because I think these represent a user's explicit intent, and they will be surprised when that intent has silently been ignored because it is fundamentally invalid. Looking for some thoughts on this before I submit my changes.

eriknelson:

* Filters unchanged parameters from update requests
* Returns 400 error when a user requests updating a non-updatable parameter

fixes #732 

I'm labeling this as DO-NOT-MERGE because I want to close #732 with this as well. Quoting here:

> There are a couple branches of parameter validation where we are ignoring erroneous inputs:
> * Users passing updated parameters that don't exist on a plan.
> * Users passing enum values that aren't part of the acceptable set of values.
>
> In both cases, we strip them from the request parameters. I think they should be Bad Request 400s, mainly because I think these represent a user's explicit intent, and they will be surprised when that intent has silently been ignored because it is fundamentally invalid.

~~See the TODOs, I'm arguing they should also be Bad Requests, but I'd like to get some feedback here before adding the changes.~~

I updated them to return an error after talking with @jmontleon. If anyone feels differently, feel free to chirp at me.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -364,8 +364,16 @@ func (h handler) update(w http.ResponseWriter, r *http.Request, params map[strin
 			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
 		case broker.ErrorNoUpdateRequested:
 			writeResponse(w, http.StatusOK, resp)
-		default:
+		case broker.ErrorNoUpdateRequested:
+			writeResponse(w, http.StatusOK, resp)
+		case broker.ErrorPlanNotFound,
+			broker.ErrorParameterNotUpdatable,
+			broker.ErrorParameterNotFound,
+			broker.ErrorParameterUnknownEnum,
+			broker.ErrorPlanUpdateNotPossible:
 			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+		default:
+			writeResponse(w, http.StatusInternalServerError, broker.ErrorResponse{Description: err.Error()})
 		}
 	} else if async {
 		writeDefaultResponse(w, http.StatusAccepted, resp, err)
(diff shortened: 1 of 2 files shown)


rthallisey:
CI testing against release-1.1 branch is always reporting true right now because it thinks this is only a doc change. I'll fix this soon.

eriknelson:
I ran a `make check` on it returning green + some integration testing. I need to post against master and that should run through the flow.

pkg/handler/handler.go
```diff
@@ -364,8 +364,16 @@ func (h handler) update(w http.ResponseWriter, r *http.Request, params map[strin
...
 			writeResponse(w, http.StatusOK, resp)
-		default:
+		case broker.ErrorNoUpdateRequested:
+			writeResponse(w, http.StatusOK, resp)
+		case broker.ErrorPlanNotFound:
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+		case broker.ErrorParameterNotUpdatable:
 			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
+		case broker.ErrorPlanUpdateNotPossible:
+			writeResponse(w, http.StatusBadRequest, broker.ErrorResponse{Description: err.Error()})
```
jmrodri: I think you can do the multiple case by comma separating them:
```
case broker.ErrorPlanNotFound, broker.ErrorParameterNotUpdatable, broker.ErrorPlanUpdateNotPossible:
default:
```

eriknelson: Brilliant! I rescind my earlier complaints.

PR #740, opened 2018-02-07
Bug 1540121 - Remove redundant validation copy

eriknelson:
* Noticed an oversight in the recent update validation fix.


jmrodri:
Waiting to confirm if this was part of the bug from yesterday so I can add the appropriate bug number to the merge commit.

PR #742, opened 2018-02-07
Bug 1542387 - Print out status of response from RHCC registry if not 200

dymurray:
Previous logging output was not helpful in determining the issue. Before we were getting 'invalid character' in the logs along with "Got Image Response from RHCC".

Now if RHCC fails to bootstrap it returns the status response (for current bug):
```
[2018-02-07T09:32:26.16-05:00] [ERROR] - unable to retrieve image names for registry rhcc - 500 INTERNAL SERVER ERROR
```

### DIFF
--- a/pkg/registries/adapters/rhcc_adapter.go
+++ b/pkg/registries/adapters/rhcc_adapter.go
@@ -18,6 +18,7 @@ package adapters
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


eriknelson:
Needs a PR to bring this into `release-1.1` as well.

dymurray:
https://github.com/openshift/ansible-service-broker/pull/743

PR #744, opened 2018-02-07
Bug 1543029 - last_operation returns correct body with 410 response

### LINKED ISSUE #708: unbind last_operation returns invalid response body
**What happened**:
I did an async unbind and then called the last_operation endpoint with the correct token. I received this response:

```
HTTP/1.1 410 Gone
Date: Mon, 29 Jan 2018 20:10:52 GMT
Content-Length: 5
Set-Cookie: d05351dd24520dce41f310fcfc112523=cf44571ff518c281d2b9338b30a5e779; path=/; HttpOnly; Secure
Cache-control: private

null
```

I saw this on the ``release-1.1`` branch.

The spec says for a 410 response, "The expected response body is {}."

**How to reproduce it**:
1. Configure the broker to run apbs on bind and unbind, and to auto-escalate.
2. initiate an async unbind
3. make a request to the bind/unbind last_operation endpoint with the token proved in step 2

mhrivnak:
fixes #708

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -720,7 +720,7 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
 		if err != nil {
 			switch err {
 			case broker.ErrorNotFound:
-				writeResponse(w, http.StatusGone, nil)
+				writeResponse(w, http.StatusGone, make(map[string]interface{}, 1))
 			default:
 				writeResponse(w, http.StatusInternalServerError, broker.ErrorResponse{Description: err.Error()})
 			}


pkg/handler/handler.go
```diff
@@ -720,7 +720,7 @@ func (h handler) lastoperation(w http.ResponseWriter, r *http.Request, params ma
 		if err != nil {
 			switch err {
 			case broker.ErrorNotFound:
-				writeResponse(w, http.StatusGone, nil)
+				writeResponse(w, http.StatusGone, make(map[int]bool, 1))
```
mhrivnak: Is there a better way to provide something that will serialize to ``{}`` ?

eriknelson: This is probably pedantic, but the canonical golang json representation is `map[string]interface{}`. https://blog.golang.org/json-and-go.

mhrivnak: That's exactly what I needed. Thanks!

mhrivnak:
Yep, here it is: https://bugzilla.redhat.com/show_bug.cgi?id=1543029

PR #749, opened 2018-02-08
Include necessary roles for local_openshift reg in simple-broker-template

dymurray:
Escalate asb service account privileges for simple-broker-template

 - Add more rules to asb-auth sa

Currently the up and running blog post is broken without this.


jmrodri:
I will say at what point do we stop adding the features to the simple template. The simple template works as an example for the blog post. It worked just fine. What didn't work is if a users MODIFIES the template to add the local registry. In that case they also needed to add these entries.

djzager:
I think that is a fair point @jmrodri. It probably would also have worked to say, "use this other template instead" which actually includes the local OpenShift registry + all of the permissions required.

PR #750, opened 2018-02-08
Bug 1543521 - Fix async bind with job already in progress.

### LINKED ISSUE #670: second async bind call returns incorrect response
**What happened**:
I made exactly the same async bind call twice in rapid succession. When the second call was made, the first operation was still running on the server.

Per the proposed OSB API spec, the second call should return the same token as the first in the body and a response code of 202. Instead it returned the provision credentials and a 200.

The first response:
```
HTTP/1.1 202 Accepted
Date: Fri, 19 Jan 2018 21:46:23 GMT
Content-Length: 58
Set-Cookie: d05351dd24520dce41f310fcfc112523=5fcc1b36cc26f6c610e7a29ef28f3e51; path=/; HttpOnly; Secure

{
  "operation": "ddb8668f-3b1d-49b9-a8f4-0c2b362600b3"
}
```

The second response:

```
HTTP/1.1 200 OK
Date: Fri, 19 Jan 2018 21:46:23 GMT
Content-Length: 228
Set-Cookie: d05351dd24520dce41f310fcfc112523=5fcc1b36cc26f6c610e7a29ef28f3e51; path=/; HttpOnly; Secure
Cache-control: private

{
  "credentials": {
    "DB_ADMIN_PASSWORD": "zwyWS7AHkic9BzNtkuuI",
    "DB_HOST": "postgresql",
    "DB_NAME": "admin",
    "DB_PASSWORD": "foo",
    "DB_PORT": "5432",
    "DB_TYPE": "postgres",
    "DB_USER": "admin"
  }
}
```

Same response twice in a row

**What the proposed spec says**
A 202 http response code "MUST be returned if the binding is in progress. This triggers the Platform to poll the Polling Last Operation for Service Bindings endpoint for operation status. Information regarding the Service Binding (i.e. credentials) MUST NOT be returned in this asynchronous request. Note that a re-sent PUT request MUST return a 202 Accepted, not a 200 OK, if the binding is not yet fully created."

https://github.com/mattmcneeney/servicebroker/blob/a57c1723960490bc5b31096c204cbd21d266cea3/spec.md#response-5

**Likely Cause**:
I suspect that when the second request is processed, it finds a record of the in-progress binding and misinterprets that as the binding already existing, and then defaults to returning whatever credentials it finds, which are only the provision credentials. But that's just a vague theory.

**How to reproduce it**:
1. configure the broker to run APBs on bind/unbind and auto-escalate
2. make identical PUT requests to create a binding with ``accepts_complete=true`` in rapid succession

mhrivnak:
The broker now stores a reference to a bind job with the BindInstance, so that
it can be looked up in the future if successive equivalent requests are
received.

fixes #670

The solution was discussed here: https://github.com/openshift/ansible-service-broker/pull/680

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -18,6 +18,7 @@ package apb
 
 import (
 	"encoding/json"
+	"reflect"
 
 	"github.com/openshift/ansible-service-broker/pkg/config"
 	logutil "github.com/openshift/ansible-service-broker/pkg/util/logging"
@@ -193,6 +194,15 @@ const (
 
 	// ApbContainerName - The name of the apb container
 	ApbContainerName = "apb"
+
+	// ProvisionCredentialsKey parameter name passed to APBs
+	ProvisionCredentialsKey = "_apb_provision_creds"
+	// BindCredentialsKey parameter name passed to APBs
+	BindCredentialsKey = "_apb_bind_creds"
+	// ClusterKey parameter name passed to APBs
+	ClusterKey = "cluster"
+	// NamespaceKey parameter name passed to APBs
+	NamespaceKey = "namespace"
 )
 
 // SpecLogDump - log spec for debug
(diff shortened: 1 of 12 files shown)


pkg/broker/broker.go
```diff
@@ -880,33 +881,57 @@ func (a AnsibleBroker) Bind(instance apb.ServiceInstance, bindingUUID uuid.UUID,
...
+			}
+
+			switch {
+			// unknown error
+			case err != nil && !client.IsKeyNotFound(err):
+				return nil, false, err
+			// If there is a job in "succeeded" state, or no job at all, or
+			// the referenced job no longer exists (we assume it got
+			// cleaned up eventually), assume everything is complete.
+			case createJob.State == "succeeded", existingBI.CreateJobKey == "", client.IsKeyNotFound(err):
```
shawn-hurley: Can we use the [apb job state](https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/types.go#L184) value here?

pkg/apb/types.go
```diff
@@ -274,9 +275,38 @@ func (si *ServiceInstance) RemoveBinding(bindingUUID uuid.UUID) {
...
+// should represent what a OSB API client provides in a bind request.
+func (bi *BindInstance) UserParameters() Parameters {
+	if bi.Parameters == nil {
+		return nil
+	}
+	userparams := make(Parameters)
+	for key, value := range *bi.Parameters {
+		switch key {
+		// Do not copy keys that are generally added by the broker itself.
+		case "cluster", "namespace", "_apb_provision_creds":
```
shawn-hurley: One day we should make these constants. Possibly outside the scope of this PR but I think that this is the second place we use those values.

jmrodri:
@eriknelson good catch.

eriknelson:
PR names are important for us because we've taken the approach of merging to our master with github's squash and merge feature, so the master commits take the name of the PR.

mhrivnak:
> PR names are important for us because we've taken the approach of merging to our master with github's squash and merge feature, so the master commits take the name of the PR.

Ah, gotcha. In this case there are conflicts that require resolving. I have a branch with these changes rebased onto master that I've been maintaining as revisions happen here. I'll make a PR with it once this PR gets approved.

jmrodri:
@eriknelson I've started editing the PR and commit after hitting squash and merge. Since the commits also end up in the release notes that @djzager sends out.

eriknelson:
I prematurely ack'd, I don't know how to revert. Can we make the linter happy before this comes in? There are a bunch of new public types that need docs.

mhrivnak:
It works for me?

```
$ make lint
$ echo $?
0
```

eriknelson:
@mhrivnak just don't want to merge when the CI is red.

PR #754, opened 2018-02-12
Bug 1541461 - Deal with buggy encoded scopes from service catalog.

shawn-hurley:
Will allow us to handle scope that has been incorrectly encoded

 - Deal with the other way of encoding scopes
 - Default to close if both scopes are preset

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -858,7 +858,18 @@ func (h handler) validateUser(userInfo broker.UserInfo, namespace string) (bool,
 		return false, http.StatusInternalServerError, fmt.Errorf("Unable to connect to the cluster")
 	}
 	// Retrieving the rules for the user in the namespace.
-	prs, err := openshiftClient.SubjectRulesReview(userInfo.Username, userInfo.Groups, userInfo.Extra, namespace)
+	s := userInfo.Scopes
+	if userInfo.Extra != nil {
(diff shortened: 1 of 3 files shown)


pkg/handler/handler.go
```diff
@@ -858,7 +858,18 @@ func (h handler) validateUser(userInfo broker.UserInfo, namespace string) (bool,
...
 	}
 	// Retrieving the rules for the user in the namespace.
-	prs, err := openshiftClient.SubjectRulesReview(userInfo.Username, userInfo.Groups, userInfo.Extra, namespace)
+	s := userInfo.Scopes
+	if userInfo.Extra != nil {
+		scope, ok := userInfo.Extra["scopes.authorization.openshift.io"]
+		switch {
+		case ok && userInfo.Scopes != nil:
+			log.Infof("Unable to determine correct scope to use. Found both top level scope and scope in extras.")
+			return false, http.StatusForbidden, fmt.Errorf("unable to determine correct scope to use")
```
jmrodri: @shawn-hurley how likely is this situation? Will we get into a situation where we're screwed and always getting forbidden? What would cause both information to come in?

shawn-hurley:
I believe the situation would not occur with just the service catalog. I
believe this is just to be very explicit about what would happen if it did.
There is no version of the service catalog this should happen on.

jmrodri:
@shawn-hurley okay, ACK!

PR #756, opened 2018-02-12
Bug 1541461 - Deal with buggy encoded scopes from service catalog.

shawn-hurley:
Will allow us to handle scope that has been incorrectly encoded

 - Deal with the other way of encoding scopes
 - Default to close if both scopes are preset


djzager:
Do we intend to add this to 3.7?

shawn-hurley:
I do think that this needs to be added to 3.7

eriknelson:
CI is just busted here, correct? Or is there a legitimate reason why this should not get merged?

PR #757, opened 2018-02-12
[WIP] Debugging openshift bind CI failure

rthallisey:
openshift uses: ```oc env dc ...``` to load bind creds and kubernetes uses: ```kubectl set env deploy ...```.  It could be something in the client causing openshift's ci job to periodically fail.


rthallisey:
bumping sleep to 10 on verify script

PR #762, opened 2018-02-14
Remove upgrade from Ubuntu 14.0 to 16.04

rthallisey:
This should speed up the openshift job!

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -13,8 +13,6 @@ if [[ "$action" == "before_install" ]]; then
   echo "================================="
   echo "        Before Install           "
   echo "================================="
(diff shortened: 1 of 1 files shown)


jmrodri:
I thought we put this upgrade in to fix a problem. Is this problem no longer an issue?

rthallisey:
@jmrodri it was fixed in new versions of kubernetes.

rthallisey:
CI on the master branch has been having issues lately.  The issue isn't related to this PR because the upgrade fixed oc cluster up failing on the service-catalog stating

jmrodri:
If removing the upgrade will move us towards fixing CI, then ACK. I just didn't want to undo something doen to solve a previous issue without knowing if said issue has been addressed. Sounds like it has.

PR #765, opened 2018-02-15
Fix last op endpoint unbind response

### LINKED ISSUE #709: unbind last_operation response always responds with 410 Gone
**What happened**:

When an async unbind operation is in-progress, the last_operation endpoint incorrectly responds with a ``410 Gone`` response code.

This is on the ``release-1.1`` branch.

The spec says a 410 response to a last_operation request is "Appropriate only for asynchronous delete operations. The Platform MUST consider this response a success and remove the resource from its database."

In our case, a 410 response does not indicate success because it gets returned even while the job is in-progress. Thus our response is misleading.

**How to reproduce it**:
Use the last_operation endpoint to query the state of an in-progress unbind operation.

**Cause**
This is likely caused by the fact that when the DELETE is requested on the binding, the ``BindInstance`` is deleted from etcd before even a 202 response is returned. It may be better to do that deleting at the end of the unbind job.

PhilipGough:
Ready for review

 - If job already in progress return correct status code to handler
 - Wrap clean up of unbind process in function, which can be called from async job or broker

fixes #709

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -64,7 +64,9 @@ var (
 	ErrorPlanUpdateNotPossible = errors.New("plan update not possible")
 	// ErrorNoUpdateRequested - Error for when no valid updates are requested
 	ErrorNoUpdateRequested = errors.New("no valid updates requested")
-	log                    = logutil.NewLog()
+	// ErrorUnbindingInProgress - Error when unbind is called that has an unbinding job in progress
+	ErrorUnbindingInProgress = errors.New("unbinding in progress")
+	log                      = logutil.NewLog()
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/broker/broker.go
```diff
@@ -1024,29 +1035,16 @@ func (a AnsibleBroker) Unbind(
...
-			return nil, false, err
-		}
-	}
-
-	err = a.dao.DeleteBindInstance(bindInstance.ID.String())
-	if err != nil {
-		return nil, false, err
+		err = cleanupUnbind(&bindInstance, &serviceInstance, a.dao, true)
+	} else {
+		err = cleanupUnbind(&bindInstance, &serviceInstance, a.dao, false)
```
shawn-hurley: I think that we should remove the if else and just call `cleanupUnbind` and pass in the bind creds to `cleanupUnbind` and do the nil check in that method. I think that might be more clear what that is doing.

PhilipGough: Yeah sounds good to me.

eriknelson:
Is this considered complete and ready for review / merge? The PR message still states it's not ready, I might have missed something though.

pkg/broker/broker.go
```diff
@@ -953,6 +953,16 @@ func (a AnsibleBroker) Unbind(
...
 	}
 
+	jobInProgress, jobToken, err := a.isJobInProgress(&instance, apb.JobMethodUnbind)
+	if err != nil {
+		log.Errorf("An error occurred while trying to determine if a unbind job is already in progress for instance: %s", instance.ID)
+		return nil, false, err
+	}
+	if jobInProgress {
+		log.Infof("Unbind requested for instance %s, but job is already in progress", instance.ID)
+		return &UnbindResponse{Operation: jobToken}, false, ErrorDeprovisionInProgress
```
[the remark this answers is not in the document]
PhilipGough: Good point. Ive added logic to the handler to check the error but I think returing 202 regardless of async or not is appropriate since the job in running and it will force the client to poll last op endpoint for the result. This meets the spec `202 MUST be returned if the unbinding is in progress.` which in this case it is. 

This logic is also used for the deprovision handler

PhilipGough:
@eriknelson ready for review yes, dont think its ready for merge yet

PR #766, opened 2018-02-15
Remove plan dao methods

shawn-hurley:
Currently, we save a mapping between names and id and have the data in
the spec already.
This was a convenience method to quickly get that data, but when moving
to CRDs I don't think we should create a resource for a single value
mapping

 - Remove all interface and implementation methods for plans
 - Add helper method on spec to get plans from ID
 - Remove pointer to spec values memory when getting plans. Makes it more memory safe

### DIFF
--- a/pkg/apb/types.go
+++ b/pkg/apb/types.go
@@ -103,15 +103,26 @@ type Spec struct {
 	Plans       []Plan                 `json:"plans"`
 }
 
-// GetPlan - retrieves a reference to a plan from a spec by name. Will return
-// nil if the requested plan does not exist.
-func (s *Spec) GetPlan(name string) *Plan {
-	for i, plan := range s.Plans {
+// GetPlan - retrieves a  plan from a spec by name. Will return
+// empty plan and false if the requested plan does not exist.
+func (s *Spec) GetPlan(name string) (Plan, bool) {
+	for _, plan := range s.Plans {
 		if plan.Name == name {
-			return &s.Plans[i]
+			return plan, true
 		}
 	}
-	return nil
+	return Plan{}, false
+}
+
+// GetPlanFromID - retrieves a refernece to a plan from a spec by name. Will return
+// empty plan and false if the requested plan does not exist.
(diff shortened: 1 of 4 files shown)


jmontleon:
ACK

pkg/apb/types.go
```diff
@@ -103,15 +103,26 @@ type Spec struct {
...
 		if plan.Name == name {
-			return &s.Plans[i]
+			return plan, true
 		}
 	}
-	return nil
+	return Plan{}, false
+}
+
+// GetPlanFromID - retrieves a refernece to a plan from a spec by name. Will return
```
jmrodri: typo: refernece -> reference

pkg/apb/types.go
```diff
@@ -103,15 +103,26 @@ type Spec struct {
...
+// empty plan and false if the requested plan does not exist.
+func (s *Spec) GetPlanFromID(id string) (Plan, bool) {
+	for _, plan := range s.Plans {
+		if plan.ID == id {
+			return plan, true
+		}
+	}
+	return Plan{}, false
```
jmrodri: Is this a common pattern? Seems kind of weird to me.

jmrodri: Would it be bad to still return a pointer but then return a copy of the plan?

```golang
func (s *Spec) GetPlanFromID(id string) *Plan {
    for _, plan := range s.Plans {
        if plan.ID == id {
            var copy = &Plan{}
            *copy = *s.plan
            return copy
        }   
    }   
    return nil
}  
```

shawn-hurley: It wouldn't be bad as far as I know and if you have strong feelings then I'll change it, but the above behavior is the same behavior you get from the map. `val, ok := map["unknownkey"]` returns the empty value for val and false. 

I would argue that the pointer here does not make sense. We don't want to manipulate the plan in the Spec and therefore a pointer return is misleading IMO.

eriknelson: Returning-by-value makes sense to me here if we're going to be returning a copy anyway.

PR #767, opened 2018-02-15
Lock down everything in branch to this release

djzager:
Make it so things in the `release-1.0` branch use `release-1.0` images + packages + etc.

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -418,7 +418,7 @@ parameters:
 - description: Container Image to use for Ansible Service Broker in format of imagename:tag
   displayname: Ansible Service Broker Image
   name: BROKER_IMAGE
-  value: ansibleplaybookbundle/origin-ansible-service-broker:latest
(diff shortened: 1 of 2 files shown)


djzager:
I'm still trying to figure out how to make the `Dockerfile-latest` in this branch build using the 1.0 ansible-service-broker from copr. Maybe @jmontleon knows how to do that sorcery.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -418,7 +418,7 @@ parameters:
 - description: Container Image to use for Ansible Service Broker in format of imagename:tag
   displayname: Ansible Service Broker Image
   name: BROKER_IMAGE
-  value: ansibleplaybookbundle/origin-ansible-service-broker:latest
+  value: ansibleplaybookbundle/origin-ansible-service-broker:release-1.0
```
jmrodri: Do we just do this as part of our branching exercise?

djzager: I believe we should. I think in the future, when we talk about branching, we should make sure the container images (and tags) exist for the branch and are up to date then create the new branch with these changes.

PR #773, opened 2018-02-18
apb pkg public interface overhaul

### LINKED ISSUE #771: Public apb action methods should be thoughtfully considered, and reworked.
This issue stems from much of the discussion on #619 

There are a few items we felt needed to be addressed in a follow up PR, and tracked on this issue:

1) The apb pkg should expose the suite of actions (provision, depro etc.) with async and and sync variants. The async variants are expected to be **non-blocking**, and will return a channel for event-driven updates. It is the caller's responsibility to monitor this channel for status. The async variant is a recommended best practice. Alternatively, sync variants should also be available. These will block, and callers should assume the APBs have run to completion and the method should return the result of the execution.

2) The separation of `Sync/Async` jobs should be reverted. The `WorkEngine` is for asynchronous work, and there is little value for a sync job with status updates.

eriknelson:
CC: @maleck13, this is the intended follow-up for #619. We identified the way we'd like the public interfaces of the apb pkg to move in light of trying to think more consciously about what the experience should be like consuming `libapb`. Namely:

* There should be a suite of methods for callers to run APBs both asynchronously, and synchronously.
* The async calls are non-blocking, and should push updates back to the caller over a `StatusMessage` channel.
* The sync calls should block, and return with a result.

Changes proposed in this pull request:
 - Implement the above items. Introduces an `Executor` type that manages the APB lifecycle, and exposes the various methods for execution.
 - Update the broker calling code to use the new apb pkg correctly.
 - Job code has become virtually identical, so this removes the individual jobs favoring a more generic `apbJob` type, with concrete wrapper methods that call into the generic job.

Couple blockers before a merge:
* It's obviously a pretty big change to how we're doing things, so I would like to get a lot of review attention and testing.
* ~~I deleted a lot of non-relevant tests and haven't added any to test the new code paths, so I think we need a strategy to make sure the new code is properly tested and to get that implemented. I mostly wanted to just prove out in the code that this was possible~~ The new jobs routine is fully tested.

**EDIT 02/20:** I added white box testing that exercises the new job branches. I'm not too concerned with this now given a green `make ci` + `make check`.

~~**TODO**: I still need to implement the `Sync` variants of the apb actions. I'm seeing them simply call the async versions and block while watching the channel. The caller will know no better, and we're using the same codepaths.~~ This is required, but the domain of another PR.

fixes #771


pkg/apb/executor.go
```diff
@@ -32,11 +32,74 @@ import (
...
+	return e.lastStatus
+}
+
+// ExtractedCredentials - Credentials extracted from the APB while running,
+// if they were discovered.
+func (e *Executor) GetExtractedCredentials() *ExtractedCredentials {
+	return e.extractedCredentials
+}
+
+func (e *Executor) start() {
```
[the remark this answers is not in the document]
eriknelson: They're really lifecycle hooks that need to be called from the action methods to alert the `Executor` to the current state of what's going on (and it abstracts the fact that messages are being sent). I originally named `updateDescriptor` to something like `sendDescriptorMessage`, but I felt like the latter was a less abstract name. I want to call high level hooks to alert the executor to triggers in the cycle without saying, "hey send this message", because it's better if the actions don't even care about that. They just say "hey, this thing happened, do what you would like with this info".

pkg/apb/bind.go
```diff
@@ -37,41 +37,55 @@ func Bind(instance *ServiceInstance,
...
-	if err != nil {
-		log.Error("Something went wrong getting kubernetes client")
-		return executionContext.PodName, nil, err
-	}
+	go func() {
+		e.start()
+		metrics.ActionStarted("bind")
+		executionContext, err := e.executeApb(
+			"bind", instance.Spec, instance.Context, parameters)
+		defer runtime.Provider.DestroySandbox(
```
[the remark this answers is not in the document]
eriknelson: I'm not clear on the best practice with defers. If I'm doing something and I know I'll need to call the relevant "finish" function at the end of it, it makes sense to me to defer the call immediately after, assuming the mechanics of defer apply. A good example of this is the metrics hooks in the `apbJob`, I just partner the Start with a defered finish right way. I'll take a closer look at this, thanks for raising it.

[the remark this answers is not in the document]
eriknelson: After looking at this, I think it doesn't actually matter in practice? Regardless of where the defer call is made, it will execute at the conclusion of the function. Optically, it might make sense to be made elsewhere, though.

pkg/apb/executor.go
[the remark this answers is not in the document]
eriknelson: I'll also say that I tried to avoid imposing additional requirements on callers since I expect these to turn into the primary methods that people will use with `libapb`. The worst of it is to ask folks to instantiate an `Executor` with `NewExecutor`; I've personally seen us on this path for a while, so I'm happy to finally get this down on paper, so to speak.

pkg/apb/executor.go
```diff
@@ -32,11 +32,74 @@ import (
 	"k8s.io/api/core/v1"
 )
 
-// ExecuteApb - Runs an APB Action with a provided set of inputs
-func ExecuteApb(action string,
-	spec *Spec,
-	context *Context,
-	p *Parameters) (ExecutionContext, error) {
+// Executor - Main struct used for running APBs. Manages their lifecycle.
+type Executor struct {
```
shawn-hurley: I would like to make it more clear that a single Executor for a single APB run. At least that is how I am reading this code am I wrong?

eriknelson: Yeah that's exactly correct. What do you think the best way is to convey that?

shawn-hurley: I would say documentation for the end user. Code examples will help too. I think that nothing from the code will make it more obvious. At least I can't think of anything at the moment.

pkg/broker/jobs.go
```diff
@@ -0,0 +1,242 @@
...
+
+type metricsHookFn func()
+type runFn func(*apb.Executor) <-chan apb.StatusMessage
+
+type apbJob struct {
+	serviceInstanceID      string
+	specID                 string
+	bindingID              *string
+	method                 apb.JobMethod
+	metricsJobStartHook    metricsHookFn
```
maleck13: So potentially as part of the work I have outlined under the proposal for the changes to the work engine these would become StateObservers / Subscribers and not needed here

pkg/apb/provision_or_update.go
```diff
@@ -35,10 +35,8 @@ const (
 )
 
 // returns PodName, ExtractedCredentials, error
-func provisionOrUpdate(method executionMethod,
-	instance *ServiceInstance,
-	stateUpdates chan<- JobState) (string, *ExtractedCredentials, error) {
-
+func (e *Executor) provisionOrUpdate(method executionMethod, instance *ServiceInstance) {
+	e.start()
```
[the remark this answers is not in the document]
maleck13: Also see comment below around the start method

pkg/apb/executor.go
```diff
@@ -32,11 +32,74 @@ import (
...
+	close(e.statusChan)
+}
+
+func (e *Executor) finishWithError(err error) {
+	status := e.lastStatus
+	status.Error = err
+	status.State = StateFailed
+	e.lastStatus = status
+	e.statusChan <- status
+	close(e.statusChan)
```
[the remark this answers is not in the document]
eriknelson: Very good point. Reading a bit, it looks like there isn't a safe way to check if a channel has been closed without reading from it, which feels ugly and indirect to me. Alternatively, we can lock the sensitive code and nil it out, then check if it's still present before closing.

pkg/apb/executor.go
```diff
@@ -32,11 +32,74 @@ import (
...
+func NewExecutor() *Executor {
+	exec := &Executor{
+		statusChan: make(chan StatusMessage),
+		lastStatus: StatusMessage{State: StateNotYetStarted},
+	}
+	return exec
+}
+
+// GetPodName - Returns the name of the pod running the APB
+func (e *Executor) GetPodName() string {
```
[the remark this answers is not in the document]
eriknelson: Updated!

pkg/apb/executor.go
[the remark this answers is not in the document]
eriknelson: Added some locks.

pkg/apb/executor.go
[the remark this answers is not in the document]
eriknelson: Do we think there is a better name for `Executor`? I'm still hesitant to name these something like "sendX" or other related messaging concept. They're internal lifecycle hooks intended to be called by the internal APB code, and do more than send a message over the buffer (not much today, but tomorrow they might).

> also would give an opportunity for this to be an interface that could be overridden for tests or other uses if we exposed it via a constructor.

Definitely interested in this since decent test coverage is the biggest thing missing right now. I don't think callers should have to worry about injecting that via ctor though. A sensible default that can get overriden by test code I think makes sense, given this is an internal detail.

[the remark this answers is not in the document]
maleck13: > Definitely interested in this since decent test coverage is the biggest thing missing right now. I don't think callers should have to worry about injecting that via ctor though. A sensible default that can get overriden by test code I think makes sense, given this is an internal detail.

Yep agree no need to expose as long as it can be overridden in test code

pkg/apb/executor.go
```diff
@@ -32,11 +33,117 @@ import (
...
+type ExecutorAsync interface {
+	Provision(*ServiceInstance) <-chan StatusMessage
+	Deprovision(instance *ServiceInstance) <-chan StatusMessage
+	Bind(instance *ServiceInstance, parameters *Parameters) <-chan StatusMessage
+	Unbind(instance *ServiceInstance, parameters *Parameters) <-chan StatusMessage
+	Update(instance *ServiceInstance) <-chan StatusMessage
+}
+
+// Executor - Composite executor interface.
+type Executor interface {
```
shawn-hurley: 👍

pkg/apb/types.go
```diff
@@ -380,17 +389,17 @@ type ExecutionContext struct {
 	ProxyConfig    *ProxyConfig
 }
 
-// Provisioner defines a function that knows how to provision an apb
-type Provisioner func(si *ServiceInstance, statusUpdate chan<- JobState) (string, *ExtractedCredentials, error)
+// Provisioner - a function that knows how to provision an apb
+type Provisioner func(si *ServiceInstance) (<-chan StatusMessage, error)
```
shawn-hurley: I think we can delete these interfaces with this PR?

shawn-hurley:
I would also like us to make sure that we are not Exposing methods and types that we may not want to. I am thinking specifically of the ExtractCredentials functions.

pkg/apb/executor.go
```diff
@@ -32,11 +33,117 @@ import (
...
+}
+
+// ExecutorAsync - Main interface used for running APBs asynchronously.
+type ExecutorAsync interface {
+	Provision(*ServiceInstance) <-chan StatusMessage
+	Deprovision(instance *ServiceInstance) <-chan StatusMessage
+	Bind(instance *ServiceInstance, parameters *Parameters) <-chan StatusMessage
+	Unbind(instance *ServiceInstance, parameters *Parameters) <-chan StatusMessage
+	Update(instance *ServiceInstance) <-chan StatusMessage
+}
```
jmrodri: Will there be an `ExecutorSync`?

pkg/apb/types.go
eriknelson: I've actually been going back and forth on this. I deleted them and then force pushed them back in. I'm trying to test asb_jobs, and the `runFn` should render these moot, so I'll probably pull them again.

eriknelson:
@shawn-hurley Are you taking issue with the `Executor.ExtractedCredentials`? I went back and forth on this one... originally I did not have an accessor because it's a pointer type and could easily be mutated by a caller, but I added the accessor method for consistency because it felt silly to have the rest of the fields retrieved by method and not this one.

I think you should be able to get `ExtractedCredentials` from the `Executor` somehow.

shawn-hurley:
No we are exposing how we pull creds from the pod in the ext_creds.go I
think.

If it’s out of scope just tell me to go away until your done :)

eriknelson:
@shawn-hurley can you elaborate? This is all good discussion; we're going to have to figure out how to split up the changes so this refactoring doesn't apply to the whole world. I would really like to keep PRs small and focused, and I know I've expressed that to you. I'm completely violating that here.

pkg/apb/bind.go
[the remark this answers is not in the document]
jmrodri: It doesn't matter, what I was saying was tying statements together.

```golang
    go func() {
        e.start()
        metrics.ActionStarted("bind")

        defer runtime.Provider.DestroySandbox(
            executionContext.PodName,
            executionContext.Namespace,
            executionContext.Targets,
            clusterConfig.Namespace,
            clusterConfig.KeepNamespace,
            clusterConfig.KeepNamespaceOnError,
        )

        executionContext, err := e.executeApb("bind", instance.Spec, instance.Context, parameters)
        if err != nil {
            log.Errorf("Problem executing apb [%s] bind", executionContext.PodName)
            e.finishWithError(err)
            return
        }
```

So the `err != nil` follows the executeApb so it is obvious that it is related and highly unlikely that someone will come and add another line of code that set err. As it stands now the defer statement separates the executeApb and the err check. If someone comes in later they could accidentally set err and we'll be eating an err.

pkg/apb/bind.go
```diff
@@ -37,41 +37,55 @@ func Bind(instance *ServiceInstance,
...
+		}
 
-	if instance.Spec.Runtime >= 2 {
-		err := watchPod(executionContext.PodName, executionContext.Namespace, k8scli.Client.CoreV1().Pods(executionContext.Namespace), stateUpdates)
+		creds, err := ExtractCredentials(
+			executionContext.PodName,
+			executionContext.Namespace,
+			instance.Spec.Runtime,
+		)
+		e.extractedCredentials = creds
```
jmrodri: What if there is an err with `ExtractCredentials`? Do we want to set `e.extractedCredentials` before checking the err value? If this is a conscience decision then okay, just wanted to make sure we're not setting something invalid.

pkg/apb/deprovision.go
```diff
@@ -34,42 +34,49 @@ func Deprovision(instance *ServiceInstance, stateUpdates chan<- JobState) (strin
...
+			executionContext.Namespace,
+			executionContext.Targets,
+			clusterConfig.Namespace,
+			clusterConfig.KeepNamespace,
+			clusterConfig.KeepNamespaceOnError,
+		)
+		if err != nil {
+			log.Errorf("Problem executing apb [%s] deprovision", executionContext.PodName)
+			e.finishWithError(err)
+			return
```
jmrodri: same comment from bind.go regarding defer separating the err check.

pkg/apb/provision.go
```diff
@@ -30,9 +27,7 @@ func Provision(
...
 	log.Notice("============================================================")
 
-	// Nearly all of the logic for provisioning or updating is shared between
-	// provision and update, save for passing through the method type. Provision
-	// provides a nice public interface, but the bulk of the work is passed to
-	// provisionOrUpdate as an implementation detail.
-	return provisionOrUpdate(executionMethodProvision, instance, stateUpdates)
+	go e.provisionOrUpdate(executionMethodProvision, instance)
+
+	return e.statusChan
```
jmrodri: :+1: provision.go looks okay

pkg/apb/update.go
```diff
@@ -27,9 +27,7 @@ func Update(instance *ServiceInstance, statusUpdate chan<- JobState) (string, *E
 	log.Noticef("Spec.Description: %s", instance.Spec.Description)
 	log.Notice("============================================================")
 
-	// Nearly all of the logic for provisioning or updating is shared between
-	// provision and update, save for passing through the method type. Update
-	// provides a nice public interface, but the bulk of the work is passed to
-	// provisionOrUpdate as an implementation detail.
-	return provisionOrUpdate(executionMethodUpdate, instance, statusUpdate)
+	go e.provisionOrUpdate(executionMethodUpdate, instance)
+
```
jmrodri: update.go is okay

pkg/apb/provision_or_update.go
```diff
@@ -76,31 +78,39 @@ func provisionOrUpdate(method executionMethod,
...
+		return
 	}
 
 	creds, err := ExtractCredentials(
 		executionContext.PodName,
 		executionContext.Namespace,
 		instance.Spec.Runtime,
 	)
+	e.extractedCredentials = creds
+
```
jmrodri: same comment as bind, are we sure we want to set this in case of an err. If so, okay.

pkg/apb/types.go
```diff
@@ -139,6 +139,13 @@ type ExtractedCredentials struct {
 // State - Job State
 type State string
 
+// StatusMessage - Describes the latest known status of a running APB
+type StatusMessage struct {
+	State       State
+	Description string
+	Error       error
+}
+
```
jmrodri: +1 to `StatusMessage`

pkg/apb/types.go
jmrodri: I don't see Provisioner used anywhere.

pkg/apb/unbind.go
```diff
@@ -34,31 +37,42 @@ func Unbind(instance *ServiceInstance, parameters *Parameters, stateUpdates chan
...
+		if err != nil {
+			log.Errorf("Unbind action failed - %v", err)
+			e.finishWithError(err)
+			return
+		}
 
-	return nil
+		e.finishWithSuccess()
+	}()
+
```
jmrodri: unbind.go looks okay

pkg/apb/watch_pod.go
```diff
@@ -34,7 +34,12 @@ var (
 	ErrorActionNotFound = fmt.Errorf("action not found")
 )
 
-func watchPod(podName string, namespace string, podClient v1.PodInterface, statusUpdates chan<- JobState) error {
+type updateDescriptionFn func(string)
+
+func watchPod(
+	podName string, namespace string, podClient v1.PodInterface,
+	updateDescription updateDescriptionFn,
```
jmrodri: watch_pod changes :+1:

pkg/broker/broker.go
```diff
@@ -1241,7 +1242,7 @@ func (a AnsibleBroker) Update(instanceUUID uuid.UUID, req *UpdateRequest, async
 	log.Debugf("toPlanName: [%s]", toPlan.Name)
 	log.Debugf("PreviousValues: [ %+v ]", req.PreviousValues)
 	log.Debugf("ServiceInstance Parameters: [%v]", *si.Parameters)
-	ujob := NewUpdateJob(si, apb.Update)
+	ujob := &UpdateJob{si}
```
jmrodri: I'm okay with using the job structs directly :+1:

pkg/broker/jobs.go
```diff
@@ -0,0 +1,248 @@
...
+		specID:                 j.serviceInstance.Spec.ID,
+		method:                 apb.JobMethodUpdate,
+		metricsJobStartHook:    metrics.UpdateJobStarted,
+		metricsJobFinishedHook: metrics.UpdateJobFinished,
+		skipExecution:          false,
+		run: func(e apb.Executor) <-chan apb.StatusMessage {
+			return e.Update(j.serviceInstance)
+		},
+	}
+	job.Run(token, msgBuffer)
```
jmrodri: I looked over the entire jobs.go. :exclamation:  looks great. I like it.

pkg/metrics/metrics.go
```diff
@@ -161,14 +161,14 @@ func DeprovisionJobStarted() {
 	deprovisionJob.Inc()
 }
 
-// BindingJobStarted - Add a provision job to the counter.
-func BindingJobStarted() {
+// BindJobStarted - Add a provision job to the counter.
+func BindJobStarted() {
 	defer recoverMetricPanic()
```
jmrodri: okay, this is fine.

pkg/apb/executor.go
```diff
@@ -32,11 +33,117 @@ import (
...
+func (e *executor) ExtractedCredentials() *ExtractedCredentials {
+	return e.extractedCredentials
+}
+
+func (e *executor) start() {
+	log.Debug("executor::start")
+	status := e.lastStatus
+	status.State = StateInProgress
+	e.lastStatus = status
+	e.statusChan <- status
```
jmrodri: I'm confused by this, we copy the lastStatus into status, set its state, then set it back on lastStatus. Why isn't this just:

```golang
func (e *executor) start() {
    log.Debug("executor::start")
    e.lastStatus.State = StateInProgress
    e.statusChan <- e.lastStatus
}
```

shawn-hurley:
@eriknelson re: extracted credentials

This is what I was thinking about: https://github.com/openshift/ansible-service-broker/blob/master/pkg/apb/ext_creds.go#L48

I agree that this should be a follow-on, your right that this is something that can be done separately. I mentioned it here because #771  I thought was referring to the `apb` package and not just the actions. Sorry if this is off topic.

pkg/apb/bind.go
eriknelson: It was actually originally written this way, I thought about it and the expectation is that creds will be nil if an error is returned from `ExtractCredentials`. Your comment is probably safer, however. Will update.

pkg/apb/bind.go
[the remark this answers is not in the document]
eriknelson: The issue is that `executionContext` isn't in scope until the `executeApb` call finished. I'll update to keep the err check tied closely.

eriknelson:
@shawn-hurley I actually had the same thought and had marked it private, but the broker actually calls it in broker.go as part of recovery.

shawn-hurley:
I didn't realize that we may want to move recover out of the broker then. I think that may require a new issue to track and discuss, thoughts @jmrodri @rthallisey @eriknelson?

eriknelson:
@shawn-hurley IMO recover is firmly a concept in the broker domain. I'm not sure where else it would belong aside from there.

pkg/apb/executor.go
eriknelson: I think I was on autopilot here and it felt like a redux action, which is of course totally irrelevant here. I'll update :smile:

eriknelson:
@shawn-hurley @maleck13 @rthallisey Think this should be considered complete at this point. I'm planning on submitting another PR with the sync work.

eriknelson:
This utterly ballooned to almost unreviewable after I added the vendor stuff. If you look at the individual commits prior to that, I added `jobs.go` tests and in an effort to test it completely, I needed an `Executor` mock. I used [mockery](https://github.com/vektra/mockery) to generate `mock_Executor.go`, which was enormously helpful. It requires [testify](https://github.com/stretchr/testify), which I had to add to the vendor deps.

Gotta say, it was crazy easy to integrate and incredibly useful. I would love for us to continue with that pattern.

pkg/apb/mock_Executor.go
```diff
@@ -0,0 +1,134 @@
+// Code generated by mockery v1.0.0
```
shawn-hurley: I am wondering if we should have a mock package for this? 

Just makes it more clear in your import that you're importing a mock?

eriknelson: I started thinking about this as well, but it implies each pkg would have a corresponding child mock package? Does that make sense?

maleck13: what if you were to create a top level mock package that contained shared mocks instead of a mock package in each individual package?

eriknelson: I don't think we want to be mixing mocks for a bunch of different packages into one package.

eriknelson:
Thank you for the reviews everyone! Merging since there is a bunch of work dependent on this, and it's green.

PR #774, opened 2018-02-19
Bug 1539542 - return bindings_retrievable

jmrodri:
the async bind spec change requires bindings_retrievable when we support
async binding for a service. We are taking the approach that if an apb
is bindable, it will support async bind.

fixes BZ 1539542 https://bugzilla.redhat.com/show_bug.cgi?id=1539542

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -470,6 +470,14 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 		if err != nil {
 			log.Errorf("not adding spec %v to list of services due to error transforming to service - %v", spec.FQName, err)
 		} else {
+			// Bug 1539542 - in order for async bind to work,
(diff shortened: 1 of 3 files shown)


pkg/broker/broker.go
```diff
@@ -470,6 +470,14 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 		if err != nil {
 			log.Errorf("not adding spec %v to list of services due to error transforming to service - %v", spec.FQName, err)
 		} else {
+			// Bug 1539542 - in order for async bind to work,
```
shawn-hurley: I don't think that we need this in an else block. Just a style thing, surprised if this passes lint.

jmrodri: it was already there so I didn't want to muck with the existing code. I'll fix it though.

PR #775, opened 2018-02-19
Save extracted credentials into a secret

shawn-hurley:
This PR will implement #768
- [x] ~~Add kubernetes client methods to interact with extracted credentials in the [namespace](https://github.com/openshift/ansible-service-broker/blob/master/docs/config.md#openshift-configuration).~~
- [x] ~~Add runtime methods for interacting with extracted credentials. These methods should be overridable.~~
- [x] ~~Remove all dao implementation and interface methods regarding extracted credentials.~~
- [x] ~~Remove all instances of interacting with dao extracted credentials in the `broker` package. Add back call to APB package to get extracted credentials when needed.~~
- [x] ~~Add exposed method on APB  that will retrieve the extracted credentials.~~
- [x] ~~Update APB package to create/save/delete extracted credentials for the correct actions. this should call the correct `runtime` package methods.~~

### DIFF
--- a/pkg/apb/deprovision.go
+++ b/pkg/apb/deprovision.go
@@ -74,6 +74,12 @@ func (e *executor) Deprovision(instance *ServiceInstance) <-chan StatusMessage {
 			e.actionFinishedWithError(err)
 			return
 		}
+		err = runtime.Provider.DeleteExtractedCredential(instance.ID.String(), clusterConfig.Namespace)
+		if err != nil {
+			log.Errorf("unable to delete the extracted credentials - %v", err)
+			e.actionFinishedWithError(err)
+			return
+		}
 
 		e.actionFinishedWithSuccess()
 	}()
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -42,6 +42,11 @@ const (
 	GatherCredentialsCommand = "broker-bind-creds"
 )
 
+var (
+	// ErrExtractedCredentialsNotFound - Extracted Credentials are not found.
+	ErrExtractedCredentialsNotFound = fmt.Errorf("credentials not found")
+)
+
 type extractCreds func(string, string) (*ExtractedCredentials, error)
 
 // ExtractCredentials - Extract credentials from pod in a certain namespace.
@@ -191,7 +196,14 @@ func decodeOutput(output []byte) ([]byte, error) {
 func GetExtractedCredentials(id string) (*ExtractedCredentials, error) {
 	creds, err := runtime.Provider.GetExtractedCredential(id, clusterConfig.Namespace)
 	if err != nil {
-		log.Errorf("unable to get the extracted credential secret - %v", err)
+		switch {
+		case err == runtime.ErrCredentialsNotFound:
+			log.Debugf("extracted credential secret not found - %v", id)
+			return nil, ErrExtractedCredentialsNotFound
+		default:
+			log.Errorf("unable to get the extracted credential secret - %v", err)
+			return nil, err
+		}
 	}
 	return &ExtractedCredentials{Credentials: creds}, nil
 }
--- a/pkg/apb/ext_creds.go
+++ b/pkg/apb/ext_creds.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/openshift/ansible-service-broker/pkg/clients"
+	"github.com/openshift/ansible-service-broker/pkg/runtime"
 	"github.com/openshift/ansible-service-broker/pkg/version"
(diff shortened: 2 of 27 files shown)


pkg/runtime/runtime.go
```diff
@@ -94,7 +96,14 @@ func NewRuntime() {
 		panic(err.Error())
 	}
 
-	Provider = &provider{coe: cluster}
+	var c ExtractedCredential
+	if extCreds == nil {
```
rthallisey: Can you explain why we should have this if/else?

pkg/apb/ext_creds.go
```diff
@@ -185,3 +191,31 @@ func decodeOutput(output []byte) ([]byte, error) {
 
 	return decodedjson, nil
 }
+
+// GetExtractedCredentials - Will get the extracted credentials for a caller of the APB package.
+func GetExtractedCredentials(id string) (*ExtractedCredentials, error) {
+	creds, err := runtime.Provider.GetExtractedCredential(id, clusterConfig.Namespace)
+	if err != nil {
+		switch {
+		case err == runtime.ErrCredentialsNotFound:
```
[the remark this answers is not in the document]
shawn-hurley: So here is my thinking on why each package has its own NotFound Error.

I don't want the error to bleed across package boundaries. Right now runtime package is only called by apb. I think we should keep it that way. The same is not currently true of the client, but eventually, we need to encapsulate the client calls in the runtime package. 

I think this allows a user who wants to override the runtime package calls for extracted credentials can send back the runtime error and now our APB package can use that. Does that make sense what I was trying to do?

Would love to hear thoughts on this approach.

pkg/runtime/runtime.go
shawn-hurley: if nil we will use the default extracted credentials. This should be in the godoc for the function and will add now.

eriknelson:
Caught up with @shawn-hurley to coordinate this PR and #773 since they're in direct conflict. Plan is to bring in #773 first and then rework this PR on top of it. Should be relatively straightforward, @shawn-hurley happy to help however I can.

pkg/runtime/runtime.go
rthallisey: I want to clarify my last comment.  The runtime interface is structured so that if you call ```NewRuntime``` the Runtime pkg will identify what provider you are using.   If we add ```ExtractedCredential``` type to the ```NewRuntime``` call, ```pkg/app/app.go``` needs to know what runtime provider is being used in order to provide a working ```ExtractedCredential``` type.  The reason for that is because ```ExtractedCredential``` is stored in a cluster specific resource, a secret.  My point is I don't think we want to pass ```ExtractedCredential``` as a param to ```NewRuntime```, rather we should add it behind the runtime interface it be initialized after the ```NewRuntime``` call.

pkg/apb/deprovision.go
```diff
@@ -74,6 +74,12 @@ func (e *executor) Deprovision(instance *ServiceInstance) <-chan StatusMessage {
 			e.actionFinishedWithError(err)
 			return
 		}
+		err = runtime.Provider.DeleteExtractedCredential(instance.ID.String(), clusterConfig.Namespace)
+		if err != nil {
+			log.Errorf("unable to delete the extracted credentials - %v", err)
+			e.actionFinishedWithError(err)
+			return
```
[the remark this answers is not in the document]
eriknelson: Yeah it will, think this is where it might make sense for us to have our own error type, and skip the action finished call. Log with INFO level.

pkg/apb/ext_creds.go
```diff
@@ -41,6 +42,11 @@ const (
 	GatherCredentialsCommand = "broker-bind-creds"
 )
 
+var (
+	// ErrExtractedCredentialsNotFound - Extracted Credentials are not found.
+	ErrExtractedCredentialsNotFound = fmt.Errorf("credentials not found")
```
eriknelson: :+1:

pkg/apb/unbind.go
```diff
@@ -70,6 +70,13 @@ func (e *executor) Unbind(
 			e.actionFinishedWithError(err)
 			return
 		}
+		// Delete the binding extracted credential here.
+		err = runtime.Provider.DeleteExtractedCredential(bindingID, clusterConfig.Namespace)
+		if err != nil {
```
eriknelson: I think we might want to just log this at an info level if it's a miss error instead of finishing with error here too.

shawn-hurley:
@eriknelson I have updated and tested based on our quick conversation.

eriknelson:
@shawn-hurley cool, will test again this morning.

Issue #745, opened 2018-02-07
local_openshift whitelist is failing to exclude non-APBs


mhrivnak:
**What happened**:
I built the broker from master. On startup, it complains about images it finds in the ``local_openshift`` registry:

```
[2018-02-07T17:36:40.471Z] [INFO] - Validating specs...
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[2018-02-07T17:36:40.471Z] [WARNING] - Spec [  ] failed validation for the following reason: [ APB Spec version [] out of bounds 1.0 <= 1.0 ]. It will not be made available.
[...]

PR #781, opened 2018-02-21
Filter everything when whitelist is empty

rthallisey:
We currently filter nothing when blacklist is empty and whitelist is empty. 

Another way of saying this is if I have this whitelist config:
```yaml
    white_list:
      - ".*-apb$"
```
In my registry I have nothing that ends in ```*-apb```.  The result is that everything goes unfiltered.
```
[2018-02-21T12:12:20.868-05:00] [DEBUG] - APBs passing white/blacklist filter:
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/python-36-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/postgresql-95-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/openshift/mongodb-24-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/openshift/jenkins-2-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/ruby-23-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/openshift/ruby-20-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/nodejs-6-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/nginx-110-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/perl-520-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/openshift/wildfly-100-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/mariadb-102-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/openshift/nodejs-010-centos7
[2018-02-21T12:12:20.868-05:00] [DEBUG] - -> docker.io/centos/python-34-centos7
...
```
 
 - Filter everything when nothing is whitelisted

fixes: https://github.com/openshift/ansible-service-broker/issues/745
fixes: https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/issues/215

### DIFF
--- a/pkg/registries/filter.go
+++ b/pkg/registries/filter.go
@@ -192,7 +192,11 @@ func applyMatchSets(
 	filteredVals := []string{}
 	totalSet := toMatchSetT(totalList)
 
-	if len(whiteMatchSet) != 0 && len(blackMatchSet) != 0 {
+	if len(whiteMatchSet) == 0 {
+		// If nothing is whitelisted, filter everything
+		filteredVals = totalList
+		totalSet = nil
+	} else if len(blackMatchSet) != 0 && len(whiteMatchSet) != 0 {
 		// Blacklist matches override white
 		for k := range blackMatchSet {
 			if _, ok := blackMatchSet[k]; ok {
--- a/docs/filtering_apbs.md
+++ b/docs/filtering_apbs.md
@@ -12,7 +12,7 @@ total set of discovered APBs for a given registry, determining matches.
 |     present    |                     allowed                     |                blocked               |
 |:--------------:|:-----------------------------------------------:|:------------------------------------:|
 | only whitelist | matches a regex in list                         | *ANY* APB that does not match        |
-| only blacklist | *ALL* APBs that do not match                    | APBs that match a regex in list      |
+| only blacklist | *No* APBs from the registry                     | *All* APBs from that registry        |
 |  both present  | matches regex in whitelist but NOT in blacklist | APBs that match a regex in blacklist |
 |  None | *No* APBs from the registry | *All* APBs from that registry |
 
--- a/pkg/registries/filter_test.go
+++ b/pkg/registries/filter_test.go
@@ -29,6 +29,7 @@ import (
 
 const testWhitelistFile = "whitelist.yaml"
 const testBlacklistFile = "blacklist.yaml"
+const emptyWhiteListFile = "empty_whitelist.yaml"
 const testBlacklistOverrideFile = "blacklist_override.yaml"
 
 func testGetRegexFromFile(file string) []string {
@@ -103,7 +104,9 @@ func TestOnlyBlacklist(t *testing.T) {
 		blacklist: testGetRegexFromFile(testBlacklistFile)}
 	filter.Init()
 
-	expectedValidNames := []string{
+	expectedValidNames := []string{}
+
+	expectedFilteredNames := []string{
 		"legitimate-postgresql-apb",
 		"legitimate-mediawiki-apb",
 		"foo-apb",
@@ -111,9 +114,6 @@ func TestOnlyBlacklist(t *testing.T) {
 		"rhscl-postgresql-apb",
 		"baz-apb",
 		"foobar-apb",
-	}
-
-	expectedFilteredNames := []string{
 		"totally-not-malicious-apb",
 		"malicious-bar-apb",
 		"specific-blacklist-apb",
@@ -147,6 +147,34 @@ func TestOnlyWhitelist(t *testing.T) {
 		// Not explicitly whitelisted, so should be filtered
 		"baz-apb",
 		"foobar-apb",
+	}
+
+	validNames, filteredNames := filter.Run(testNames())
+
+	expectedTotal := append(expectedValidNames, expectedFilteredNames...)
+	ft.AssertTrue(t, testSetEq(expectedValidNames, validNames))
+	ft.AssertTrue(t, testSetEq(expectedFilteredNames, filteredNames))
+	ft.AssertTrue(t, testSetEq(expectedTotal, testNames()))
+}
+
+func TestEmptyWhitelist(t *testing.T) {
+	filter := Filter{whitelist: testGetRegexFromFile(emptyWhiteListFile),
+		blacklist: []string{}}
+	filter.Init()
+
+	expectedValidNames := []string{}
+
+	expectedFilteredNames := []string{
+		"totally-not-malicious-apb",
+		"malicious-bar-apb",
+		"specific-blacklist-apb",
+		"baz-apb",
+		"foobar-apb",
+		"legitimate-postgresql-apb",
+		"legitimate-mediawiki-apb",
+		"foo-apb",
+		"bar-apb",
+		"rhscl-postgresql-apb",
 	}
 
 	validNames, filteredNames := filter.Run(testNames())
--- a/pkg/registries/testdata/empty_whitelist.yaml
+++ b/pkg/registries/testdata/empty_whitelist.yaml
@@ -1,0 +1,2 @@
+---
+- ".*-nothingwillmatchthis$"


shawn-hurley:
As long as we are okay with the log message:
`[2018-02-21T18:48:12.717Z] [INFO] - APBs filtered by white/blacklist filter:-> docker.io/openshift/wildfly-100-centos7-> docker.io/centos/mysql-57-centos7-> docker.io/centos/nginx-18-centos7-> docker.io/centos/php-56-centos7-> docker.io/centos/python-35-centos7-> docker.io/centos/ruby-24-centos7-> docker.io/openshift/php-55-centos7-> docker.io/openshift/mongodb-24-centos7-> docker.io/centos/postgresql-94-centos7-> docker.io/openshift/wildfly-101-centos7-> docker.io/openshift/wildfly-81-centos7-> docker.io/centos/ruby-23-centos7-> docker.io/centos/php-71-centos7-> docker.io/centos/postgresql-95-centos7-> docker.io/openshift/ruby-20-centos7-> docker.io/openshift/wildfly-90-centos7-> docker.io/openshift/python-33-centos7-> docker.io/openshift/mysql-55-centos7-> docker.io/openshift/jenkins-1-centos7-> docker.io/centos/mongodb-26-centos7-> docker.io/centos/nginx-112-centos7-> registry.centos.org/dotnet/dotnet-20-runtime-centos7-> docker.io/centos/mongodb-34-centos7-> docker.io/centos/nodejs-8-centos7-> docker.io/centos/mariadb-101-centos7-> docker.io/centos/redis-32-centos7-> docker.io/openshift/nodejs-010-centos7-> registry.centos.org/dotnet/dotnet-20-centos7-> docker.io/centos/python-34-centos7-> docker.io/centos/perl-524-centos7-> docker.io/centos/nodejs-6-centos7-> docker.io/centos/ruby-22-centos7-> docker.io/centos/mongodb-32-centos7-> docker.io/centos/nginx-110-centos7-> docker.io/centos/postgresql-96-centos7-> docker.io/openshift/postgresql-92-centos7-> docker.io/centos/php-70-centos7-> docker.io/centos/mariadb-102-centos7-> docker.io/openshift/jenkins-2-centos7-> docker.io/centos/mysql-56-centos7-> docker.io/centos/httpd-24-centos7-> docker.io/centos/python-36-centos7-> docker.io/openshift/perl-516-centos7-> docker.io/centos/python-27-centos7-> docker.io/centos/nodejs-4-centos7-> docker.io/centos/perl-520-centos7`

then I think this is good

eriknelson:
@shawn-hurley is that log actually all one line? Because it's totally unreadable in that form. I do think we should continue to `Info` log the apbs that were filtered, I just want it to be actually useful.

pkg/registries/filter.go
```diff
@@ -192,7 +192,11 @@ func applyMatchSets(
 	filteredVals := []string{}
 	totalSet := toMatchSetT(totalList)
 
-	if len(whiteMatchSet) != 0 && len(blackMatchSet) != 0 {
+	if len(whiteMatchSet) == 0 {
```
eriknelson: I'm concerned with this change. There is an intended use-case for specifying only a blacklist. You are saying you want to accept everything except those that don't pass my blacklist filter.

rthallisey: What I could do is default the white_list to ```".*"``` if only black_list exists.

jmrodri:
My understanding was that we filtered NOTHING, and that you can use the white and black lists to filter things as you choose.

my registry contents

- mediawiki-apb
- postgresql-apb
- mysql
- mariadb
- shitty-apb
- shitty-cool-apb

=========================

whitelist EMPTY
blacklist EMPTY

- mediawiki-apb
- postgresql-apb
- mysql
- mariadb
- shitty-apb
- shitty-cool-apb

=========================

whitelist EMPTY
blacklist shitty*apb

- mediawiki-apb
- postgresql-apb
- mysql
- mariadb

=========================

whitelist -apb
blacklist shitty*apb


- mediawiki-apb
- postgresql-apb

=========================

whitelist -apb
blacklist EMPTY

- mediawiki-apb
- postgresql-apb
- shitty-apb
- shitty-cool-apb

=========================

whitelist EMPTY
blacklist *apb, mysql, mariadb

EVERYTHING FILTERED

shawn-hurley:
> whitelist EMPTY
blacklist EMPTY
> 
> mediawiki-apb
postgresql-apb
mysql
mariadb
shitty-apb
shitty-cool-apb

This is not the case. We default to closed and no apb's show up. apbs are opt in. This was an ask of the security team.

rthallisey:
@jmrodri My view was the opposite. I thought for security purposes we wanted to allow nothing.

eriknelson:
```
whitelist EMPTY
blacklist shitty*apb

    mediawiki-apb
    postgresql-apb
    mysql
    mariadb
```

This is the case I'm concerned with. It's a legitimate use-case to want to blacklist X, but accept everything else. We should be smart enough to ignore `mysql` and `mariadb` since they are not APBs. We need to be able to detect whether or not an image is an APB beyond its name.

shawn-hurley:
@eriknelson if we saying APBs are opt in, which was the ask, then I think it follows that you can't just  add a blacklist and that defaults to letting everything in. I could be wrong on this one but I do think this is the case.

Also re: the log yes that is all one line for me.

rthallisey:
> This is the case I'm concerned with. It's a legitimate use-case to want to blacklist X, but accept everything else. 
>
Not ideal to add an extra config options, but this would do the trick.
```yaml
    black_list:
      - ".*-apb$"
    white_list:
      - ".*"
```

> We should be smart enough to ignore mysql and mariadb since they are not APBs. We need to be able to detect whether or not an image is an APB beyond its name.
>
We check for the APB label, but that is outside the whitelist/blacklist filter

eriknelson:
> APBs are opt in, which was the ask

In an effort to be transparent, there was no original ask or requirement. I personally implemented the filters in the way that I felt appropriate. If you only specify a `blacklist`, I feel that you are expressing you want those filtered, but everything else allowed. I still feel that should be the case, but I'm open to considering otherwise.

jmrodri:
> @jmrodri My view was the opposite. I thought for security purposes we wanted to allow nothing.

So is there currently a security risk that we need to fix? I didn't think there was, I believe this all stems from the removal of the hard-coded `-apb` we had in the code. Prior to that things seemed be okay until someone wanted to add an apb not named `-apb`. 

I still think that leaving them empty shows everything. From an ease of use and configuration. I only want to learn enough regex to lock things down, not to open everything up. 

If I want to extra paranoid in my corporate deployment, I can come up with a naming strategy that I approve. Then whitelist those. And blacklist everything else. If I'm in a development shop, I just leave them empty and see everything.

If the user wants to block everything for some reason, we have given them a mechanism to do that.

eriknelson:
Going to send a mailing list question on the issue.

shawn-hurley:
@eriknelson I think that is correct that you initially implemented in the way described. 

When we were discussing security changes with security team, they asked that we make the APBs default to empty if no whitelist and blacklist.

Here is the PR for that change https://github.com/openshift/ansible-service-broker/pull/411

enj:
So, if you guys are willing to change the current behavior (and possibly break some users), I would go with the `Opt-in Filter Behavior` as defined in #785.  If you believe it will cause too much issue for users, then we may simply to have live with a fail open default config option.

With the `Opt-in Filter Behavior`, a user can always recreate the current behavior by setting whitelist to `.*` and blacklist to whatever they currently have.  The choice is just extremely explicit.  I do not know how the ASB handles config validation, but if you followed OpenShift's logic, the broker would fail to start if `only blacklist` or `None` was set (unless there is some valid reason to have no allowed APBs).

Using `.*` as the whitelist may be a very reasonable option in situations where you have full trust in the source of the APB (i.e Red Hat).  But as others start to create APBs, we should assume they are dangerous since they are effectively black boxes (and it is unreasonable to expect users to look through the source code to make sure the APB is safe).

rthallisey:
Given the feedback on the issue, we going with opt-in.  This PR will enable opt-in so I'll change the docs to reflect this behavior.

jmrodri:
@eriknelson @shawn-hurley docs were updated, please re-review.

rthallisey:
@shawn-hurley @jmrodri @eriknelson Do you think we should backport this?

PR #790, opened 2018-02-24
Bug 1548311 - Return success on depro extcred miss

eriknelson:

Deprovision tries to delete extracted credentials despite the fact that they legitimately may not exist in etcd (in the case of failed provisions), and it reports that deprovision as a failure. Instead, the miss should not be considered a critical error and debug logged.

 - Debug log when a delete of extracted credentials fails during a deprovision, and continue to report success in that case, since it is not considered a critical error.

### DIFF
--- a/pkg/broker/deprovision_subscriber.go
+++ b/pkg/broker/deprovision_subscriber.go
@@ -17,6 +17,7 @@
 package broker
 
 import (
+	"github.com/coreos/etcd/client"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 	"github.com/openshift/ansible-service-broker/pkg/dao"
 )
@@ -77,8 +78,10 @@ func cleanupDeprovision(instance *apb.ServiceInstance, dao *dao.Dao) error {
 	var err error
 	id := instance.ID.String()
 
-	if err = dao.DeleteExtractedCredentials(id); err != nil {
-		log.Error("failed to delete extracted credentials - %v", err)
+	err = dao.DeleteExtractedCredentials(id)
+	if err != nil && client.IsKeyNotFound(err) {
+		log.Infof("Attempted to delete extracted credentials, but key was not found: [%v]", id)
+	} else if err != nil {
 		return err
 	}
 


pkg/broker/deprovision_subscriber.go
```diff
@@ -77,8 +82,10 @@ func cleanupDeprovision(instance *apb.ServiceInstance, dao *dao.Dao) error {
 	var err error
 	id := instance.ID.String()
 
-	if err = dao.DeleteExtractedCredentials(id); err != nil {
-		log.Error("failed to delete extracted credentials - %v", err)
+	err = dao.DeleteExtractedCredentials(id)
+	if err != nil && strings.Contains(err.Error(), etcdKeyNotFoundContents) {
```
jmrodri: Same comment as the other PR, seems like we usually do `IsKeyNotFound` for etcd errors.

```
    if err != nil {
        // no jobs or states to recover, this is OK.
            log.Info("No jobs to recover")
            return "", nil
        }
        return "", err
    }
```

eriknelson: @jmrodri +1. I'm unsure if we want to couple this subscriber etcd specifically, with an etcd `clientNotFound` check. It's probably fine because subscribers are designed to be specific to their use-cases (like persisting etcd data). What do you think?

eriknelson: @jmrodri reading back over this, we definitely don't want a coupling point to etcd, especially with the move to CRDs. I think this needs to be it's own high level "ExtractedCredentialsNotFound" error. @shawn-hurley thoughts?

shawn-hurley: I have to re-write this already because of the move to CRDs. 
I do think that having exposable DAO errors that our dao implementations uses makes sense here and something that is outside of the scope of this PR. 

It will be changed very soon, would rather be able to find all instances by searching for the `IsKeyNotFound` rather then having a one off.

eriknelson: Cool, makes sense. I will update and ping you guys.

eriknelson:
@jmrodri updated to use `client.IsKeyNotFound`.

PR #791, opened 2018-02-24
Bug 1548311 - Return success on depro extcred miss

eriknelson:
Forward port of #790 to master

### DIFF
--- a/pkg/broker/deprovision_subscriber.go
+++ b/pkg/broker/deprovision_subscriber.go
@@ -17,6 +17,7 @@
 package broker
 
 import (
+	"github.com/coreos/etcd/client"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/broker/deprovision_subscriber.go
```diff
@@ -66,8 +71,11 @@ func setFailedDeprovisionJob(dao SubscriberDAO, dmsg JobMsg) {
 }
 
 func cleanupDeprovision(instanceID string, dao SubscriberDAO) error {
-	if err := dao.DeleteExtractedCredentials(instanceID); err != nil {
-		log.Error("failed to delete extracted credentials - %v", err)
+	err := dao.DeleteExtractedCredentials(instanceID)
+
+	if err != nil && strings.Contains(err.Error(), etcdKeyNotFoundContents) {
```
jmrodri: All the other places we do `dao` calls this is how we check for the not found key error:

```
    if err != nil {
        // no jobs or states to recover, this is OK.
            log.Info("No jobs to recover")
            return "", nil
        }
        return "", err
    }
```

The above comes from `broker.go`

pkg/broker/deprovision_subscriber.go
```diff
@@ -66,8 +71,11 @@ func setFailedDeprovisionJob(dao SubscriberDAO, dmsg JobMsg) {
...
+		log.Debugf("Attempted to delete extracted credentials, but key was not found: [%v]", instanceID)
```
jmrodri: Will we ever want to know this in production? That is would we want to see it at INFO level? If not, leave it at debug. Just wondering if this will be useful information or not.

eriknelson: It's probably useful in production, I'll bump it up to INFO.

eriknelson:
@jmrodri updated to use `client.IsKeyNotFound`.

PR #795, opened 2018-02-26
Broker client dao

shawn-hurley:
Implements Broker client dao.
 -  Adds a config value for CRD dao
 - Adds implementation for conversion and dao from CRD to apb package structures

depends-on <PR>
~~#775~~
~~#794~~

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -193,7 +193,7 @@ func CreateApp() App {
 	}
 
 	log.Debug("Connecting Dao")
-	app.dao, err = dao.NewDao()
+	app.dao, err = dao.NewDao(app.config)
 	if err != nil {
 		log.Error(err.Error())
 		os.Exit(1)
@@ -390,26 +390,29 @@ func initClients(c *config.Config) error {
 	// method on the app. Forces developers at authorship time to think about
 	// dependencies / make sure things are ready.
 	log.Notice("Initializing clients...")
-	log.Debug("Trying to connect to etcd")
 
-	// Intialize the etcd configuration
-	clients.InitEtcdConfig(c)
-	etcdClient, err := clients.Etcd()
-	if err != nil {
-		return err
+	if strings.ToLower(c.GetString("dao.type")) != "crd" {
+
+		log.Debug("Trying to connect to etcd")
+		// Intialize the etcd configuration
+		clients.InitEtcdConfig(c)
+		etcdClient, err := clients.Etcd()
+		if err != nil {
+			return err
+		}
+
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		defer cancelFunc()
+
+		version, err := etcdClient.GetVersion(ctx)
+		if err != nil {
+			return err
+		}
+
+		log.Infof("Etcd Version [Server: %s, Cluster: %s]", version.Server, version.Cluster)
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
-
-	version, err := etcdClient.GetVersion(ctx)
-	if err != nil {
-		return err
-	}
-
-	log.Infof("Etcd Version [Server: %s, Cluster: %s]", version.Server, version.Cluster)
-
-	_, err = clients.Kubernetes()
+	_, err := clients.Kubernetes()
 	if err != nil {
 		return err
 	}
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -606,10 +606,14 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 	// if err is not nil, we will just bubble that up
 
 	if si, err := a.dao.GetServiceInstance(instanceUUID.String()); err == nil {
+		log.Infof("Here - %v", si.ID.String())
+		log.Infof("Here - %v", serviceInstance.ID.String())
 		// This will use the package to make sure that if the type is changed
 		// away from []byte it can still be evaluated.
 		if uuid.Equal(si.ID, serviceInstance.ID) {
+			log.Infof("Here - %v", serviceInstance)
 			if reflect.DeepEqual(si.Parameters, serviceInstance.Parameters) {
+				log.Infof("Here - %v reflect deep equals", serviceInstance)
 				alreadyInProgress, jobToken, err := a.isJobInProgress(serviceInstance, apb.JobMethodProvision)
 				if err != nil {
 					return nil, fmt.Errorf("An error occurred while trying to determine if a provision job is already in progress for instance: %s", serviceInstance.ID)
@@ -625,6 +629,7 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 			return nil, ErrorDuplicate
 		}
 	}
+	log.Infof("Here not found service instance. - %v", serviceInstance)
 
 	//
 	// Looks like this is a new provision, let's get started.
@@ -730,6 +735,7 @@ func (a AnsibleBroker) isJobInProgress(instance *apb.ServiceInstance,
 	method apb.JobMethod) (bool, string, error) {
 
 	allJobs, err := a.dao.GetSvcInstJobsByState(instance.ID.String(), apb.StateInProgress)
+	log.Infof("All Jobs for instance: %v in state:  %v - \n%#v", instance.ID, apb.StateInProgress, allJobs)
 	if err != nil {
 		return false, "", err
 	}
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -605,14 +605,10 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 	// if err is not nil, we will just bubble that up
 
 	if si, err := a.dao.GetServiceInstance(instanceUUID.String()); err == nil {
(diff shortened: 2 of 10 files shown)


eriknelson:
@shawn-hurley merged both blocker PRs, so this should be good for a rebase.

pkg/broker/broker.go
```diff
@@ -152,7 +151,7 @@ func NewAnsibleBroker(dao dao.Dao,
 func (a AnsibleBroker) GetServiceInstance(instanceUUID uuid.UUID) (apb.ServiceInstance, error) {
 	instance, err := a.dao.GetServiceInstance(instanceUUID.String())
 	if err != nil {
-		if client.IsKeyNotFound(err) {
+		if a.dao.IsNotFoundError(err) {
```
jmrodri: +1 makes sense.

pkg/app/app.go
```diff
@@ -390,26 +390,29 @@ func initClients(c *config.Config) error {
...
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
+		log.Debug("Trying to connect to etcd")
+		// Intialize the etcd configuration
+		clients.InitEtcdConfig(c)
+		etcdClient, err := clients.Etcd()
+		if err != nil {
+			return err
+		}
```
jmrodri: :+1: took a while to figure out what really changed here, but it was the if statement that caused the indentation of the block of code. looks good.

pkg/app/app.go
```diff
@@ -390,26 +390,29 @@ func initClients(c *config.Config) error {
...
 	log.Notice("Initializing clients...")
-	log.Debug("Trying to connect to etcd")
 
-	// Intialize the etcd configuration
-	clients.InitEtcdConfig(c)
-	etcdClient, err := clients.Etcd()
-	if err != nil {
-		return err
-	}
+	if strings.ToLower(c.GetString("dao.type")) != "crd" {
```
jmrodri: What happens if it isn't crd? What is the other valid value?

templates/k8s-ansible-service-broker.yaml.j2
```diff
@@ -319,3 +319,245 @@ metadata:
...
+                plateform:
+                  type: string
+                namespace:
+                  type: string
+            parameters:
+              type: string
+            bindingIds:
+              type: array
+              items:
+                type: string
```
jmrodri: :open_mouth:

OMG

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -93,6 +93,10 @@ objects:
     attributeRestrictions: null
     resources: ["networkpolicies"]
     verbs: ["create", "delete"]
+  - apiGroups: ["automationbroker.io"]
+    attributeRestrictions: null
+    resources: ["bundles", "jobstates", "servicebindings", "serviceinstances"]
+    verbs: ["*"]
```
jmrodri: :+1:

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -402,6 +406,242 @@ objects:
       ${{BROKER_AUTH}}
     caBundle: ${BROKER_CA_CERT}
 
+# CRDs for the broker.
+- apiVersion: apiextensions.k8s.io/v1beta1
```
jmrodri: OMG

pkg/app/app.go
shawn-hurley: All other values will mean we use `etcd`. This is so every old config should still work as expected with `etcd` as it's DAO.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -402,6 +406,242 @@ objects:
...
+  kind: CustomResourceDefinition
+  metadata:
+    name: bundles.automationbroker.io
+  spec:
+    group: automationbroker.io
+    version: v1
```
eriknelson: Is there any reason to consider these v1alpha or v1beta?

pkg/dao/dao.go
```diff
@@ -18,15 +18,21 @@ package dao
...
+	crd "github.com/openshift/ansible-service-broker/pkg/dao/crd"
 	etcd "github.com/openshift/ansible-service-broker/pkg/dao/etcd"
 	logutil "github.com/openshift/ansible-service-broker/pkg/util/logging"
 )
 
 var log = logutil.NewLog()
 
 // NewDao - Create a new Dao object
-func NewDao() (Dao, error) {
+func NewDao(c *config.Config) (Dao, error) {
```
eriknelson: This is great. :+1:

pkg/dao/crd/dao.go
```diff
@@ -0,0 +1,306 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package dao
+
+import (
+	"fmt"
+
+	automationbrokerv1 "github.com/automationbroker/broker-client-go/client/clientset/versioned/typed/automationbroker.io/v1"
```
eriknelson: I'm concerned that we're starting to mix `automationbroker` and `ansible-service-broker` language now in the codebase. `automationbroker` is the correct choice IMO, but I think we need to get this issue captured and start to move forward with the rebrand.

pkg/dao/crd/dao.go
```diff
@@ -0,0 +1,306 @@
...
+
+// GetSpec - Retrieve the spec from the k8s API.
+func (d *Dao) GetSpec(id string) (*apb.Spec, error) {
+	log.Debugf("get spec: %v", id)
+	s, err := d.client.Bundles(d.namespace).Get(id, metav1.GetOptions{})
+	if err != nil {
+		log.Errorf("unable to get bundle from k8s api - %v", err)
+		return nil, err
+	}
+	return bundleToSpec(s.Spec, s.GetName()), nil
```
eriknelson: So it's expected that the dao layer will need to transform between the ASB types and their CRD counterparts for most of these, so the rest of the codebase is getting objects they expect?

pkg/dao/crd/dao.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return convertServiceInstanceToAPB(servInstance.Spec, spec, servInstance.GetName()), nil
+}
+
+// SetServiceInstance - Set service instance for an id in the kvp API.
+func (d *Dao) SetServiceInstance(id string, serviceInstance *apb.ServiceInstance) error {
+	log.Debugf("set service instance: %v", id)
+	spec := convertServiceInstanceToCRD(serviceInstance)
+	if si, err := d.client.ServiceInstances(d.namespace).Get(id, metav1.GetOptions{}); err == nil {
+		log.Debugf("updateing service instance: %v", id)
```
eriknelson: `updateing` -> `updating`

pkg/dao/crd/dao.go
```diff
@@ -0,0 +1,306 @@
...
+	return err
+}
+
+// SetState - Set the Job State in the kvp API for id.
+func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
+	log.Debugf("set job state for instance: %v token: %v", instanceID, state.Token)
+	j := convertJobStateToCRD(&state)
+	if js, err := d.client.JobStates(d.namespace).Get(state.Token, metav1.GetOptions{}); err == nil {
+		js.Spec = j
+		js.ObjectMeta.Labels[jobStateLabel] = fmt.Sprintf("%v", convertStateToCRD(state.State))
```
eriknelson: Why are we setting the labels here? Selector filtering? That's actually pretty slick.

pkg/dao/etcd/dao.go
```diff
@@ -333,6 +333,11 @@ func (d *Dao) GetStateByKey(key string) (apb.JobState, error) {
 	return state, nil
 }
 
+// IsNotFoundError - Will determine if an error is a key is not found error.
+func (d *Dao) IsNotFoundError(err error) bool {
```
eriknelson: Nice this got pushed into the dao.

templates/deploy-ansible-service-broker.template.yaml
shawn-hurley: it is not possible to have more than 1 version for an API in CRDs so it is not like we could bump up and start using v1 and still have v1beta1 around. I think because of that, just make it easy and call them v1.

pkg/dao/crd/dao.go
shawn-hurley: Yes, that is the plan and what I was hoping to do here.

shawn-hurley:
@eriknelson as far as testing against CRDs on master branch. I am slightly worried about switching over to CRDs right now. Mostly, this has worked in my testing of the actions but I am gun shy. If someone else's tests and thinks it is good to go, then I say we change the default template to use crds.

eriknelson:
I've been banging on this for the last hour, `make ci` is good, I ran through a number of scenarios I know were regression bugs, failure cases, it's handling them fine. I can also read the resources with `oc get`, which is actually making this much more friendly for broker developers debugging compared to raw etcd. Big +1, nice work @shawn-hurley.

jmrodri:
@shawn-hurley Tested with my crash simulation patch (i.e. random adapter failing). I see that the bundles all get deleted and slowly populate as they get parsed. The final state was right, not sure if this is just being picky or not.

eriknelson:
I have scenarios I run knowing they were existing bugs. We should get these
captured in make ci somehow so they're always checked.

PR #799, opened 2018-02-28
Parametize namespace in deployment template and update the 'run latest build' script

johnkim76:
Parameterize the ASB namespace in the deployment template and changed the run_latest_build script so that it would be easier to update AWS broker templates and scripts

 - replaced 'ansible-service-broker' with ${NAMESPACE} in the deployment template where appropriate
 - created a 'deploy_broker.sh' script to launch the broker in an existing cluster
 - modified the 'run_latest_build.sh' script to utilize the 'deploy_broker.sh' script

To test the run script BEFORE merging this PR, the TEMPLATE_URL should be defined:
$ TEMPLATE_URL=<link-to-updated-template-file> ./run_latest_build.sh

### DIFF
--- a/scripts/deploy_broker.sh
+++ b/scripts/deploy_broker.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+#
+# Helper script for deploying the Broker to an existing OpenShift cluster
+# using an OpenShift template.
+#
+
+#
+# VERIFY and UPDATE the variables below
+#
+CLUSTER_ADMIN_USER="system:admin"
+TEMPLATE_URL=${TEMPLATE_URL:-"https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/deploy-ansible-service-broker.template.yaml"}
+DOCKERHUB_ORG=${DOCKERHUB_ORG:-"ansibleplaybookbundle"} # DocherHub org where APBs can be found, default 'ansibleplaybookbundle'
+BROKER_IMAGE=${BROKER_IMAGE:-"ansibleplaybookbundle/origin-ansible-service-broker:latest"}
+ENABLE_BASIC_AUTH=${ENABLE_BASIC_AUTH:-"false"}
+PROJECT_NAME=${PROJECT_NAME:-"ansible-service-broker"}
+
+#
+# Login as the CLUSTER_ADMIN_USER
+#
+oc login -u ${CLUSTER_ADMIN_USER}
+
+#
+# Get the BROKER_CA_CERT
+#
+BROKER_CA_CERT=`oc get secret -n kube-service-catalog -o go-template='{{ range .items }}{{ if eq .type "kubernetes.io/service-account-token" }}{{ index .data "service-ca.crt" }}{{end}}{{"\n"}}{{end}}' | awk NF | tail -n 1`
+if [ "${BROKER_CA_CERT}" == "" ]; then
+    echo -e "\nUnable to set the BROKER_CA_CERT variable!"
+    echo -e "Please VERIFY that CLUSTER_ADMIN_USER is set to a user with cluster admin privileges\n"
+    exit
+fi
+
+#
+# creating ${PROJECT_NAME} project
+#
+oc new-project ${PROJECT_NAME}
+
+# Creating openssl certs to use.
+mkdir -p /tmp/etcd-cert
+openssl req -nodes -x509 -newkey rsa:4096 -keyout /tmp/etcd-cert/key.pem -out /tmp/etcd-cert/cert.pem -days 365 -subj "/CN=asb-etcd.ansible-service-broker.svc"
+openssl genrsa -out /tmp/etcd-cert/MyClient1.key 2048 \
+&& openssl req -new -key /tmp/etcd-cert/MyClient1.key -out /tmp/etcd-cert/MyClient1.csr -subj "/CN=client" \
+&& openssl x509 -req -in /tmp/etcd-cert/MyClient1.csr -CA /tmp/etcd-cert/cert.pem -CAkey /tmp/etcd-cert/key.pem -CAcreateserial -out /tmp/etcd-cert/MyClient1.pem -days 1024
+
+ETCD_CA_CERT=$(cat /tmp/etcd-cert/cert.pem | base64)
+BROKER_CLIENT_CERT=$(cat /tmp/etcd-cert/MyClient1.pem | base64)
+BROKER_CLIENT_KEY=$(cat /tmp/etcd-cert/MyClient1.key | base64)
+
+curl -s $TEMPLATE_URL \
+  | oc process \
+  -n ${PROJECT_NAME} \
+  -p DOCKERHUB_ORG="$DOCKERHUB_ORG" \
+  -p ENABLE_BASIC_AUTH="$ENABLE_BASIC_AUTH" \
+  -p ETCD_TRUSTED_CA_FILE=/var/run/etcd-auth-secret/ca.crt \
+  -p BROKER_CLIENT_CERT_PATH=/var/run/asb-etcd-auth/client.crt \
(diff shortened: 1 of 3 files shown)


scripts/deploy_broker.sh
```diff
@@ -0,0 +1,80 @@
...
+# Helper script for deploying the Broker to an existing OpenShift cluster
+# using an OpenShift template.
+#
+
+#
+# VERIFY and UPDATE the variables below
+#
+CLUSTER_ADMIN_USER="system:admin"
+TEMPLATE_URL=${TEMPLATE_URL:-"https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/deploy-ansible-service-broker.template.yaml"}
+DOCKERHUB_ORG=${DOCKERHUB_ORG:-"ansibleplaybookbundle"} # DocherHub org where APBs can be found, default 'ansibleplaybookbundle'
```
jmrodri: TYPO in the comment: DocherHub should be DockerHub

scripts/deploy_broker.sh
```diff
@@ -0,0 +1,80 @@
...
+#
+# Then login as 'developer'/'developer' to WebUI
+# Create a project
+# Deploy mediawiki to new project (use a password other than
+#   admin since mediawiki forbids admin as password)
+# Deploy PostgreSQL(ABP) to new project
+# After they are up
+# Click 'Create Binding' on the kebab menu for Mediawiki,
+#   select postgres
+# Click deploy on mediawiki, after it's redeployed access webui
```
jmrodri: What's the purpose of these comments?

scripts/run_latest_build.sh
```diff
@@ -42,6 +45,13 @@ ROUTING_SUFFIX="${HOSTNAME}"
 ORIGIN_IMAGE=${ORIGIN_IMAGE:-"docker.io/openshift/origin"}
 ORIGIN_VERSION=${ORIGIN_VERSION:-"latest"}
 
+CLUSTER_ADMIN_USER="system:admin"
+TEMPLATE_URL=${TEMPLATE_URL:-"https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/deploy-ansible-service-broker.template.yaml"}
+DOCKERHUB_ORG=${DOCKERHUB_ORG:-"ansibleplaybookbundle"} # DocherHub org where APBs can be found, default 'ansibleplaybookbundle'
```
jmrodri: TYPO: DocherHub should be DockerHub

scripts/run_latest_build.sh
```diff
@@ -58,64 +68,9 @@ oc cluster up --image=${ORIGIN_IMAGE} \
...
-  -p ETCD_TRUSTED_CA="$ETCD_CA_CERT" \
-  -p BROKER_CLIENT_CERT="$BROKER_CLIENT_CERT" \
-  -p BROKER_CLIENT_KEY="$BROKER_CLIENT_KEY" \
-  -p NAMESPACE=ansible-service-broker \
-  $VARS -f - | oc create -f -
-if [ "$?" -ne 0 ]; then
-  echo "Error processing template and creating deployment"
-  exit
-fi
+./deploy_broker.sh
```
jmrodri: I'm now confused. Looks like `deploy_broker.sh` does all of the work, then why do we have both `run_latest_build.sh` and `deploy_broker.sh`?

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -45,7 +45,7 @@ objects:
   kind: ServiceAccount
   metadata:
     name: asb
-    namespace: ansible-service-broker
+    namespace: "${NAMESPACE}"
```
jmrodri: is there a catasb related PR to allow overriding this? I mean most of the variables have a corresponding value in `all.yml`.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -112,14 +112,14 @@ objects:
   metadata:
     name: access-asb-role
   rules:
-  - nonResourceURLs: ["${BROKER_URL_PREFIX}", "${BROKER_URL_PREFIX}/*"]
+  - nonResourceURLs: ["/${BROKER_URL_PREFIX}", "/${BROKER_URL_PREFIX}/*"]
```
shawn-hurley: Pretty sure if you do this you need to update: https://github.com/openshift/ansible-service-broker/pull/799/files#diff-54e88cebcb04cc49b8796b2f774cb6c7R425

scripts/deploy_broker.sh
johnkim76: this is just left overs from previous scripts.. I can update that.

scripts/deploy_broker.sh
johnkim76: thanks, I'll update

scripts/run_latest_build.sh
johnkim76: thanks, I'll update

shawn-hurley: I would prefer to not add variables to this script for changing how it runs. My understanding of this script has always been that it is the sane defaults that we decide and not meant to be an everything to everyone script. These changes make it more and more overlapping with catasb IMO.

scripts/run_latest_build.sh
johnkim76: 'run_latest_build.sh' sets up the cluster and deploys the broker via template.  The 'deploy_broker.sh' simply launches the broker in an existing cluster. The purpose of these scripts is to make it easier to update their AWS broker counter parts when the template and the way to deploy the broker changes in this repo.  Currently AWS desires to have a script to only launch the broker in an existing cluster for people who may already have their cluster environment up and running. So currently two sets of scripts are being maintained.  This PR would make it much easier to maintain the deployment of the broker only script.

templates/deploy-ansible-service-broker.template.yaml
johnkim76: these scripts would be outside of catasb. these are essentially a 'striped down' versions of the catasb functionality.

templates/deploy-ansible-service-broker.template.yaml
johnkim76: In my testing this works.

However, I've also changed the way I pass in the parameter here:
https://github.com/openshift/ansible-service-broker/pull/799/files#diff-45df7e79b5e14d6545c816ae8b2dc180R61

and also made changes here:
https://github.com/openshift/ansible-service-broker/pull/799/files#diff-54e88cebcb04cc49b8796b2f774cb6c7L400

johnkim76: it's worth noting that with the above changes, I've also had to always add this line to the AWS template counter part here:
https://github.com/openshift/ansible-service-broker/pull/799/files#diff-54e88cebcb04cc49b8796b2f774cb6c7R348

shawn-hurley: Because of your testing with the change you are setting the correct value.  If you were to run this with default parameter for BROKER_URL then it would cause issue.

scripts/run_latest_build.sh
johnkim76: If one wishes to 'run_latest_build.sh' with different variable settings, they would need to modify the script itself.

I've made those variable user-changeable at run-time, so that the user would not have to manually edit the script to change these values.  Again, this may not be necessary, but it seemed to be a desired feature in the AWS side and it makes it easier to update these scripts when they're modified.

I can remove here if it's really not desired.

johnkim76:
I have so far tested these in local env by themselves.
I will test this with Catasb to see if any further changes need to be made

templates/deploy-ansible-service-broker.template.yaml
johnkim76: ahhh aplogies.. my fault.  I will update that

PR #800, opened 2018-02-28
parameterize the 'ansible-service-broker' namespace w/ NAMESPACE var

johnkim76:
use the NAMESPACE var in the deployment template where applicable.

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -45,7 +45,7 @@ objects:
   kind: ServiceAccount
   metadata:
     name: asb
-    namespace: ansible-service-broker
+    namespace: "${NAMESPACE}"
 
 - apiVersion: rbac.authorization.k8s.io/v1beta1
   kind: ClusterRoleBinding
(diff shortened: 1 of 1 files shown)


templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -112,14 +112,14 @@ objects:
   metadata:
     name: access-asb-role
   rules:
-  - nonResourceURLs: ["${BROKER_URL_PREFIX}", "${BROKER_URL_PREFIX}/*"]
+  - nonResourceURLs: ["/${BROKER_URL_PREFIX}", "/${BROKER_URL_PREFIX}/*"]
```
eriknelson: What's the necessity for this change?

johnkim76: I needed to make changes in other areas in the template as well to include the '/'.  It's the way to make consistent with how the 'run_latest.build.sh' script specifies the NAMESPACE parameter (without '/').

johnkim76: I will update

PR #804, opened 2018-03-01
Force canary image be built from release-1.1

djzager:
Make it so that canary images built from this branch actually contain the code from this branch.


djzager:
Once this PR is merged, I'll go ahead and update the `release-1.1` and `v3.9` images on dockerhub.

PR #806, opened 2018-03-01
Bug 1550385 - Add .default to proxy docs NO_PROXY

eriknelson:


docs/proxy.md
```diff
@@ -26,9 +26,8 @@ Some important prerequisites necessary for running the broker behind a proxy:
 **NO_PROXY**
 
 The cluster must be configured to *not* proxy internal cluster requests. This
-is typically configured with a `NO_PROXY` setting of ".cluster.local,.svc", in addition
-to any other desired `NO_PROXY` settings. This is because the broker must be able
```
shawn-hurley: Did you mean to delete the reasoning for the no proxy settings?

eriknelson: Yes, it's not really the only reason and won't be entirely accurate soon.

PR #808, opened 2018-03-01
Adds Service Bundle contract document

mhrivnak:
This document captures a detailed description of the existing contract between a Service Bundle and the Automation Broker.

Please review how accurately this represents the current relationship between a Service Bundle and the Automation Broker as implemented today. Please do not (yet) review the technical merits of this design. Once this document is reviewed and accepted as an accurate representation of reality, a next step will be to review the design on technical grounds and look for opportunities to improve it.

### DIFF
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -1,7 +1,7 @@
 # Service Bundle Specification
 
 A Service Bundle is a container image that the broker can use to manage
-deployment of a service in a cluster. It adheres to the properties and
+the deployment of a service in a cluster. It adheres to the properties and
 behaviors described below.
 
 ## Label
@@ -90,7 +90,7 @@ Example of JSON document passed to __provision__:
 
 * cluster: the type of cluster on which the service bundle is being run
   (currently either ‘openshift’ or ‘kubernetes’).
-* _apb_plan_id: the name of a “plan”, as defined through the OSB API, that is
+* _apb_plan_id: the name of a “plan”, as defined by the OSB API, that is
   available on the service class.
 * _apb_service_class_id: the “service_id”, as defined through the OSB API, that
   is being used for the current action.
@@ -99,7 +99,7 @@ Example of JSON document passed to __provision__:
 * namespace: the target namespace in which resources should be created,
   deleted, or acted upon.
 
-Other keys represent parameter names from the Plan, the values for which have
+Other keys represent parameter names from the Plan, the values of which have
 the type specified in the Plan.
 
 Example of JSON document passed to __bind__:
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -192,10 +192,11 @@ operation.
 
 The container will be run in a pod by itself with ``restartPolicy: Never``. The
 pod will run in a sandbox namespace created by the broker for that specific
-operation only. The namespace will be deleted at some point after the pod
-finishes.
+operation only. The namespace is not intended to persist, and in normal
+operation it will be deleted as soon as the pod completes. The broker can
+optionally be configured to retain the namespace for debugging purposes.
 
-APB is run with a service account granted access to the ``POD_NAMESPACE`` and
+The pod is run with a service account granted access to the ``POD_NAMESPACE`` and
 to the “namespace” that the pod is to deploy to. Networks between the two
 namespaces are tied together so that you can reliably find your services in the
 target namespace. That only works for
@@ -205,8 +206,12 @@ or for multi tenant openshift-sdn.
 ### Proxy
 
 If there are proxy settings for the broker, they will be forwarded to the pod
-by setting environment variables for ``HTTP_PROXY``, ``HTTPS_PROXY``, and
-``NO_PROXY``. 
+by setting environment variables for ``HTTP_PROXY``, ``HTTPS_PROXY``,
+``NO_PROXY``, ``http_proxy``, ``https_proxy``, and ``no_proxy``.
+
+No action is taken to apply proxy settings to pods that a service bundle
+creates. Each service bundle is repsonsible for configuring its provisioned
+resources to use a proxy as appropriate.
 
 ### Service Account
 
@@ -215,10 +220,10 @@ Details of a kubernetes service account are mounted at
 
 Keys contained in the Secret:
 
-* ca.crt: a base64-encoded CA certificate that was used to sign the certificate
-  used by the broker to serve its API.
+* ca.crt: a base64-encoded CA certificate
 * service-ca.crt: a base64-encoded CA certificate that was used to sign the
-  certificate used by etcd to serve its API.
+  certificate used by the broker to serve its API and the certificate used by
+  etcd to serve its API.
 * namespace: a base64-encoded string containing the namespace the service
   bundle is running in.
 * token: a base64-encoded string used to authenticate to the kubernetes API.
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -216,15 +216,30 @@ resources to use a proxy as appropriate.
 ### Service Account
 
 Details of a kubernetes service account are mounted at
-``/var/run/secrets/kubernetes.io/serviceaccount``
+``/var/run/secrets/kubernetes.io/serviceaccount`` by Kubernetes. That mount is
+documented [here](https://kubernetes.io/docs/admin/service-accounts-admin/).
 
-Keys contained in the Secret:
+Clients running in a Service Bundle will usually need a ``.kube/config`` file
+such as the following to use the service account:
 
-* ca.crt: a base64-encoded CA certificate
-* service-ca.crt: a base64-encoded CA certificate that was used to sign the
-  certificate used by the broker to serve its API and the certificate used by
-  etcd to serve its API.
-* namespace: a base64-encoded string containing the namespace the service
-  bundle is running in.
-* token: a base64-encoded string used to authenticate to the kubernetes API.
+```
+apiVersion: v1
+clusters:
+- cluster:
+    certificate-authority: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+    server: https://kubernetes.default:443
+  name: kubernetes
+contexts:
+- context:
+    cluster: kubernetes
+    user: apb-user
+  name: /kubernetes/apb-user
+current-context: /kubernetes/apb-user
+kind: Config
+preferences: {}
+users:
+- name: apb-user
+  user:
+    tokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+```
 
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -1,0 +1,225 @@
+# Service Bundle Specification
+
+A Service Bundle is a container image that the broker can use to manage
+deployment of a service in a cluster. It adheres to the properties and
+behaviors described below.
+
+## Label
+
+Every service bundle container image has a label ‘com.redhat.apb.spec’ that
+contains a base64 encoded string. This label holds [important
+metadata](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#apb-spec-file)
+related to the service bundle. Example:
+
+```
+version: 1.0        # The version of the Service Bundle Spec being conformed to
+name: example-apb   # The name of the Service Bundle
+description:        # A short description of the Service Bundle
+bindable: True      # Whether this Service Bundle can be bound to other services
+async: optional     # Does this Service Bundle support asynchronous provision
+metadata: 
+  documentationUrl: <link to documentation>
+  imageUrl: <link to URL of image>
+  dependencies:
+  - '<registry>/<organization>/<dependency-name-1>'   
+  - '<registry>/<organization>/<dependency-name-2>'
+  displayName: Example App (APB)
+  longDescription: A longer description of what this APB does
+  providerDisplayName: "Red Hat, Inc."
+plans: # An array of plans supported by this Service Bundle (must have at least 1)
+  - name: default
+    description: A short description of what this plan does
+    free: true
+    metadata:
+      displayName: Default
+      longDescription: A longer description of what this plan deploys
+      cost: $0.00
+    Parameters: # Parameters for this plan
+      - name: parameter_one
+        required: true
+        default: foo_string
+        type: string
+        title: Parameter One
+        maxlength: 63
+      - name: parameter_two
+        required: true
+        default: true
+        title: Parameter Two
+        type: boolean
+```
+
+## Input
+
+The container is run with three arguments.
+
+* Action: a single word such as “provision”. The full list is below.
+* --extra-vars: This is ansible-specific and always has exactly the value
+  “--extra-vars”. An Ansible Playbook Bundle uses this and the third argument
+  as-is when calling ansible by appending them to the ansible command.
+* JSON: a json document of information that is useful to the bundle. Currently
+  it is expressed in a form that is useful for populating variables in ansible
+  that can be used by roles.
+
+### Actions
+
+Possible actions include
+* provision
+* deprovision
+* bind
+* unbind
+* update
+* test
+
+### JSON
+Example of JSON document passed to __provision__:
+
+```
+{
+    "cluster": "openshift",
+    "_apb_plan_id": "default",
+    "_apb_service_class_id": "c23ec213bb8dea1577230c5ce005b9c2",
+    "_apb_service_instance_id": "54636ad3-0378-49a1-a494-f97d0a0daf8e",
+    "mediawiki_admin_pass": "pass",
+    "mediawiki_admin_user": "admin",
+    "mediawiki_db_schema": "mediawiki",
+    "mediawiki_site_lang": "en",
+    "mediawiki_site_name": "MediaWiki",
+    "namespace": "bazproject"
+}
+```
+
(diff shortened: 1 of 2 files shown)


docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+__Binding Credentials__ - If a service bundle saves such a secret during the bind
+operation, the broker saves the passed data as “binding credentials”. These
+credentials get returned to the client that initiated that specific binding
+operation.
+
+## Runtime
+
+The container will be run in a pod by itself with ``restartPolicy: Never``. The
+pod will run in a sandbox namespace created by the broker for that specific
+operation only. The namespace will be deleted at some point after the pod
```
shawn-hurley: If the pod fails OR certain parameters are set the namespace and pod could not be deleted. 

I don't know if it matters for this document.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+finishes.
+
+APB is run with a service account granted access to the ``POD_NAMESPACE`` and
```
shawn-hurley: Should we also discuss the how the permissions are given to the APB? 

Eventually, APB authors are going to need to know the permissions they are running at right?

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+``NO_PROXY``. 
+
+### Service Account
+
+Details of a kubernetes service account are mounted at
+``/var/run/secrets/kubernetes.io/serviceaccount``
+
+Keys contained in the Secret:
+
+* ca.crt: a base64-encoded CA certificate that was used to sign the certificate
```
shawn-hurley: My understanding is that the `service-ca.crt is used to sign the certificate used by the broker to server its API as well as etcds API

docs/service-bundle.md
dymurray: Yeah I agree with Shawn. Worth calling out in this document.

mhrivnak: Good call. Would you mind supplying that info? Or pointing me to the right place where I can find it?

shawn-hurley: Permission for the service account comes from the `openshift.sandbox_role` from the config. The default for the template is `edit` but I think that most things deploy with this set as `admin` IIRC. here is a link to the roles https://kubernetes.io/docs/admin/authorization/rbac/#user-facing-roles

docs/service-bundle.md
mhrivnak: You appear to be right based on this: https://github.com/openshift/ansible-service-broker/blob/master/scripts/run_latest_build.sh#L89

I'm not sure what to do with this extra ``ca.crt`` that we don't know the purpose of.

docs/service-bundle.md
mhrivnak: I was trying to make the point that even if the namespace sticks around for a while, it is not intended to be in any way permanent. An administrator can make the broker keep it for one reason or another so they can inspect it, but the normal expectation should be that it is transient.

Maybe I'll add another sentence explaining ^.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
+# Service Bundle Specification
+
+A Service Bundle is a container image that the broker can use to manage
+deployment of a service in a cluster. It adheres to the properties and
+behaviors described below.
+
+## Label
+
+Every service bundle container image has a label ‘com.redhat.apb.spec’ that
```
eriknelson: Has anyone asked the question: Is an APB considered one of many types of service bundles that simply adheres to the contract, and chooses to implement its behavior in Ansible? I think this label may need to be more generic.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+        type: boolean
+```
+
+## Input
+
+The container is run with three arguments.
+
+* Action: a single word such as “provision”. The full list is below.
+* --extra-vars: This is ansible-specific and always has exactly the value
+  “--extra-vars”. An Ansible Playbook Bundle uses this and the third argument
```
eriknelson: This happens to be ansible specific due to our heritage, but do we want to express arguments in a more generic way and have an apb-base perform a transform before delivering them to the playbook?

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+  as-is when calling ansible by appending them to the ansible command.
+* JSON: a json document of information that is useful to the bundle. Currently
```
eriknelson: I honestly didn't know about this. If it's possible to pass in additional JSON, is that just a 3rd raw string? An example could be very helpful here for all the permutations.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+    "mediawiki_db_schema": "mediawiki",
+    "mediawiki_site_lang": "en",
+    "mediawiki_site_name": "MediaWiki",
+    "namespace": "bazproject"
+}
+```
+
+* cluster: the type of cluster on which the service bundle is being run
+  (currently either ‘openshift’ or ‘kubernetes’).
+* _apb_plan_id: the name of a “plan”, as defined through the OSB API, that is
```
eriknelson: `_apb` prefixed params need to be reconsidered in a world of generic bundles. The prefix should probably be changed to be agnostic.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+* 8 - action not implemented/supported
+
+### Encoded Credentials
+
+Binding credentials may be returned during a provision or bind operation by
+creating a named secret (POD_NAME) in the sandbox namespace (POD_NAMESPACE).
+The secret should have a single key called “fields” whose value is a
+base64-encoded JSON document.
+
+Ansible Playbook Bundles use the “asb_encode_binding” module to do this.
```
eriknelson: Again, language here is not generic.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+represent any data the APB wants to return to the broker and ultimately to a
+client that retrieves a binding.
+
+__Provision Credentials__ - If a service bundle saves such a secret during the
+provision operation, the broker saves the passed data as “provision
+credentials”. That data gets passed into all further actions as part of the
+JSON document under the key ``_apb_provision_creds``, as a sub-document. As such,
+it can be useful for making root-level credentials created during provision
+available to other operations that need to use them. If a request for non-async
+binding is made and the bundle does not run, any “provision credentials” that
```
eriknelson: I think this is an artefact of our development and not something that belongs as part of a service bundle contract. It should be technically possible to do a sync bind and run an APB, IMO.

docs/service-bundle.md
eriknelson: It's not always the case that the namespace or pod is deleted, it should at least be called out that it's possible to keep them regardless of the outcome, as a lot of developers working with this are going to be interested in that fact.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+APB is run with a service account granted access to the ``POD_NAMESPACE`` and
+to the “namespace” that the pod is to deploy to. Networks between the two
+namespaces are tied together so that you can reliably find your services in the
+target namespace. That only works for
+[NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+or for multi tenant openshift-sdn.
+
+### Proxy
+
+If there are proxy settings for the broker, they will be forwarded to the pod
```
eriknelson: This is correct, but doesn't entirely address the story. We also include lowercased variants, which is actually technically important because there are a lot of ansible modules (more specifically, python libraries) that are utterly inconsistent in their expectations, or downright don't respect the presence of these env vars. The first boto being one of them.

In addition, we take no action to pass on these variables to the app pods that are spawned by your APB. That means you are responsible for writing your APB to check for proxy settings in their environment, and conditionally applying them to your service if it is required. There is interest in some kind of framework support for delivering these settings automatically, but that's not currently available.

docs/service-bundle.md
eriknelson: This is valuable to document, so not questioning its place in here, but do for my own edification, do we have a use-case for APB authors to use these files directly? The majority of the time, they just serve to authenticate the SA to do the APB work, no?

docs/service-bundle.md
rthallisey: ```bundle.spec```?

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+
+These are ignored.
+
+### Exit Code
+
+* 0 - success
+* 1 - error
+* 8 - action not implemented/supported
+
+### Encoded Credentials
```
rthallisey: I think we should also have a section for last_operation since it's something we pass back to the broker.

docs/service-bundle.md
djzager: There really isn't reason for a Service Bundle developer (APBs included as an instance of a Service Bundle) to interact with these directly. We may just want to link to the k8s documentation on what happens to service accounts in pods.

docs/service-bundle.md
djzager: I agree with making args more generic and having Service Bundle base images do the transform.

docs/service-bundle.md
djzager: I think the JSON is just the value for the extra vars parameter (aka. all the parameters being passed to the bundle)

docs/service-bundle.md
djzager: Wow. It is good to know that. That reminds me that it may also be a requirement of a Service Bundle developer to annotate the objects they create in the cluster a particular way, like the service instance they belong to.

docs/service-bundle.md
dymurray: Yeah perhaps worth removing `redhat` too eh? Label could just become `bundle.spec`.

docs/service-bundle.md
dymurray: If we are moving to generic definition of service bundle then perhaps its not even worth mentioning in the contract document. The contract should be agnostic of what the bundle is actually doing. In this case we would be assuming a service bundle is deploying on-cluster services.

Perhaps we also create a "Best Practices" document for provision on-cluster services? Unsure.

docs/service-bundle.md
djzager: Just for clarity, `last_operation` is an annotation on the running "Service Bundle" pod that can be read from the broker to get progress information right?

docs/service-bundle.md
eriknelson: The hope is that we might be able to use PodPresets to deliver them automatically via label, and then we could write a module that wraps that to make it easy for developers.

docs/service-bundle.md
djzager: I know that @mhrivnak and I had discussed this and the answer was yes. An APB is an instance of a Service Bundle adhering to the contract.

It wasn't clear to me, @rthallisey if you are suggesting the label be `com.redhat.bundle.spec` or just `bundle.spec`. Only ask since we already have `com.redhat.apb.runtime` (which would probably remain since it is APB specific).

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+behaviors described below.
+
+## Label
+
+Every service bundle container image has a label ‘com.redhat.apb.spec’ that
+contains a base64 encoded string. This label holds [important
+metadata](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#apb-spec-file)
+related to the service bundle. Example:
+
+```
```
djzager: Does the `apb.yml` become `bundle.yml`?

docs/service-bundle.md
rthallisey: Only ```bundle.spec```

docs/service-bundle.md
rthallisey: Yes.  The annotation on the pod is called ```apb_last_operation```.

mhrivnak:
As a reminder, let's please keep review within scope of making this reflect the current reality, and not let the discussion run away to a myriad of desired changes to how things work. I am also eager to have those discussions, but we'll benefit from first articulating a well-defined frame of reference.

I appreciate the feedback so far and will make a round of edits this morning. Thanks!

docs/service-bundle.md
shawn-hurley: Hey guys, I think that this is a follow-up issue:

> Please review how accurately this represents the current relationship between a Service Bundle and the Automation Broker as implemented today. Please do not (yet) review the technical merits of this design.

A discussion that we need to have I just don't think that this is the right place just yet IMO

docs/service-bundle.md
eriknelson: Gotcha, sorry was confused. That makes sense.

docs/service-bundle.md
jmrodri: That would be part of a proposal doc. So as far as I can tell, it should remain apb.yml until we decide.

docs/service-bundle.md
jmrodri: @rthallisey I don't think this document should read 'bundle.spec' because that would need to go through a proposal process. @mhrivnak is simply trying to document what it does today but use the terms "service bundle" in place of APB for the vernacular. Which leads to some confusing items as pointed out by @eriknelson, but it is where we are today. We will have to update this doc as we migrate.

jmrodri:
> This is valuable to document, so not questioning its place in here, but do for my own edification, do we have a use-case for APB authors to interact with these files directly? The majority of the time, they just serve to authenticate the SA to do the APB work, no?

@eriknelson I agree they should be documented even though the APB author might not interact with them directly, but it could help them understand why some things might not be working for them.

docs/service-bundle.md
mhrivnak: Here I was intentionally ansible-specific to call out an example of how this is done in practice. If someone wants to see this taking place, they can look for this ansible module being used in an APB.

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
+# Service Bundle Specification
+
+A Service Bundle is a container image that the broker can use to manage
+deployment of a service in a cluster. It adheres to the properties and
```
jmrodri: the deployment of a service

docs/service-bundle.md
jmrodri: defined by the OSB API

docs/service-bundle.md
```diff
@@ -0,0 +1,225 @@
...
+* _apb_plan_id: the name of a “plan”, as defined through the OSB API, that is
+  available on the service class.
+* _apb_service_class_id: the “service_id”, as defined through the OSB API, that
+  is being used for the current action.
+* _apb_service_instance_id: the “instance_id”, as defined through the OSB API,
+  that uniquely identifies the service instance being acted upon.
+* namespace: the target namespace in which resources should be created,
+  deleted, or acted upon.
+
+Other keys represent parameter names from the Plan, the values for which have
```
jmrodri: values of which have

docs/service-bundle.md
mhrivnak: It is possible to request a synchronous bind, and the broker may run a service bundle or not depending on its settings and what the bundle supports.

Is there a change you think should be made to this documentation right now? Or are you suggesting the behavior should change in the future and possibly be removed from the contract?

docs/service-bundle.md
eriknelson: ACK, makes sense. I'd say my comments re: the generic nature of this remain concerns of mine, but we're meeting to tackle that problem specifically, so for the purposes of this PR it's safe to ignore.

docs/service-bundle.md
eriknelson: @mhrivnak I think I mis-read this saying "bundles do not run for sync requests", which upon rereading again, is not what you meant.

docs/service-bundle.md
mhrivnak: I'm reading up on all of this, including this page: https://github.com/openshift/ansible-service-broker/blob/master/docs/administration.md

But I think I have too many questions in my mind about how things fit together to write this right now. Would one of you mind writing a couple of sentences about this?

docs/service-bundle.md
mhrivnak: Does the broker mount this information to this location? Or is it something kubernetes is doing automatically for some reason?

docs/service-bundle.md
```diff
@@ -0,0 +1,230 @@
...
+``NO_PROXY``, ``http_proxy``, ``https_proxy``, and ``no_proxy``.
+
+No action is taken to apply proxy settings to pods that a service bundle
+creates. Each service bundle is repsonsible for configuring its provisioned
+resources to use a proxy as appropriate.
+
+### Service Account
+
+Details of a kubernetes service account are mounted at
+``/var/run/secrets/kubernetes.io/serviceaccount``
```
djzager: https://kubernetes.io/docs/admin/service-accounts-admin/ I think we may want to avoid getting into the specifics here.

djzager: The `apb-base` has this generic [kube-config](https://github.com/ansibleplaybookbundle/apb-base/blob/master/files/opt/apb/.kube/config) that it allows it to simply use the service account that was given.

docs/service-bundle.md
```diff
@@ -0,0 +1,245 @@
...
+
+### stdout / stderr
+
+These are ignored.
+
+### Exit Code
+
+* 0 - success
+* 1 - error
+* 8 - action not implemented/supported
```
jmrodri: I've seen us return 2 as well, I'll find out when that was happening.

Issue #530, opened 2017-11-02
Pass the parameters sent during a bind back through to unbind


maleck13:
> # Feature:

When preforming a bind operation, we pass a set of bind parameters through. These are picked up by the apb pod and can be used during the binding. These parameters may contain information that is subsequently used to create resources in the target service.
For example, when binding to keycloak, I may want to create a bearer client for a specific service. During  the bind, I pass the name of the service through as a parameter and construct the name of the client based on the consuming service name.
When unbinding I want to be able to tear down what I created during the bind. In order to do this I need to know what service I am unbinding for. Having access to the params passed during the binding would make this possible.

It looks as though the params are stored with the binding:
https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/broker.go#L908

So in unbind, if we read back this binding based on the bindingID we can pass the binding params back through to the ubind action.

If this feature makes sense, let me know as I am happy to put together a PR for this

rthallisey:
@shawn-hurley did you work on something similar to this?

shawn-hurley:
I think this is something that we need to look into. I would like for the APB to decide what it needs during unbinding, and only send the data it needs during the unbind. I think this is an enhancement that we need to think about as we get async unbind & bind.  

@cfchase I think had some thoughts on this.

maleck13:
What is the rationale behind having to declare the params you are interested in (I assume you mean somewhere like the apb.yaml) seems it may potentially over complicate things? Is there a concern around information leaking?

maleck13:
hey guys any more thoughts on this, we would definitely find this useful. I am happy to put together a proposal on this.

shawn-hurley:
Information leaking is the primary concern that I was thinking of.

I thought that @cfchase had some thoughts on how this could/should work.

cfchase:
I agree I'd like to see this get implemented.  I think it's really necessary for a properly working async bind and was planning on using it in the unbind, myself: https://github.com/ansibleplaybookbundle/hello-world-db-apb/blob/async-bind/roles/unbind-hello-world-db-apb/tasks/main.yml#L24.

As far as passing the binding parameters to unbind, I didn't have any issues.  I couldn't think of a use case I was really worried about the information leakage, so I think someone else might better voice those concerns.

My concerns with service name weren't binding related as I remember and you can probably ignore the below if it's out of scope.

TLDR; We probably need some way to save variables for future use, not just bindings.

[...]

PR #809, opened 2018-03-02
APB state support

maleck13:
Proposal outlining a new feature for storing state as an APB developer

fixes https://github.com/openshift/ansible-service-broker/issues/530

@jmrodri ping initial proposal.

### DIFF
--- a/docs/proposals/apb-state-support.md
+++ b/docs/proposals/apb-state-support.md
@@ -1,7 +1,7 @@
 # Abstract
 
-This proposal will outline a simple and secure method for allowing APB developers to set state during a APB
-action and then be able to access that state during a subsequent APB action.
+This proposal will outline a simple and secure method for allowing Service Bundle developers to set state during a Service Bundle
+action and then be able to access that state during a subsequent Service Bundle action.
 
 ## Actions
    - provision
@@ -13,24 +13,25 @@ action and then be able to access that state during a subsequent APB action.
    
 ## Problem Description
 
-APBs are stateless. All state is managed for them by the broker. Currently the broker passes in parameters specified
+Service Bundle's are stateless. All state is managed for them by the broker. Currently the broker passes in parameters specified
 during a request to the catalog. It also passes in some additional parameters such as the namespace etc, additionally 
-we also pass in credentials created during the provision. While this is useful, there is no mechanism for an APB to store 
-and access data across actions without exposing it to the end user. We want to avoid APB developers working around this 
-limitation by doing suboptimal things such as storing extra data in the credentials or by creating a ConfigMap in the users 
-namespace as both of these expose the data to the user and as the namespace is controlled by the user, it is naturally 
+we also pass in credentials created during the provision. While this is useful, there is no mechanism for a Service Bundle to store 
+and access data across actions without exposing it to the end user. We want to avoid Service Bundle developers working around this 
+limitation by doing suboptimal things such as storing extra data in the credentials or by creating a ConfigMap in the user's 
+namespace as both of these expose the data to the user and, as the namespace is controlled by the user, it is naturally 
 not trustworthy or reliable.
 
 
 
 # Proposed Solution
 
-## APB module for handling state
+## Service Bundle contract for handling state
 
-This module will expose an API to the APB developer for setting and getting key value pairs.
+### Add APB specific module
+
+This module will expose an API to the APB developer for setting key value pairs.
 
 ```
-
 - name: Save some stuff
   asb_set_state:
     service_name: "{{ service_name }}"
@@ -38,23 +39,21 @@ This module will expose an API to the APB developer for setting and getting key 
 ```
 
 Under the hood this APB module would take the key value pair, and store it in a
-ConfigMap (or possibly a secret) labelled ```apb:state```. This ConfigMap would live within the 
-temporary namespace where the APB was running. 
+ConfigMap named ```$POD_NAME```. This ConfigMap would live within the 
+temporary namespace ```$POD_NAMESPACE``` and be created by the broker before the APB pod was created. 
 
-## Update broker to manage APB created state ConfigMaps
+## Update broker to manage Service Bundle created state ConfigMaps
 
-To ensure the state was persisted across APB action, the broker would watch for ConfigMaps being added in the 
-APB namespace with the expected label. When a new ConfigMap was created in the APB's namespace, the broker would 
-copy this ConfigMap over to the broker's namepsace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
-that name was already present, the broker would update and append the values. The broker would also watch 
-for ``Modified`` events on the ConfigMap and ensure to update the corresponding ConfigMap in the broker's namespace. 
+To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```. 
+After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. 
 There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
-once the ServiceInstance was deleted. 
+once the the deprovision action completed successfully. 
 
-## Update broker to pass through initial state to APB
+## Update broker to pass through initial state to Service Bundle
 
-For every APB action, except provision as there would be no state at this point, if a ConfigMap (ServiceInstanceID-state) is present, 
-in the broker's namespace, its key value pairs will be passed through to the APB prefixed with ```state_<key> = value ```
+For every Service Bundle action, except provision as there would be no state at this point, if a ConfigMap (ServiceInstanceID-state) is present, 
+in the broker's namespace, its key value pairs will be passed through to the Service Bundle prefixed with ```state_<key> = value ```
   
     
    
--- a/docs/proposals/apb-state-support.md
+++ b/docs/proposals/apb-state-support.md
@@ -1,0 +1,60 @@
+# Abstract
+
+This proposal will outline a simple and secure method for allowing APB developers to set state during a APB
+action and then be able to access that state during a subsequent APB action.
+
+## Actions
+   - provision
+   - deprovision
+   - update
+   - bind 
+   - unbind
+   
+   
+## Problem Description
+
+APBs are stateless. All state is managed for them by the broker. Currently the broker passes in parameters specified
+during a request to the catalog. It also passes in some additional parameters such as the namespace etc, additionally 
+we also pass in credentials created during the provision. While this is useful, there is no mechanism for an APB to store 
+and access data across actions without exposing it to the end user. We want to avoid APB developers working around this 
+limitation by doing suboptimal things such as storing extra data in the credentials or by creating a ConfigMap in the users 
+namespace as both of these expose the data to the user and as the namespace is controlled by the user, it is naturally 
+not trustworthy or reliable.
+
+
+
+# Proposed Solution
+
+## APB module for handling state
+
+This module will expose an API to the APB developer for setting and getting key value pairs.
+
+```
+
+- name: Save some stuff
+  asb_set_state:
+    service_name: "{{ service_name }}"
+
+```
+
+Under the hood this APB module would take the key value pair, and store it in a
+ConfigMap (or possibly a secret) labelled ```apb:state```. This ConfigMap would live within the 
+temporary namespace where the APB was running. 
+
+## Update broker to manage APB created state ConfigMaps
+
+To ensure the state was persisted across APB action, the broker would watch for ConfigMaps being added in the 
+APB namespace with the expected label. When a new ConfigMap was created in the APB's namespace, the broker would 
+copy this ConfigMap over to the broker's namepsace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. The broker would also watch 
+for ``Modified`` events on the ConfigMap and ensure to update the corresponding ConfigMap in the broker's namespace. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the ServiceInstance was deleted. 
+
+## Update broker to pass through initial state to APB
+
+For every APB action, except provision as there would be no state at this point, if a ConfigMap (ServiceInstanceID-state) is present, 
+in the broker's namespace, its key value pairs will be passed through to the APB prefixed with ```state_<key> = value ```
+  
+    
+   


docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+and access data across actions without exposing it to the end user. We want to avoid APB developers working around this 
+limitation by doing suboptimal things such as storing extra data in the credentials or by creating a ConfigMap in the users 
+namespace as both of these expose the data to the user and as the namespace is controlled by the user, it is naturally 
+not trustworthy or reliable.
+
+
+
+# Proposed Solution
+
+## APB module for handling state
```
shawn-hurley: Can we rename `bundle contract for handling state`? A minor nit right now because we don't have other modules. I just think this will need to be supported more broadly by the bundle/broker contract that @djzager and @mhrivnak have been working on.

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+- name: Save some stuff
+  asb_set_state:
+    service_name: "{{ service_name }}"
+
+```
+
+Under the hood this APB module would take the key value pair, and store it in a
+ConfigMap (or possibly a secret) labelled ```apb:state```. This ConfigMap would live within the
```
shawn-hurley: I wonder if a name for the config map/secret that conforms to a certain standard might be more appropriate then a label here?

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+copy this ConfigMap over to the broker's namepsace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. The broker would also watch 
+for ``Modified`` events on the ConfigMap and ensure to update the corresponding ConfigMap in the broker's namespace. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the ServiceInstance was deleted. 
+
+## Update broker to pass through initial state to APB
+
+For every APB action, except provision as there would be no state at this point, if a ConfigMap (ServiceInstanceID-state) is present, 
+in the broker's namespace, its key value pairs will be passed through to the APB prefixed with ```state_<key> = value ```
```
shawn-hurley: I wonder if this should actually be that the configmap/secret gets copied into the transient namespace and we add it to the pod. The contract then just has to define where the state gets mounted and it is up to the bundle to use that?

docs/proposals/apb-state-support.md
djzager: I like the idea of following the pattern we established for bind credentials. Create a configmap with name=`POD_NAME` in namespace=`POD_NAMESPACE`

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Update broker to manage APB created state ConfigMaps
+
+To ensure the state was persisted across APB action, the broker would watch for ConfigMaps being added in the 
+APB namespace with the expected label. When a new ConfigMap was created in the APB's namespace, the broker would 
+copy this ConfigMap over to the broker's namepsace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. The broker would also watch 
+for ``Modified`` events on the ConfigMap and ensure to update the corresponding ConfigMap in the broker's namespace. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the ServiceInstance was deleted.
```
djzager: I imagine what you could do, similar to how we handle bind credentials, would be to wait for the APB to be successfully executed (complete with `$? == 0`) and then move the ConfigMap under the Broker's control.

docs/proposals/apb-state-support.md
djzager: Originally, I thought that this Proposal was suggesting that the APB (or service bundle) has the option to create a configmap that the broker would save for subsequent runs for that service instance. However, if it were possible for the broker to create a configmap on behalf of the APB (or service bundle) mounted at `/bundle/state` that the bundle could use for saving state, that sounds even better to me (if it is possible).

djzager: @maleck13 I'm curious your thoughts on the above. If the broker were to provide Service Bundles a mount point `/bundle/state` as an example for saving state information, is that keeping with what you had in mind?

I'm actually curious if the broker could create a configmap in the `ansible-service-broker` namespace and provide that as a volume to an APB pod in the sandbox namespace.

maleck13: I don't think it is possible to mount a configmap from one namespace into a pod in another namespace . 
However it would be possible for the broker to create a configmap in the transient namespace during a provision and for the service bundle to be able to save state to it, then right before clean up the broker would copy this configmap to the broker's namespace. During any follow on action for that serviceinstance, the broker could then create a copy of the configmap in the transient namespace and add a volume to the pod. The service bundle could then add more state if needed and access any previous state from the mount point. 

How would you see the service bundle accessing the state? would this be via a ```apb_get_state <key>``` or just directly accessing the mount location?

maleck13: > mounted at /bundle/state that the bundle could use for saving state, that sounds even better to me (if it is possible).
Not sure this is possible. I don't think the contents of the configmap would be updated just by saving the state here.

docs/proposals/apb-state-support.md
maleck13: This sounds reasonable 👍

docs/proposals/apb-state-support.md
maleck13: No issue with this approach will update 👍

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+
+- name: Save some stuff
+  asb_set_state:
+    service_name: "{{ service_name }}"
+
+```
+
+Under the hood this APB module would take the key value pair, and store it in a
```
rthallisey: Can we decide on secret or configmap? I don't think secrets are meant for large amounts of key value pairs, but we probably want all the application data to be hidden.  IMO let's go with a secret until we have a need to change.

docs/proposals/apb-state-support.md
djzager: I played around with configmaps in pods a little and see that it is a read-only file system. I should have known better about mapping configmaps across namespaces...I've done worse :sunglasses:.

docs/proposals/apb-state-support.md
djzager: I'm not certain to what extent we want this data to be hidden. Given that the configmap would exist in the transient namespace and not the target namespace, it sounds sufficiently hidden. With this being configuration data, I vote for it being a configmap.

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
+# Abstract
+
+This proposal will outline a simple and secure method for allowing APB developers to set state during a APB
```
djzager: There is an ongoing discussion about the APB as an instance of a more generic Service Bundle (Bundle). This is the reason for @shawn-hurley earlier comment about "bundle contract".

My gut says that we should say Service Bundle or Bundle when referring to the generic and only use APB when we are specifically talking about Ansible Playbook Bundles.

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,60 @@
...
+during a request to the catalog. It also passes in some additional parameters such as the namespace etc, additionally 
+we also pass in credentials created during the provision. While this is useful, there is no mechanism for an APB to store 
+and access data across actions without exposing it to the end user. We want to avoid APB developers working around this 
+limitation by doing suboptimal things such as storing extra data in the credentials or by creating a ConfigMap in the users 
+namespace as both of these expose the data to the user and as the namespace is controlled by the user, it is naturally 
+not trustworthy or reliable.
+
+
+
+# Proposed Solution
```
djzager: This is where I think you should talk about the generic solution for Service Bundles or Bundles **and** a proposed module addition for use in APBs.

docs/proposals/apb-state-support.md
djzager: It makes sense to have this section detail the ansible module for handling state in the `ansible-asb-modules`.

docs/proposals/apb-state-support.md
maleck13: Yeah my instinct was also a configmap

maleck13:
@djzager  is there a link to more details on this contract? I am unaware of it currently

djzager:
@maleck13 https://github.com/openshift/ansible-service-broker/blob/master/docs/service-bundle.md

docs/proposals/apb-state-support.md
maleck13: I am a bit uncertain here. Should I just replace the heading or all instances of APB?

maleck13:
@djzager  @shawn-hurley  Updated to use the term ```ServiceBundle``` removed references to secrets as configmaps seem like the correct approach. Updated name the configmap $POD_NAME in $POD_NAMESPACE

Could you guys give it another look over when you get chance. Still a little unsure about the ServiceBudle wording but think I have made the changes suggested now

docs/proposals/apb-state-support.md
djzager: I think we may be back to @shawn-hurley 's original comment. Instead of passing through the key/value pairs could we:

1. Create the configmap `$POD_NAME` in `$POD_NAMESPACE` with the current set of key/value pairs
1. Mount the configmap, read-only :sunglasses:, to a known location on the image for reading those values

maleck13:
@djzager ok I will update with an ```apb_get_state <key>``` which will read the file in the container

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+## Service Bundle contract for handling state
+
+### Add APB specific module
+
+This module will expose an API to the APB developer for setting key value pairs.
+
+```
+- name: Save some stuff
+  asb_set_state:
+    service_name: "{{ service_name }}"
```
eriknelson: `service_name` in this case is the arbitrary state data, correct? Just want to make sure it's not suggested that this is some kind of special key.

docs/proposals/apb-state-support.md
eriknelson: I think I agree with @rthallisey, why not use secrets? Or maybe we have a `store_secret_state` and `store_state` variants?

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Under the hood this APB module would take the key value pair, and store it in a
+ConfigMap named ```$POD_NAME```. This ConfigMap would live within the 
+temporary namespace ```$POD_NAMESPACE``` and be created by the broker before the APB pod was created. 
+
+## Update broker to manage Service Bundle created state ConfigMaps
+
+To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```. 
+After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with
```
eriknelson: If we consider this in a large scale production deployment, are there scaling or quota concerns having potentially 10s of thousands of state maps around?

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+ConfigMap named ```$POD_NAME```. This ConfigMap would live within the 
+temporary namespace ```$POD_NAMESPACE``` and be created by the broker before the APB pod was created. 
+
+## Update broker to manage Service Bundle created state ConfigMaps
+
+To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```. 
+After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the the deprovision action completed successfully.
```
eriknelson: Is there any concern that a failed deprovision will leak a state configmap? Does there need  to be some kind of garbage collector to remove these? (I'm wondering if the operator pattern is valuable here for managing them).

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```. 
+After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the the deprovision action completed successfully. 
+
+## Update broker to pass through initial state to Service Bundle
+
+For every Service Bundle action, except provision as there would be no state at this point, if a ConfigMap (ServiceInstanceID-state) is present, 
+in the broker's namespace, its key value pairs will be passed through to the Service Bundle prefixed with ```state_<key> = value ```
```
eriknelson: We have precedence for passing `_apb` prefixed params for system params like this. Would it be acceptable to pass `_apb_state=<state_json_blob>`?

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+   - bind 
+   - unbind
+   
+   
+## Problem Description
+
+Service Bundle's are stateless. All state is managed for them by the broker. Currently the broker passes in parameters specified
+during a request to the catalog. It also passes in some additional parameters such as the namespace etc, additionally 
+we also pass in credentials created during the provision. While this is useful, there is no mechanism for a Service Bundle to store 
+and access data across actions without exposing it to the end user. We want to avoid Service Bundle developers working around this
```
eriknelson: > We want to avoid Service Bundle developers working around this

Big +1, if there's enough desire to have folks hacking around to make this happen, we need to provide first-class tools to enable them.

eriknelson:
One additional comment, this is going to require apb module development, I would call that out here and maybe submit something to github.com/ansibleplaybookbundle/ansible-playbook-bundle (not sure if that's where our module lives).

docs/proposals/apb-state-support.md
maleck13: just arbitrary

docs/proposals/apb-state-support.md
maleck13: I initially suggested that when the developer calls ```set_state``` it would create a configmap if was not already present. I think it would be better to have lazy creation over eager creation as the requirement for state is not going to be something every apb needs. So happy to go back to that approach. It should significantly reduce the number of objects created although there is still the potential for many objects over a long period of time. 
Is this enough to reduce the concern? I do not know enough about what the upper bound would be for configmaps in a namespace, it is also worth noting that we are not iterating through these objects but instead are accessing them by name. I am assuming that in etcd this is a O(1) action in cost rather than O(n)

docs/proposals/apb-state-support.md
maleck13: This is interesting as I know we should clean up all resources created and this would count as one of those. If a deprovision can be retried by the user then we should keep the state as it will be needed again during the next attempt. If it cannot be retried (which I think is the case as you cannot delete the service instance twice) then you are correct and we should always clean up the state configmap no matter what the exit code is. Good spot 👍

docs/proposals/apb-state-support.md
maleck13: Based on other comments, we were looking at adding a ```apb_get_state``` that would read the state from a mount point in the pod where the configmap would be mounted. WDYT?

docs/proposals/apb-state-support.md
shawn-hurley: I like the `store_secret_state`, this way`PostgreSQL` would never want to return w/ provision time credentials because they are admin credentials. They could store them in this state to make the subsequent bind calls functional. 

I do think that this creates a larger scope for this PR. If we have two variants of saving state, I could see a first phase and then a second phase IMO.

docs/proposals/apb-state-support.md
eriknelson: > I think it would be better to have lazy creation over eager creation as the requirement for state is not going to be something every apb needs.

That's a fair point that will probably dramatically reduce the number of configmaps lying around. :+1:

docs/proposals/apb-state-support.md
eriknelson: I like that a lot because it provides an abstraction layer on top of however it is that we choose to implement it. We'd be able to transparently change how we're actually implementing it without the APBs ever caring. I'm tempted to suggest that's a better pattern than using "_apb" parameters, but there's also the consideration that not all Service Bundles are APBs and have this luxury, and they aren't all going have the modules available. It's a problem we need to solve independently though, since we already have some unfriendly Service Bundle features that need to be made more independent.

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+   - provision
+   - deprovision
+   - update
+   - bind 
+   - unbind
+   
+   
+## Problem Description
+
+Service Bundle's are stateless. All state is managed for them by the broker. Currently the broker passes in parameters specified
```
jmrodri: typo: Bundle's -> Bundles

jmrodri: add a comma after "Currently"

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+during a request to the catalog. It also passes in some additional parameters such as the namespace etc, additionally
```
jmrodri: add comma after "additionally"

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+This module will expose an API to the APB developer for setting key value pairs.
+
+```
+- name: Save some stuff
+  asb_set_state:
+    service_name: "{{ service_name }}"
+
+```
+
+Under the hood this APB module would take the key value pair, and store it in a
```
jmrodri: add comma after hood

docs/proposals/apb-state-support.md
jmrodri: typo: duplicate "the". Remove the second one.

docs/proposals/apb-state-support.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Update broker to manage Service Bundle created state ConfigMaps
+
+To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```. 
+After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with 
+that name was already present, the broker would update and append the values. 
+There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace 
+once the the deprovision action completed successfully. 
+
+## Update broker to pass through initial state to Service Bundle
```
jmrodri: add a "the" in front of initial state.

maleck13:
Going to make a small update to ensure the proposal reflects the decisions. 
* [ ]  Add reference to mounting the configmap in the ServiceBundle Pods rather than just passing the params through 
* [ ] Add reference to ```apb_get_state:
                                          key: "somekey"```

* [ ] Remove reference to passing through the state via the extra vars
* [ ] Call out always removing the configmap on deprovision

PR #811, opened 2018-03-02
Allow playbooks names to be something other than $action

rthallisey:
Instead of forcing the user to have a playbook named provision, deprovision, bind, or unbind, allow the user to create a single playbook with the  ```Name``` field in apb.yml.  A single playbook would work because we also pass in ```action``` to the playbook, which will then call the ```$action``` role or tasks in the ```$name``` playbook.

Here's an example of the action being used at the task level: https://github.com/ansibleplaybookbundle/kubevirt-apb/blob/master/roles/kubevirt-apb/tasks/main.yml

 - Add BUNDLE_NAME to the apb environment

depends-on https://github.com/ansibleplaybookbundle/apb-base/pull/20

### DIFF
--- a/pkg/apb/executor.go
+++ b/pkg/apb/executor.go
@@ -229,7 +229,7 @@ func (e *executor) executeApb(
 						"--extra-vars",
 						extraVars,
 					},
-					Env:             createPodEnv(executionContext),
+					Env:             createPodEnv(executionContext, spec.Name),
 					ImagePullPolicy: pullPolicy,
 					VolumeMounts:    volumeMounts,
 				},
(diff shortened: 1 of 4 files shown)


djzager:
As I have been playing around with the apb reference example, I have thought about this some. In my opinion though, this seems more appropriate as a proposal first.

This is especially important with the service bundle discussion that we are having next week.

rthallisey:
Were you thinking a proposal on the different options here?  The options I thought of were 1) adding a new apb.yml config option 2) using ```Name``` (which I don't think we use for anything) 3) Exposing a new parameter

What else did you have in mind?  I also didn't know there was a service bundle discussion happening next week.

jmrodri:
I can see this making the apb more flexible, but I fear it could make it more difficult. I mean today it's pretty easy. What options do I have? there are 5 actions: provision, deprovision, bind, unbind, update. I'd like to hear more about what we're trying to do with this.

rthallisey:
Closing for now.  We can reopen after proposal is discussed.

rthallisey:
https://github.com/openshift/ansible-service-broker/pull/815

PR #815, opened 2018-03-05
Custom Playbook Name Proposal

rthallisey:


docs/proposals/custom-playbook-names.md
```diff
@@ -0,0 +1,58 @@
+# Custom Playbook Names
+
+## Introduction
+Instead of forcing the user to have a playbook named provision, deprovision,
+bind, or unbind, allow the user to create a single playbook and select the
```
djzager: I know that this was the direction that I was headed with the APB reference example. The only difference between the different playbooks was the `action` variable (provision, deprovision, ...). If you have a look at https://github.com/ansibleplaybookbundle/hello-world-apb/pull/3 you will see that the APB is structured as a single ansible role installed as `apb` in the resulting APB image. Not sure how much that factors into this discussion, but I think it is something you could consider as you develop this proposal.

docs/proposals/custom-playbook-names.md
```diff
@@ -0,0 +1,58 @@
...
+broker action at the role or task level.
+
+
+## Problem Description
+If I want to convert my ansible playbooks to an APB, it's required that I create
+provision, bind, unbind, or deprovision playbook(s). This is inconvenient
+because I would need to change my playbook name, documentation, possibly some
+variables, and break any existing users just so I can support the APB format.
+
+If I want to have multiple APBs in a single git repo I get a similar issue.
```
djzager: It would help me to see a more concrete example, maybe even with the relevant portion of the file tree, when discussing this. It is not immediately obvious to my why we need to modify the apb-base to handle this. Some changes to your APBs dockerfile could accomplish this for you.

docs/proposals/custom-playbook-names.md
rthallisey: I want to use a single playbook with a variable to reflect the action.  Which is the same structure as your PR.

docs/proposals/custom-playbook-names.md
```diff
@@ -0,0 +1,88 @@
...
+multiple_apb_repo
+├── apb_number_1
+│   └── provision.yml   # Provision Playbook
+└── apb_number_2
+    └── provision.yml
+```
+
+The APB structure is rigid and if I want to turn my existing playbook into
+an APB I will have to make some changes to the structure of my playbooks.
+This will make is difficult for folks who already have playbooks to buy into
```
jmrodri: TYPO: make is -> make it

djzager:
I do believe that the easiest way for you to be able to move forward w/o having to modify the apb-base would be to simply update the Dockerfile of the impacted APBs.

Instead of [this example](https://github.com/ansibleplaybookbundle/hello-world-apb/blob/master/Dockerfile):

```
FROM ansibleplaybookbundle/apb-base
LABEL "com.redhat.apb.spec"="..."
ADD playbooks /opt/apb/actions
ADD roles /opt/ansible/roles
RUN chmod -R g=u /opt/{ansible,apb}
USER apb
```

Your APB's dockerfile could look something like:

```
FROM ansibleplaybookbundle/apb-base
LABEL "com.redhat.apb.spec"="..."

COPY playbooks/apb_number1.yml /opt/apb/actions/provision.yml
COPY playbooks/apb_number1.yml /opt/apb/actions/deprovision.yml

ADD roles /opt/ansible/roles
RUN chmod -R g=u /opt/{ansible,apb}

USER apb
```

I'm not clear what these projects look like structure wise but I suspect you could put the files in the correct place at build time w/o breaking the pattern.

rthallisey:
That would work @djzager, thanks

docs/proposals/custom-playbook-names.md
tsanders-rh: This could be handled with a shim playbook that calls the existing role without modification, true?  Also, help me understand more about why one would would want multiple APBs in the same repo?

docs/proposals/custom-playbook-names.md
tsanders-rh: +1 A concrete example would be helpful.

rthallisey: Here's an example file tree: https://github.com/openshift/ansible-service-broker/pull/815/commits/d219b713b64b8c59fe293d892b70733f1ff7f16b

eriknelson:
Can we summarize the results of the F2F discussion here? It sounded like maybe this should get either updated, or closed, and we should work with @dymurray to implement some best practices based on where he sees this going.

rthallisey:
A best practices doc would cover this

PR #816, opened 2018-03-05
parameterize dao:type. Defaults to etcd

jmrodri:


jmrodri:
Related to https://github.com/fusor/catasb/pull/216

PR #827, opened 2018-03-08
Default to an open localregistry policy

### LINKED ISSUE #823: Change default development template whitelist to `*`
For development we don't want to enforce a whitelist in the local registry adapter because new APB developers do not realize that they have to name their APB `$name-apb`. Let's change the default behavior to allow all naming conventions unless the user explicitly changes it since this template will be used for development.

eriknelson:
Having a default policy of only allowing APBs with a name suffixed by "-apb" has been confusing our development community. This change removes the above restriction by default. The justification for this change is that the deployment environments that use this template from the github repo include catasb, minishift, and `run_latest_build.sh`. These are all specifically development environments. `openshift-ansible`, which is likely used for production deployments, carries its own production deployment template. 

fixes #823

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -317,8 +317,11 @@ objects:
         - type: local_openshift
           name: localregistry
           namespaces: ['openshift']
+          # NOTE TO ADMINS: The default whitelist policy here will pass *all* APBs
+          # found in the local openshift registry. If this is not desired,
+          # manipulate the following regex to only match APBs you wish to be made available.
           white_list:
-            - ".*-apb$"
+            - ".*"
       dao:
         type: "${DAO_TYPE}"
         etcd_host: asb-etcd.${NAMESPACE}.svc


templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -317,8 +317,11 @@ objects:
         - type: local_openshift
           name: localregistry
           namespaces: ['openshift']
+          # NOTE TO ADMINS: The default whitelist policy here will pass *all* APBs
+          # found in the local openshift registry. If this is not desired,
+          # manipulate the following regex to only match APBs you wish to be made available.
           white_list:
-            - ".*-apb$"
```
djzager: Does it work if you just comment it out? Maybe just add it to the block comment.

eriknelson: It would if you removed the `white_list` altogether, or just commented it out, but I sort of feel like it's not a bad thing to ask admins who would like to tweak this to write their own regex. Also there is an example immediately above for dockerhub.

djzager: I agree with you.

PR #828, opened 2018-03-09
[wip] initial changes to subscriber pattern

maleck13:
Updates to how we handle the channels and subscribers based on the proposal outlined as part of 
https://github.com/openshift/ansible-service-broker/issues/638

@jmrodri @eriknelson  initial WIP work here for early feedback. Not complete yet but also not miles away. Would like any feedback you have to ensure we are still on the same page and I am not heading down any rabbit holes 😄


pkg/broker/broker.go
```diff
@@ -1059,13 +1059,6 @@ func (a AnsibleBroker) Unbind(
 	} else {
 		log.Warning("Broker configured to *NOT* launch and run APB unbind")
 	}
-
-	err = cleanupUnbind(&bindInstance, &serviceInstance, bindExtCreds, a.dao)
```
jmrodri: Why is this gone? It was part of the unbind_subscriber before, what is doing this now?

jmrodri: I see it was moved to `cleanupAfterUnbind` in the `job_state_subscriber`.

pkg/broker/types.go
```diff
@@ -304,4 +304,12 @@ type SubscriberDAO interface {
 	SetState(id string, state apb.JobState) (string, error)
 	GetServiceInstance(id string) (*apb.ServiceInstance, error)
 	DeleteServiceInstance(id string) error
+	GetBindInstance(id string) (*apb.BindInstance, error)
+	DeleteBindInstance(id string) error
+	SetServiceInstance(id string, serviceInstance *apb.ServiceInstance) error
```
jmrodri: I'm not a fan of the SubscriberDAO. It feels weird to me. But that's for a different day.

pkg/broker/types.go
```diff
@@ -304,4 +304,12 @@ type SubscriberDAO interface {
...
+}
+
+// WorkSubscriber - Interface tha wraps the Subscribe method
```
jmrodri: typo: tha -> that

jmrodri: also it says that wraps the `Subscribe` method, is that correct?

pkg/broker/work_engine.go
```diff
@@ -29,13 +29,13 @@ type Work interface {
...
-	bufsz  int
+	subscribers map[WorkTopic][]WorkSubscriber
+	jobs        map[string]chan JobMsg
 }
 
 // NewWorkEngine - creates a new work engine
-func NewWorkEngine(bufferSize int) *WorkEngine {
-	return &WorkEngine{topics: make(map[WorkTopic]chan JobMsg), bufsz: bufferSize}
+func NewWorkEngine() *WorkEngine {
+	return &WorkEngine{jobs: make(map[string]chan JobMsg), subscribers: map[WorkTopic][]WorkSubscriber{}}
```
jmrodri: again, why is the channel a non buffered channel?

pkg/broker/work_engine.go
```diff
@@ -50,10 +50,33 @@ func (engine *WorkEngine) StartNewAsyncJob(
 	if token == "" {
 		token = engine.Token()
 	}
-	go work.Run(token, engine.topic(topic))
+	go engine.start(token, work, topic)
```
jmrodri: This feels really weird to me. Starting the engine  for each job. Feels like the engine should always be running and waiting for new jobs to be added to it to be run.

pkg/mock/dao.go
```diff
@@ -86,6 +86,49 @@ func (mp *SubscriberDAO) GetServiceInstance(id string) (*apb.ServiceInstance, er
 	return retOb.(*apb.ServiceInstance), mp.Errs["GetServiceInstance"]
 }
 
+//DeleteBindInstance mock impl
```
jmrodri: spaced after the `//`

pkg/broker/types.go
shawn-hurley: As this is adding more and more methods, why don't we just make DAO now that DAO is an interface itself?

We either have or easily can create a mock for it as well.

pkg/broker/work_engine.go
```diff
@@ -50,10 +50,33 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+		close(engine.jobs[token])
+		delete(engine.jobs, token)
+	}()
+
+	go func() {
+		// listen for new messages and hand them off to the subscribers
+		for msg := range engine.jobs[token] {
+			for _, sub := range engine.subscribers[topic] {
+				//TODO edge case consider the fact that a subscriber may never exit and so we would leak go routines
+				go sub.Notify(msg)
```
shawn-hurley: In regards to the comment, I do think that having a context timeout makes sense here. 

I think a simple implementation would be to wrap this in a func() that waited for a configurable amount of time. 

The notify interface could then have an error return that we would log if something does go wrong. and if it returns with nil then everythng is all good. 

```go
...
for _, sub := range engine.subscribers[topic] {
         c := make(chan error, 1)
        go waitForNotify(engine.duration, sub, msg)
....

func waitForNotify(d time.Duration,  w WorkSubscriber, j JobMessage) {
              select {
            case err := w.Notify(j):
                 handle error or if nil return
            case <-d:
            }
}

```

pkg/broker/work_engine.go
shawn-hurley: This could be called startSubscriberListeningAndRun(). I would prefer to keep in smaller maybe something like `startJob`?

shawn-hurley:
I think this is looking really good, does a good job of reducing complexity!

pkg/app/app.go
```diff
@@ -62,8 +62,6 @@ var (
 )
 
 const (
-	// MsgBufferSize - The buffer for the message channel.
-	MsgBufferSize = 20
```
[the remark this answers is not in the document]
maleck13: Yes also we are reading from that channel and handing of to the subscribers async, so we are always ready to read the next message. If we wanted to slow throughput we could very easily use a ```sync.WaitGroup``` in combination with a buffered channel.

pkg/broker/types.go
maleck13: Yep no issue with changing that as we already have the DAO interface

pkg/broker/work_engine.go
maleck13: As currently we are handing off the messages immediately and asynchronously we should always be able to read from and send to the channel so in this case there is no benefit to a buffered channel.  

Where we may benefit from a buffered channel is if we used it in conjunction with a ```sync.WaitGroup``` 

Something like:

```
  wg := sync.WaitGroup{}

  for msg := range engine.jobs[token] {
      wg.Add(len(engine.subscribers))
      for _, sub := range engine.subscribers[topic] {
          go sub.Notify(msg, wg) // Subscriber will call Done()
      }
      wg.Wait() // wait for all subscribers to be done
  }
```

This design would allow each of the subscribers to receive the messages at the same time, but force us to wait until they are all complete before accepting the next message. In this case having a buffer would be valuable as it would allow the Job to continue to send messages even if our subscribers were running a little slow. 
The question really is if there is value in waiting for all the subscribers to finish before accepting the next message. 
I am leaning towards yes on this due to the fact that state will be changed and modified and there is no guarantee on that the order a set of go routines were created will also be the order they are executed in.

pkg/broker/work_engine.go
maleck13: Going to chew this one over 🤔

shawn-hurley: I think your right, I think you need to pass a channel to `waitForNotify` function and add the timeout bits to a function that gets launched with the go routine above.

Something like this?
https://play.golang.org/p/pRjyCXfZgSi

maleck13: updated to take into account your feedback plus look at adding in the wait group and using context https://gist.github.com/maleck13/d2c903eac6064422966d64715118345e

maleck13:
Ok so I have updated based on feedback and discussions. It is still WIP, but again wanted to give opportunity for feedback. 
After much discussion (thanks guys) the design changed to make the start job more complex than the original commit, but I think it provides a more robust solution. 

**Main changes**
We have introduced a WaitGroup for each message that comes in, along with a context object with timeout. 
We now have the following on a per job basis:
```
msg-> buffered-channel -> fanout-to-subscribers-async -> lock-until-all-subs-done-> read-next-msg
```
The key concern for me here was to avoid a situation where writes could be executed in the wrong order.
There is no guarantee on the order that go routines will execute. So without the synchronization, there would be the possibility that a write could happen in the wrong order. Now however I do not believe this is the case as we use the WaitGroup to ensure subscribers are finished and the order messages come off a channel is FIFO  

ping @shawn-hurley  @jmrodri  @eriknelson 

Will be adding a number of unit tests here and also looking at a metrics subscriber in the next day or so.

pkg/app/app.go
```diff
@@ -62,10 +62,10 @@ var (
...
 	// defaultClusterURLPreFix - prefix for the ansible service broker.
 	defaultClusterURLPreFix = "/ansible-service-broker"
+	// MsgBufferSize - The buffer for the message channel.
+	MsgBufferSize = 20
```
jmrodri: +1 this makes me feel better :)

pkg/app/app.go
```diff
@@ -213,43 +213,43 @@ func CreateApp() App {
 	validateRegistryNames(app.registry)
 
 	log.Debug("Initializing WorkEngine")
+	stateSubscriber := broker.NewJobStateSubscriber(app.dao)
```
jmrodri: :+1: to the single stateSubscriber.

pkg/broker/job_state_subscriber.go
```diff
@@ -0,0 +1,106 @@
...
+func (jss *JobStateSubscriber) ID() string {
+	return "jobstate"
+}
+
+// Notify external API to notify this subscriber of a change in the Job
+func (jss *JobStateSubscriber) Notify(msg JobMsg) {
+	log.Debugf("JobStateSubscriber Notify : msg state %v ", msg.State)
+	id := msg.InstanceUUID
+	if isBinding(msg) {
+		id = msg.BindingUUID
```
jmrodri: Can `msg.BindingUUID` ever be `""`? I would probably change `isBinding` to also ensure `BindingUUID` is not empty. Otherwise, we would store an empty string as the key. Maybe there is no way this could happen, I'll see how the flow goes.

pkg/broker/work_engine.go
```diff
@@ -29,13 +32,14 @@ type Work interface {
 
 // WorkEngine - a new engine for doing work.
 type WorkEngine struct {
-	topics map[WorkTopic]chan JobMsg
-	bufsz  int
+	subscribers   map[WorkTopic][]WorkSubscriber
+	jobs          map[string]chan JobMsg
```
jmrodri: jobs is weird because these aren't the jobs. They're the channels used by the jobs for messages. The name implies the job structs are in this map but that is not the case. Calling `jobs["token"]` will not return me the job I created. Consider renaming it.

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
 	if token == "" {
 		token = engine.Token()
 	}
-	go work.Run(token, engine.topic(topic))
+	go engine.startJob(token, work, topic)
```
jmrodri: `startJob` is better and more clear.

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
 
+func waitForNotify(sub WorkSubscriber, msg JobMsg, signal chan<- struct{}) {
+	sub.Notify(msg)
+	signal <- struct{}{}
+}
+
+func (engine *WorkEngine) startJob(token string, work Work, topic WorkTopic) {
+	// create a channel specifically for use with this job
+	jobChannel := make(chan JobMsg, engine.jobBufferSize)
+	engine.jobs[token] = jobChannel
```
jmrodri: :+1: to `jobChannel`, that is much more descriptive.

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+					select {
+					case <-notifySignal:
+						return
+					case <-ctx.Done():
+						log.Errorf("Subscriber %s timeout %v ", sub.ID(), ctx.Err())
+						return
+					}
+				}(msg, sub)
+			}
+			//ensure we wait until all subs are done before taking on the next message
```
jmrodri: add a space after the `//`

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+	go func() {
+		// listen for a new message for the job keyed to this token and hand off to the subscribers async. Wait for them all to be done before accepting
+		// the next message
+		for msg := range jobChannel {
+			wg := &sync.WaitGroup{}
+			// hand off the msg to all subscribers async
+			for _, sub := range engine.subscribers[topic] {
+				go func(msg JobMsg, sub WorkSubscriber) {
+					wg.Add(1)
+					// ensure things don't get locked up. Each subscriber has up tp the configured amount of time to complete its action
```
jmrodri: can we make this comment a 2 line comment. It would be easier to read.

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+	// ensure we always clean up
+	defer func() {
+		log.Debug("closing channel for job ", token, engine.jobs)
+		close(jobChannel)
+		delete(engine.jobs, token)
+	}()
+
+	go func() {
+		// listen for a new message for the job keyed to this token and hand off to the subscribers async. Wait for them all to be done before accepting
+		// the next message
```
jmrodri: reformat this comment to be a couple shorter lines. The first one is really long.

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+					notifySignal := make(chan struct{})
+					//If our subscriber times out or returns normally we will always clean up
+					defer func() {
+						wg.Done()
+						close(notifySignal)
+						cancel()
+					}()
+					// notify the subscriber
+					go waitForNotify(sub, msg, notifySignal)
+					//act on whichever happens first the subscriber's notify method completing or the timeout
```
jmrodri: space after `//`

pkg/broker/work_engine.go
```diff
@@ -50,10 +54,65 @@ func (engine *WorkEngine) StartNewAsyncJob(
...
+			wg := &sync.WaitGroup{}
+			// hand off the msg to all subscribers async
+			for _, sub := range engine.subscribers[topic] {
+				go func(msg JobMsg, sub WorkSubscriber) {
+					wg.Add(1)
+					// ensure things don't get locked up. Each subscriber has up tp the configured amount of time to complete its action
+					ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second) //TODO make configurable
+					// used to tell us when the subscribers notify method is completed
+					notifySignal := make(chan struct{})
+					//If our subscriber times out or returns normally we will always clean up
```
jmrodri: space after `//`

pkg/broker/work_engine.go
```diff
@@ -66,7 +125,7 @@ func (engine *WorkEngine) StartNewSyncJob(
 		token = engine.Token()
 	}
 
-	work.Run(token, engine.topic(topic))
+	engine.startJob(token, work, topic)
```
jmrodri: +1

pkg/broker/work_engine.go
```diff
@@ -93,17 +143,17 @@ func (engine *WorkEngine) AttachSubscriber(
...
-	subscriber.Subscribe(msgBuffer)
 	return nil
 }
 
-// GetActiveTopics - Get list of topics
-func (engine *WorkEngine) GetActiveTopics() map[WorkTopic]chan JobMsg {
-	return engine.topics
+// GetActiveJobs - Get list of active jobs
+func (engine *WorkEngine) GetActiveJobs() map[string]chan JobMsg {
+	return engine.jobs
```
jmrodri: is this really active jobs? I would expect it to return an array of jobs.

pkg/broker/work_engine.go
maleck13: changed to ``` jobChannels```

pkg/broker/work_engine.go
maleck13: changed to GetActiveJobChannels

pkg/broker/job_state_subscriber.go
maleck13: It seems if it can ever be empty that it should fail before getting here, similar with instanceUUID.

maleck13:
@jmrodri I added back the clean up of unbind state when launch_apb_on_bind is false. I think this will fix the CI failure. 
Wondering what you think to moving the unbind clean up into the dao layer as it is essentially a fascade around a set of dao calls?

maleck13:
added to DAO can remove if need be

maleck13:
@shawn-hurley @jmrodri  @eriknelson this is no longer WIP please review when ready.

pkg/dao/dao.go
```diff
@@ -79,6 +79,9 @@ type Dao interface {
 	// DeleteBindInstance - Delete the binding instance for an id in the kvp API.
 	DeleteBindInstance(string) error
 
+	// DeleteBinding - Delete the binding instance and remove the assocation with the service instance.
+	DeleteBinding(apb.BindInstance, apb.ServiceInstance) error
```
shawn-hurley: 👍

shawn-hurley:
@maleck13 If you could rebase this, the build should start passing

pkg/app/app.go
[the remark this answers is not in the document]
eriknelson: I don't see any reason to artificially slow the processing of msgs.

pkg/broker/work_engine.go
eriknelson: :+1: I'm not sure it should be a private fn though? I suppose it's fine since we're really the only ones using it but I thought it was useful to design it from a perspective of "this is a public lib". No requested changes, but spitballing.

PR #829, opened 2018-03-09
Add a note to run latest about current user

eriknelson:
* Run latest logs in as system:user, and it's surprising our developers
when they go to use the apb tool, as a non-token'd user will not work.
Adds a note calling this out to run_latest_build.sh

fixes https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/issues/242


shawn-hurley:
Do we give the developer account the correct permissions here for them to use the apb tool? 

If we don't I think it might just be better to grant the permissions and log in as the developer account?

PR #830, opened 2018-03-09
Proposal for Helm Chart Registry Adapter

djzager:
This describes a plan for supporting Helm Chart Repositories.

### DIFF
--- a/docs/proposals/helm-charts.md
+++ b/docs/proposals/helm-charts.md
@@ -2,10 +2,10 @@
 
 ## Introduction
 
-Add a new registry adapter type, `helm`, that supports Helm chart
-repositories. With this change, an admin could configure the broker by adding
-a `helm` registry type and have the Helm charts hosted in that Helm chart
-repository converted into objects the broker knows how to manage for use via
+Add a new registry adapter type, `helm`, that supports Helm Chart
+Repositories. With this change, an admin could configure the broker by adding
+a `helm` registry type and have the Helm charts hosted in that Helm Chart
+Repository converted into objects the broker knows how to manage for use via
 the Service Catalog.
 
 ## Problem Description
@@ -19,12 +19,12 @@ a cluster.
 
 ## Creating a Base Image for Executing Helm Charts
 
-Ansible Playbook Bundles (APBs) have already been described as an instance of
-the more generic Service Bundle. Here, we will be creating a new type of
-Service Bundle, Helm Bundles, that respect the Service Bundle contract but use
-Helm as the runtime. We already have a [Helm Bundle
-Base](https://github.com/ansibleplaybookbundle/helm-bundle-base) and this
-[PR](https://github.com/ansibleplaybookbundle/helm-bundle-base/pull/2)
+Ansible Playbook Bundles (APBs) have served as an initial implementation of the
+more generic [Service Bundle Contract](../service-bundle.md). Here, we will be
+creating a new type of Service Bundle, Helm Bundles, that respect the Service
+Bundle contract but use Helm as the runtime. We already have a
+[Helm Bundle Base](https://github.com/ansibleplaybookbundle/helm-bundle-base)
+and this [PR](https://github.com/ansibleplaybookbundle/helm-bundle-base/pull/2)
 shows the work required to implement the Service Bundle contract generically
 using Helm.
 
@@ -40,23 +40,23 @@ registry:
   - type: helm
     name: stable
     url: "https://kubernetes-charts.storage.googleapis.com"
-    base_image: "docker.io/djzager/helm-bundle-base:latest"
+    runner: "docker.io/djzager/helm-runner:latest"
     white_list:
       - ".*"
 ```
 
-1. Type: refers to the registry adapter to handle this registry
+1. Type: refers to the registry adapter to handle Helm Chart Repositories
 1. Name: gives a name to this registry item
-1. URL: the URL for the Helm chart repository
-1. Base Image: the container image to use when provisioning/deprovisioning a
-   Helm chart
-1. White List/Black List: allows the Admin to filter out Helm charts
+1. URL: the URL for the Helm Chart Repository
+1. Runner: the container image to use when interacting with a given
+   Helm Chart
+1. White List/Black List: allows the Admin to filter out Helm Charts
 
 ### Changes to the Registry Package
 
-- Add the ability to read the `base_image` from the broker config.
+- Add the ability to read the `runner` from the broker config.
 - Pass the URL and BaseImage to the registry adapter.
-- Instantiate a Helm registry adapter to handle Helm Chart registries in the
+- Instantiate a Helm registry adapter to handle Helm Chart repositories in the
   config.
 
 ### Creating a Helm Adapter
@@ -76,16 +76,41 @@ type HelmAdapter struct {
 The addition of the `Charts` field makes it possible to save our work and only
 read the registries' index file once.
 
-When the broker calls `GetImageNames()` the Helm adapter will read the
-`index.yaml` found at the Helm Chart Repository URL based on the [chart
-repository
-structure](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#the-chart-repository-structure).
-The broker will create an entry in the `Charts` field to hold a list of
-`ChartVersion` objects and  add the chart name to the list of `imageNames`
-to be returned.
+When the registry package calls `GetImageNames()` the Helm adapter will read the
+`index.yaml` found at the Helm Chart Repository URL based on the
+[chart repository structure](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#the-chart-repository-structure).
+Inside `GetImageNames()`, the `Charts` fields will be initialized and we will
+save the `ChartVersion`s into the `Charts` field based on the Chart's name. For
+example:
 
-After these image names are filtered, the subsequent call to `FetchSpecs()`
-will:
+```
+// GetImageNames - retrieve the images
+func (r *HelmAdapter) GetImageNames() ([]string, error) {
+       var imageNames []string
+
+       r.Charts = map[string][]*repo.ChartVersion{}
+
+       index, err := r.getHelmIndex()
+       if err != nil {
+               return imageNames, err
+       }
+
+       for name, entry := range index.Entries {
+               if len(entry) == 0 {
+                       continue
+               }
+
+               r.Charts[name] = entry
+               imageNames = append(imageNames, name)
+       }
+
+       return imageNames, nil
+}
+```
+
+This way we only have to evaluate the Chart Repository on the call to
+`GetImageNames()` and the subsequent call to `FetchSpecs()` can work
+to transform Charts:
 
 - Find the chart by name in the `Charts` field
 - Save all the versions of the chart:
@@ -170,7 +195,7 @@ spec := &apb.Spec{
                     Name:      "version",
                     Title:     "Helm Chart Version",
                     Type:      "enum",
-		    Enum:      chartVersions,
+                    Enum:      chartVersions,
                     Default:   chart.Version,
                     Updatable: true,
                     Required:  false,
--- a/docs/proposals/helm-charts.md
+++ b/docs/proposals/helm-charts.md
@@ -1,10 +1,10 @@
-# Helm Chart Registry Adapter Proposal
+# Basic Helm Chart Registry Adapter Proposal
 
 ## Introduction
 
 Add a new registry adapter type, `helm`, that supports Helm chart
-repositories. With this change, an admin, could configure the broker by adding
-a `helm` type registry and have the Helm charts hosted in that Helm chart
+repositories. With this change, an admin could configure the broker by adding
+a `helm` registry type and have the Helm charts hosted in that Helm chart
 repository converted into objects the broker knows how to manage for use via
 the Service Catalog.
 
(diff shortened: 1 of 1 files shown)


docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+```
+// import "k8s.io/helm/pkg/repo"
+// HelmAdapter - Helm Registry Adapter
+type HelmAdapter struct {
+    Config Configuration
+    Log    *logging.Logger
+    Charts map[string]*repo.ChartVersion
+}
+```
```
jmrodri: Let's reword this section. I had to read it twice to understand what was going on. 

The `HelmAdapter` struct will look something like the example below:

```golang
// import "k8s.io/helm/pkg/repo"
// HelmAdapter - Helm Registry Adapter
type HelmAdapter struct {
    Config Configuration
    Log    *logging.Logger
    Charts map[string]*repo.ChartVersion
}
```
The addition of the `Charts` field make it possible to save our work and only read the registries' index file once.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+When the broker calls `GetImageNames()` the Helm adapter will read the
+`index.yaml` found at the Helm Chart Repository URL based on the [chart
+repository
+structure](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#the-chart-repository-structure)
+and add the latest version of each chart to the `Charts` field as well as
+its name to the list of `imageNames` to be returned.
```
jmrodri: break up the sentence:

>...repository structure](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#the-chart-repository-structure). The broker will add the latest version of each chart to the...

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+After these image names are filtered, the subsequent call to `FetchSpecs()`
+will find the chart by name in the `Charts` field and create a `Spec` object
+for it. For example:
```
jmrodri: So will a call to `bootstrap` ever regenerate this `Charts` cache? How will we pick up changes?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+        },
+    },
+}
+```
+
+Most of the data required to create a useful `Spec` object is contained in
+the Chart object. However, in order to retrieve the default values we will use
+[Helm's Chartutil
+pkg](https://github.com/kubernetes/helm/blob/master/pkg/chartutil/load.go) to
+load the archive specified at `Chart.URLs[0]` and read the values.
```
jmrodri: Can you elaborate on this process a bit more? What sort of archive is it? Are we downloading something and cracking it open? Where is it being stored? etc.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Phase 2 - Support Authenticated Chart Repositories
+
+Helm Chart Repositories can be authenticated, [this
+issue](https://github.com/kubernetes/helm/issues/1038) provides more
+information. The broker should support authenticated chart repositories.
```
jmrodri: Should we support all of the authentication formats? How will we accomplish this? Or would we do another proposal for handling phase 2 at a later date?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Phase 3 - Chart Versions as Parameter
+
+In phase 1, only the latest chart version will be used. The broker should
+support all versions of a chart and allow the consumer to specify which version
+of a chart they wish to install. The base image would need to be updated to
+handle specifying a specific version at provision time.
```
jmrodri: Where is the version information stored? What does a versioned chart url look like? Or more specifically how would someone get the specific version of a chart? What changes do you anticipate being done to the base image? Or is this another proposal for when we do phase 3?

docs/proposals/helm-charts.md
djzager: To my understanding, yes. Every call to `bootstrap` should create a new instance of a `HelmAdapter` for each of the `helm` type config entries, calling `GetImageNames()` and `FetchSpecs()` again.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+    white_list:
+      - ".*"
+```
+
+1. Type: refers to the registry adapter to handle this registry
+1. Name: gives a name to this registry item
+1. URL: the URL for the Helm chart repository
+1. Base Image: the container image to use when provisioning/deprovisioning a
+   Helm chart
+1. White List/Black List: allows the Admin to filter out Helm charts
```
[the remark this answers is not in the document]
djzager: Even the default dockerhub config includes the URL. My fear in setting a default, like [helm stable default](kubernetes-charts.storage.googleapis.com), is that most of them will not work. My preference would be for the editor of the configmap to specify from where to get the charts.

```
      - type: "dockerhub"
        name: "dh"
        url: "docker.io"
        org: "ansibleplaybookbundle"
        tag: "latest"
        white_list:
          - ".*-apb$"
```

As for the base_image, would this be a `const` in the HelmAdapter? Would we have a different one for a downstream broker?

docs/proposals/helm-charts.md
shawn-hurley: I don't think that is correct. Registries are made with an adapter as an unexposed field and then is used for subsequent calls. I do not think that bootstrap by itself is creating new adapters. Are you planning on adding some other behavior?

djzager: Not intending to change the behavior. This concern should be overcome by rebuilding the list of image names and charts on each call to `GetImageNames()`, for example:

```
// GetImageNames - retrieve the images
func (r *HelmAdapter) GetImageNames() ([]string, error) {
        var imageNames []string

        r.Charts = map[string][]*repo.ChartVersion{}
```

Then, the subsequent call to  `FetchSpecs(image_names)` will be operating on the newly built list of image names and charts.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Creating a Base Image for Executing Helm Charts
+
+Ansible Playbook Bundles (APBs) have already been described as an instance of
+the more generic Service Bundle. Here, we will be creating a new type of
+Service Bundle, Helm Bundles, that respect the Service Bundle contract but use
+Helm as the runtime. We already have a [Helm Bundle
+Base](https://github.com/ansibleplaybookbundle/helm-bundle-base) and the
+changes shown in [Helm Bundle Base PR 2](https://github.com/ansibleplaybookbundle/helm-bundle-base/pull/2)
+show the work required to implement the Service Bundle contract using Helm.
```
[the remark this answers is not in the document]
djzager: :sunglasses:

docs/proposals/helm-charts.md
djzager: I'm making it clear that a separate proposal should be made for this work. Handling basic repos is a sufficient first step and the work required to support authenticated repos would require investigation.

docs/proposals/helm-charts.md
djzager: I went ahead and made this part of the initial proposal, it's easy to get all of the chart's versions.

jmrodri:
@eriknelson this is waiting on your review.

djzager:
[(8) Helm v2 Support in Broker: Investigation](https://trello.com/c/T4FMGlhD/678-8-helm-v2-support-in-broker-investigation)

djzager:
[(8) Helm v2 Support in Broker: Investigation](https://trello.com/c/T4FMGlhD/678-8-helm-v2-support-in-broker-investigation)

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+    Image:       r.Config.BaseImage,
+    FQName:      chart.Name,
+    Tags:        chart.Keywords,
+    Description: chart.Description,
+    Metadata: map[string]interface{}{
+        //"longDescription":  chart.Description,
+        "displayName":      fmt.Sprintf("%s (Helm)", chart.Name),
+        "documentationUrl": chart.Home,
+        "dependencies":     chart.Sources,
+        "imageUrl":         chart.Icon,
```
mhrivnak: I still haven't been able to get this to work. Is there documentation somewhere that we can cite as the authority that this is the right key to use?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+    },
+    Plans: []apb.Plan{
+        apb.Plan{
+            Name:        "default",
+            Description: "Default plan for running helm charts",
+            Parameters: []apb.ParameterDescriptor{
+                apb.ParameterDescriptor{
+                    Name:      "repo",
```
mhrivnak: What is the use case for changing this at provision time?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+                    Title:     "Helm Chart Repository URL",
+                    Type:      "string",
+                    Default:   r.Config.URL.String(),
+                    Pattern:   fmt.Sprintf("^%s$", r.Config.URL.String()),
+                    Updatable: false,
+                    Required:  false,
+                },
+                apb.ParameterDescriptor{
+                    Name:      "chart",
```
mhrivnak: What is the use case for changing this at provision time? Hasn't the user already selected a chart by name through the catalog at this point?

docs/proposals/helm-charts.md
djzager: This is strange. I have had no issues getting this to work.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+for _, chart := range charts {
+    chartVersions = append(chartVersions, chart.Version)
+}
+```
+
+- Use the latest version of the chart to create a bundle `Spec` object:
+
+```
+// Use the latest chart for creating the bundle
+chart := charts[0]
```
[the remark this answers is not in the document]
djzager: what would that look like? https://github.com/kubernetes/helm/blob/master/pkg/repo/index.go#L141 sorts so we can reliably grab the latest version.

docs/proposals/helm-charts.md
djzager: There isn't one, but there is not currently a good way to hide it. That's why I use the `Pattern` to prevent you from changing it.

docs/proposals/helm-charts.md
djzager: Same as above.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+The addition of the `Charts` field makes it possible to save our work and only
+read the registries' index file once.
+
+When the broker calls `GetImageNames()` the Helm adapter will read the
+`index.yaml` found at the Helm Chart Repository URL based on the [chart
+repository
+structure](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#the-chart-repository-structure).
+The broker will create an entry in the `Charts` field to hold a list of
```
eriknelson: Just want to make sure "The broker" here isn't referring to the registry. We definitely don't want generic registry code doing helm specific stuff like writing to the chart cache, that would break the adapter abstraction. I'm thinking it should be the adapter before returning the image name list to the registry.

Also, these adapter methods are going to be called on every bootstrap. What's the condition that we're checking to prevent doing all the "work" involved with each bootstrap to load the cache? i.e.

```go
if cacheNotLoaded() { // What's this?
    adapter.loadCache()
}
```

eriknelson: Also, does the value of the chart's entry start out as an empty list prior to the `FetchSpecs` call?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+- Load the chart in memory to grab the content of the values file:
+
+```
+resp, err := http.Get(chart.URLs[0])
+if err != nil {
+        return specs, err
+}
+defer resp.Body.Close()
+
+helmChart, err := chartutil.LoadArchive(resp.Body)
```
eriknelson: :+1: this is pretty cool

docs/proposals/helm-charts.md
eriknelson: Is it possible to just strip these from the `ParameterDescriptor` list in some post-processing if there's no reason to have them there?

docs/proposals/helm-charts.md
eriknelson: See above.

docs/proposals/helm-charts.md
djzager: There is no case where a user should be changing these. However, at runtime we need the repo url. We can't hide these and still provide them at runtime. My solution was to make it the default and set the pattern to the exact string to prevent the user from changing it.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+    url: "https://kubernetes-charts.storage.googleapis.com"
+    base_image: "docker.io/djzager/helm-bundle-base:latest"
+    white_list:
+      - ".*"
+```
+
+1. Type: refers to the registry adapter to handle this registry
+1. Name: gives a name to this registry item
+1. URL: the URL for the Helm chart repository
+1. Base Image: the container image to use when provisioning/deprovisioning a
```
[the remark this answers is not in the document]
djzager: You are correct. It is not really a base image. Should probably make the project name and config item reflect that. `runner` or `runner_image` better reflects intent.

docs/proposals/helm-charts.md
eriknelson: Gotcha, so they have to be passed through somehow and we can't take a helm specific approach. I wonder if the UI could mark these fields as immutable somehow?

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Broker Config Options
+
+It is expected that a broker admin who wants to expose Helm charts would update
+the broker's config to look something like:
+
+```
+registry:
+  - type: helm
+    name: stable
```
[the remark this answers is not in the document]
djzager: Yes. The `unstable` variant is called `incubator` and the public URL is `https://kubernetes-charts-incubator.storage.googleapis.com/`.

docs/proposals/helm-charts.md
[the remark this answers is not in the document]
djzager: I'm operating under the assumption that the Helm Chart Repository's URL is a required configuration item.

docs/proposals/helm-charts.md
```diff
@@ -0,0 +1,206 @@
...
+// HelmAdapter - Helm Registry Adapter
+type HelmAdapter struct {
+    Config Configuration
+    Log    *logging.Logger
+    Charts map[string][]*repo.ChartVersion
+}
+```
+
+The addition of the `Charts` field makes it possible to save our work and only
+read the registries' index file once.
```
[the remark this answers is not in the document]
djzager: We are rebuilding this cache every time `GetImageNames()` is called. I'm not totally familiar with how often the registry adapters are loaded and `GetImageNames()` and then `FetchSpecs(imageNames)` are called.

docs/proposals/helm-charts.md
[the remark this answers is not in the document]
mhrivnak: Ah, nice! That'll do then if they're making the guarantee.

docs/proposals/helm-charts.md
mhrivnak: Gotcha! Maybe we need an RFE for the broker to provide a way for a registry adapter to add hidden metadata that gets passed through to the executor and into the running bundle? It would be interesting to brainstorm about other possible use cases for that.

docs/proposals/helm-charts.md
djzager: Updating to make it clear that the "the broker" is actually the registry. The `Charts` field will be on the `HelmAdapter` struct. Calls to `GetImageNames()` will create the Charts map and `FetchSpecs` will use that map to create the Specs.

eriknelson:
LGTM!

Issue #674, opened 2018-01-22
make serviceInstance details available to the playbook


jaymccon:
# Feature:

As an APB developer it is useful to have the details of who provisioned a service. In the context of the AWS service broker this would enable:
* tagging of the AWS resources so that orphaned resources can be detected and cleaned up
* generation of a unique key/idempotency token for resource provisioning
* usage costs can be allocated to an openshift cluster/project

Making (at least) the following values available to the provision and de-provision playbooks at runtime would allow for this:
* OpenShift cluster dns suffix
* username
* namespace
* ServiceInstance id

johnkim76:
@jaymccon the [PR#832](https://github.com/openshift/ansible-service-broker/pull/832) was merged which provides the username information of the last requesting user.  The details of the `_apb_last_requesting_user` parameter is discussed [here](https://github.com/openshift/ansible-service-broker/blob/master/docs/service-bundle.md#last-request-user). 

I'm still working on the `openshift_routing_suffix`

jmrodri:
I'm calling this one done. If we really need the `openshift_routing_suffix` let's open a new issue. Fixed by PR #832

PR #832, opened 2018-03-12
make username available for APB in the serviceInstance

johnkim76:
username information was requested to be available for APBs

 - pass the username info from the broker to the APB during it's executions as a parameter in extra_vars

#### Addresses the following Issue
* [Issue #674](https://github.com/openshift/ansible-service-broker/issues/674)

### DIFF
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -131,7 +131,7 @@ Example of JSON document passed to __bind__:
 
 #### Last Request User
 
-The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ is available in the `_apb_last_requesting_user` parameter. The parameter is set to the `UID` if the requesting `username` is empty (e.g. auto escalation). However, this parameter may be completely empty as it's not a requirement to send the requesting user information. Also, the user info of the latest request will always overwrite the existing value.
+The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ is available in the `_apb_last_requesting_user` parameter. The parameter is set to the `UID` if the  `username` of the action is empty (e.g. auto escalation). However, this parameter may be completely empty as it's not a requirement to send the requesting user information. The user for the current action may be different from the users that initiated any previous actions on the same resource. The _apb_last_requesting_user will always reflect the username of this current action.
 
 This field was introduced in version 1.2
 
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -131,9 +131,9 @@ Example of JSON document passed to __bind__:
 
 #### Last Request User
 
-The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ will available in the `_apb_last_requesting_user` parameter. The parameter value will be the `UID` of the requesting user if the `username` is not available (e.g. auto escalation). Also, the user info of the latest request will always overwrite the existing value.
+The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ is available in the `_apb_last_requesting_user` parameter. The parameter is set to the `UID` if the requesting `username` is empty (e.g. auto escalation). Also, the user info of the latest request will always overwrite the existing value.
 
-This feature will be available in v3.10
+This field was introduced in version 1.2
 
 ### Environment Variables
 
--- a/docs/service-bundle.md
+++ b/docs/service-bundle.md
@@ -17,11 +17,11 @@ name: example-apb   # The name of the Service Bundle
 description:        # A short description of the Service Bundle
 bindable: True      # Whether this Service Bundle can be bound to other services
 async: optional     # Does this Service Bundle support asynchronous provision
-metadata: 
+metadata:
   documentationUrl: <link to documentation>
   imageUrl: <link to URL of image>
   dependencies:
-  - '<registry>/<organization>/<dependency-name-1>'   
+  - '<registry>/<organization>/<dependency-name-1>'
   - '<registry>/<organization>/<dependency-name-2>'
   displayName: Example App (APB)
   longDescription: A longer description of what this APB does
@@ -79,6 +79,7 @@ Example of JSON document passed to __provision__:
     "_apb_plan_id": "default",
     "_apb_service_class_id": "c23ec213bb8dea1577230c5ce005b9c2",
     "_apb_service_instance_id": "54636ad3-0378-49a1-a494-f97d0a0daf8e",
+    "_apb_last_requesting_user": "admin",
     "mediawiki_admin_pass": "pass",
     "mediawiki_admin_user": "admin",
     "mediawiki_db_schema": "mediawiki",
@@ -118,6 +119,7 @@ Example of JSON document passed to __bind__:
     },
     "_apb_service_class_id": "1dda1477cace09730bd8ed7a6505607e",
     "_apb_service_instance_id": "d6ac6b10-fbff-4944-8e6b-3478313e20d1",
+    "_apb_last_requesting_user": "admin",
     "cluster": "kubernetes",
     "namespace": "default"
 }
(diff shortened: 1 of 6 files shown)


johnkim76:
[(3) AWS Broker RFE: make serviceInstance details available to the playbook #674](https://trello.com/c/dsgVkgjz/632-3-aws-broker-rfe-make-serviceinstance-details-available-to-the-playbook-674)

pkg/handler/handler.go
```diff
@@ -274,6 +274,7 @@ func (h handler) provision(w http.ResponseWriter, r *http.Request, params map[st
 		return
 	}
 
+	var username = "default"
```
jmrodri: this is a little weird. I would've left it empty and if it is still empty when we call the apb, we don't pass anything to the apb.

pkg/apb/types.go
```diff
@@ -224,6 +224,8 @@ const (
 	ClusterKey = "cluster"
 	// NamespaceKey parameter name passed to APBs
 	NamespaceKey = "namespace"
+	// UsernameKey parameter name passed to APBs
+	UsernameKey = "username"
```
jmrodri: @eriknelson @shawn-hurley is this supposed to be `_apb_username` or is `username` okay?

shawn-hurley:
I think that if we are planning on passing this down to the APB we should pass it down for all methods, not just provision IMO.

johnkim76:
@shawn-hurley ok,  I will update.  I wasn't sure if it was necessary to do so for all methods, but I can make the changes.

pkg/apb/types.go
johnkim76: Would it make more sense place the username [here](https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/types.go#L33) and set it with the others [here](https://github.com/openshift/ansible-service-broker/blob/master/pkg/broker/broker.go#L592) and NOT put it in the ServiceInstance?  I put it in the serviceInstance so that it retains the username info in the instance... but I can change it.

pkg/handler/handler.go
johnkim76: I can update this...

pkg/apb/types.go
jmrodri: @johnkim76 I think it's okay on the ServiceInstance, Since an instance should never have multiple users requesting it.

jmrodri:
> I think that if we are planning on passing this down to the APB we should pass it down for all methods, not just provision IMO.
@shawn-hurley good catch.

pkg/apb/types.go
shawn-hurley: Consider the scenario where the broker is set to `auto_esclate` UserA and UserB have enough access in the target namespace. UserA creates a service `postgresql` and then UserB creates a binding for that service. That would be two users requesting the same service instance.

I agree @johnkim76 that this key should be _apb_requesting_user  and should behave similarly those that you posted above.

pkg/broker/broker.go
```diff
@@ -485,7 +485,7 @@ func (a AnsibleBroker) Catalog() (*CatalogResponse, error) {
 }
 
 // Provision  - will provision a service
-func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, async bool,
+func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, async bool, username string,
```
shawn-hurley: You will need to do the same for all the other actions to truely pass this for every method. 

@eriknelson @jmrodri What do you guys think of having him add the requesting user `UserInfo` to the provision request?

pkg/handler/handler.go
```diff
@@ -289,11 +290,13 @@ func (h handler) provision(w http.ResponseWriter, r *http.Request, params map[st
 			writeResponse(w, status, broker.ErrorResponse{Description: err.Error()})
 			return
 		}
+
+		username = userInfo.Username
```
shawn-hurley: I would suggest that you may need to add logic because it could be the case that we do not have `Username` but `UID` from the `UserInfo` of the requesting user.

pkg/broker/broker.go
jmrodri: @shawn-hurley ProvisionRequest sounds better, but it usually maps to the OSB ProvisionRequest. Unless we add it as part of the parameters, then I'd be okay with that.

pkg/broker/broker.go
```diff
@@ -660,7 +663,7 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
 
 // Deprovision - will deprovision a service.
 func (a AnsibleBroker) Deprovision(
-	instance apb.ServiceInstance, planID string, skipApbExecution bool, async bool,
+	instance apb.ServiceInstance, planID string, skipApbExecution bool, async bool, userInfo UserInfo,
```
shawn-hurley: I don't see the update of the `parameters[requestingUserKey] = getRequestingUser(userInfo)` for Deprovision or Unbind or Update. 

I think we need to do this be a separate user could request those actions and if we want to alert the APBs of the users that are interacting with them, I think that we should update it. I wonder if we should re-name this field then to `_apb_last_requesting_user`?

johnkim76: I have updated as you advised, to override the userInfo in the Deprovision, Unbind, and Update

pkg/handler/handler.go
```diff
@@ -646,8 +646,8 @@ func (h handler) unbind(w http.ResponseWriter, r *http.Request, params map[strin
 		return
 	}
 
+	userInfo, ok := r.Context().Value(UserInfoContext).(broker.UserInfo)
```
shawn-hurley: We need to make sure it is clear that this could be empty. If the service catalog does not have the correct feature gate on we will not get this header. I think this should be fine but is a documentation on the service bundle contract.

docs/service-bundle.md
```diff
@@ -127,12 +129,18 @@ Example of JSON document passed to __bind__:
   provision action for this service instance. Details are below in the Output
   section.
 
+#### Last Request User
+
+The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ will available in the `_apb_last_requesting_user` parameter. The parameter value will be the `UID` of the requesting user if the `username` is not available (e.g. auto escalation). Also, the user info of the latest request will always overwrite the existing value.
+
+This feature will be available in v3.10
```
mhrivnak: I think this should be a broker version, right?

mhrivnak: As a nit-pick, I think it makes sense for these to be present-tense. "This field is new in version X.Y." Or even past tense: "This field was introduced in version X.Y."

Users aren't likely to care about this field or read about it until after its release.

eriknelson: `release-1.2`?

pkg/broker/broker.go
```diff
@@ -708,6 +711,13 @@ func (a AnsibleBroker) Deprovision(
 		instance.Parameters = &params
 	}
 
+	// Override the lastRequestingUserKey value in the instance.Parameters
+	if instance.Parameters != nil {
+		params := *instance.Parameters
```
eriknelson: So Parameters here is already a reference type, you don't have to deref and then copy the address back to Parameters. I'd update these on each to just be a one liner:

```
(*instance.Parameters)[lastRequestingUserKey] = getLastRequestingUser(userInfo)
```

https://play.golang.org/p/wG6632X68h7
```

Comment (johnkim76) [reply pkg/broker/broker.go]: thanks, I will update

File: docs/service-bundle.md (hunk 7 above)
Comment (johnkim76) [reply docs/service-bundle.md]: oh, ok, so I'll make those updates

Comment (mhrivnak) [reply docs/service-bundle.md]: I think I'd go with just the version number rather than a branch name, but either way I think that's the best number to convey.

File: docs/service-bundle.md [hunk 9]
```diff
@@ -127,12 +129,18 @@ Example of JSON document passed to __bind__:
   provision action for this service instance. Details are below in the Output
   section.
 
+#### Last Request User
+
+The requesting username of the [actions](#actions) _provision_, _deprovision_, _bind_, _unbind_, and _update_ is available in the `_apb_last_requesting_user` parameter. The parameter is set to the `UID` if the requesting `username` is empty (e.g. auto escalation). Also, the user info of the latest request will always overwrite the existing value.
```
shawn-hurley: I think we should note that this could be completly empty as it is not a requirement to send the requesting user.

Sorry @johnkim76 one more change?

johnkim76: thanks, will update

PR #833, opened 2018-03-12
Bug 1554239 - overwriting job state that has been finished.

shawn-hurley:
removes the overwriting of the job state.

**Race condition**:
Skip APB run -> send a successful message -> then broker code sets status to in progress.
Making the job run forever.

 - Don't set the state to in progress if we are skipping the APB run.

fixes 1554239


jmrodri:
@shawn-hurley commit message and PR name has a typo: overwritting should be overwriting.

shawn-hurley:
The PR name is what the commit message will be when you squash.

Issue #675, opened 2018-01-22
custom error messaging 


jaymccon:
# Feature:

Allowing an APB to return custom error messages that gets surfaced by service catalog if a provisioning operation fails will ease troubleshooting and improve customer experience.

rthallisey:
I think this being implemented here https://github.com/openshift/ansible-service-broker/issues/475.  Let me know if this is different than last_operation.

shawn-hurley:
https://github.com/openshift/ansible-service-broker/pull/607 

This is a start of this I think.

jwmatthews:
@jaymccon (Amazon developer on AWS Service Broker) is looking for the Web Console to display a useful error when there is an error running an AWS APB.

Let's say provisioning RDS fails with a CloudFormation error.
The APB needs to grab the correct error message and send back to Broker...then we need the error message to flow up to end user at web-console so they see "RDS APB failed with CloudFormation error blah......", opposed to "Provision Failed".

This issue may involve some investigation beyond just Broker side.

maleck13:
@rthallisey @jwmatthews I think using the last_operation_description module along with the PRs associated with it would solve this as it would allow the developer to set the last_operation description and that then flows through to the catalog and to the web console under the service instance.
Related PRS

**Module** https://github.com/ansibleplaybookbundle/ansible-asb-modules/pull/9**
**Broker** https://github.com/openshift/ansible-service-broker/pull/619

The work is largely in place now will need to rebase the 619 PR

eriknelson:
I would think the `Status` of a `ServiceInstance` is a canonical way to surface that information. Would also expect to find the last_operation.description there.

maleck13:
Yes I believe the last operation is updated on the service instance .

rthallisey:
@jaymccon I think we're going to track this from #619.

> I would think the Status of a ServiceInstance is a canonical way to surface that information. Would also expect to find the last_operation.description there.
>
Here's @maleck13's test case: https://github.com/openshift/ansible-service-broker/pull/619#issuecomment-359875771

PR #837, opened 2018-03-19
Display Custom Error Message for the APB via termination-log

johnkim76:
If the APB provision fails, this PR will check for a non-empty termination message, and will display it in the WebUI, instead of the generic error message.

fixes [#675](https://github.com/openshift/ansible-service-broker/issues/675)

### DIFF
--- a/pkg/apb/watch_pod.go
+++ b/pkg/apb/watch_pod.go
@@ -36,7 +36,7 @@ var (
 
 type updateDescriptionFn func(string)
 
-// ErrorCustomMsg - custom termination msg from pod's 'terminationMessagePath'
+// ErrorCustomMsg - An error to propagate the custom error message to the callers
 type ErrorCustomMsg struct {
 	msg string
 }
--- a/pkg/apb/watch_pod.go
+++ b/pkg/apb/watch_pod.go
@@ -36,6 +36,22 @@ var (
 
 type updateDescriptionFn func(string)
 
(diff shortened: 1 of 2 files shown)


johnkim76:
[(8) AWS Broker RFE: Enhance error messages, so when a provision request fails with a CloudFormation error, the error is preserved and displayed to end user in web console.](https://trello.com/c/XqdHB5Ch/633-8-aws-broker-rfe-enhance-error-messages-so-when-a-provision-request-fails-with-a-cloudformation-error-the-error-is-preserved-and)

djzager:
Could you provide an APB to be used to see this in action?

johnkim76:
@djzager https://github.com/johnkim76/lastop-cem-demo-apb
docker.io/jkig/lastop-cem-demo-apb

pkg/apb/watch_pod.go
```diff
@@ -36,6 +36,22 @@ var (
 
 type updateDescriptionFn func(string)
 
+// ErrorCustomMsg - custom termination msg from pod's 'terminationMessagePath'
```
shawn-hurley: >custom termination msg from pod's 'terminationMessagePath'

I think that this is generic enough that it could be used for any type of custom error message. What do you think of changing this to just say  `An error to propagate the custom error message to the callers`

johnkim76: I will update

PR #839, opened 2018-03-19
[wip] Update rbac.authorization.k8s.io to use v1 API instead of v1beta1

djwhatle:
Updates the apiVersion for ClusterRoleBindings to use v1 API instead of v1beta1. Bringing up ASB in a Kubernetes cluster fails without this when using catasb in the local or EC2 env.

Changes made to three files:

- templates/deploy-ansible-service-broker.template.yaml
- templates/k8s-ansible-service-broker.yaml.j2
- templates/simple-broker-template.yaml

### DIFF
--- a/templates/deploy-ansible-service-broker.template.yaml
+++ b/templates/deploy-ansible-service-broker.template.yaml
@@ -47,7 +47,7 @@ objects:
     name: asb
     namespace: "${NAMESPACE}"
 
-- apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


djwhatle:
Marked as [wip], still some issues to resolve. Need to remove changes from the standard and non k8s template per conversation with @shawn-hurley , also need additional adjustments before this will work with existing automation.

djwhatle:
API version wasn't the root issue in this case, closing.

PR #840, opened 2018-03-19
Add new dependencies for helm

djzager:

In order to support the helm registry adapter, we need to add some helm dependencies.

Add helm as a dependency.


djzager:
[(8) Helm v2 Support in Broker: Investigation](https://trello.com/c/T4FMGlhD/678-8-helm-v2-support-in-broker-investigation)

djzager:
https://github.com/djzager/ansible-service-broker/blob/helm-ra/pkg/registries/adapters/helm_adapter.go will show you the helm_adapter although that isn't really up for review yet.

1. The [`Charts` field of the HelmAdapter](https://github.com/djzager/ansible-service-broker/blob/helm-ra/pkg/registries/adapters/helm_adapter.go#L42) is holding a mapped array of [Helm ChartVersions](https://github.com/kubernetes/helm/blob/master/pkg/repo/index.go#L216)
1. I tend to agree that [LoadArchive()](https://github.com/djzager/ansible-service-broker/blob/helm-ra/pkg/registries/adapters/helm_adapter.go#L99) is nothing special, the nice thing provided by [`LoadArchive`](https://github.com/kubernetes/helm/blob/master/pkg/chartutil/load.go#L66) is a Chart object from which values can be grabbed.

PR #842, opened 2018-03-20
minor updates to proposal to clarify a few points

maleck13:
Minor changes to ensure the proposal lines up with what will be done.

### DIFF
--- a/docs/proposals/apb-state-support.md
+++ b/docs/proposals/apb-state-support.md
@@ -45,7 +45,7 @@ This module will expose an API to the APB developer for setting key-value pairs.
 
 Under the hood, this APB module would take the key-value pair, and store it in a
 ConfigMap named ```$POD_NAME```. This ConfigMap would live within the
-temporary namespace ```$POD_NAMESPACE``` and be created by the Service Bundle when with the first call to ```asb_set_state```.
+temporary namespace ```$POD_NAMESPACE``` and be created by the Service Bundle on the first call to ```asb_set_state```.
 
 ## Update broker to manage Service Bundle created state ConfigMaps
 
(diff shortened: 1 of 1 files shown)


docs/proposals/apb-state-support.md
```diff
@@ -32,25 +32,29 @@ not trustworthy or reliable.
...
+- name: get some state
+  asb_get_state:
+    key: "key"
 
 ```
 
 Under the hood, this APB module would take the key-value pair, and store it in a
 ConfigMap named ```$POD_NAME```. This ConfigMap would live within the
-temporary namespace ```$POD_NAMESPACE``` and be created by the broker before the APB pod was created.
+temporary namespace ```$POD_NAMESPACE``` and be created by the Service Bundle when with the first call to ```asb_set_state```.
```
shawn-hurley: > Service Bundle when with the first call to...

Service Bundle on the first call to... ?

docs/proposals/apb-state-support.md
```diff
@@ -32,25 +32,29 @@ not trustworthy or reliable.
...
 
 ## Update broker to manage Service Bundle created state ConfigMaps
 
-To ensure the state is persisted across Service Bundle actions, the broker will create a ConfigMap within the ```$POD_NAMESPACE``` named ```$POD_NAME```.
 After an action was successfully completed (ie the Service Bundle exited with a 0 exit code) and before the sandbox namespace was removed, the broker would copy the ConfigMap back to the broker's namespace and name it ```<ServiceInstanceID>-state```. If a ConfigMap with
 that name was already present, the broker would update and append the values.
 There should only ever be one ConfigMap per ServiceInstance. The ConfigMap would be removed from the broker's namespace
-once the the deprovision action completed successfully.
+once the the deprovision action was invoked for that ServiceInstance.
```
shawn-hurley: This is fine for now. I think that soon we will need to worry about subsequent deprovsion calls after a failed deprovsion. I think this is something that the service catalog is working on.

maleck13: ok good to know.

maleck13:
@shawn-hurley updated wording

PR #846, opened 2018-03-21
Helm Registry Adapter

djzager:

Adds a Helm Registry Adapter without vendoring helm.


pkg/registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,322 @@
...
+	if err != nil {
+		return index, err
+	}
+
+	err = yaml.Unmarshal(body, index)
+	if err != nil {
+		return index, err
+	}
+	for _, versions := range index.Entries {
+		sort.Sort(sort.Reverse(versions))
```
shawn-hurley: I don't think that we need to sort twice

pkg/registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,322 @@
...
+			continue
+		}
+
+		for _, chart := range charts {
+			chartVersions = append(chartVersions, chart.Version)
+		}
+		// Use the latest chart for creating the bundle
+		chart := charts[0]
+
+		resp, err := http.Get(chart.URLs[0])
```
shawn-hurley: We may want to check that this has URLs

pkg/registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,322 @@
...
+		charts, ok := r.Charts[name]
+		if !ok {
+			continue
+		}
+
+		for _, chart := range charts {
+			chartVersions = append(chartVersions, chart.Version)
+		}
+		// Use the latest chart for creating the bundle
+		chart := charts[0]
```
shawn-hurley: Might want to make sure that charts is not empty

pkg/registries/adapters/helm_adapter.go
djzager: I'm going to defer to your knowledge of Golang here, but this does appear in line with https://golang.org/pkg/sort/#Reverse.

eriknelson: Looks like `sort.Sort` executes the sort built up on a `sort.Interface` wrapped type:

https://play.golang.com/

Note: It mutates the original slice, you aren't getting a copy back.

pkg/registries/adapters/helm_adapter.go
djzager: I am struggling to see what more to do beyond what's in `GetImageNames()`. If there are no charts, we don't add it to our list.

```
if len(entry) == 0 {
    continue
}
```

Gopkg.toml
```diff
@@ -87,3 +87,7 @@
 [[constraint]]
   name = "k8s.io/kubernetes"
   version = "1.9.1"
+
+[[constraint]]
```
djzager: Just want to point out that I did not fix the problem that I had with vendor:

```
$ make vendor
Solving failure: No versions of github.com/automationbroker/broker-client-go met constraints:
        master: Could not introduce github.com/automationbroker/broker-client-go@master, as it has a dependency on k8s.io/client-go with constraint release-6.0, which has no overlap with existing constraint kubernetes-1.9.1 from (root)
        master: Could not introduce github.com/automationbroker/broker-client-go@master, as it has a dependency on k8s.io/client-go with constraint release-6.0, which has no overlap with existing constraint kubernetes-1.9.1 from (root)
        v1.0.0: Could not introduce github.com/automationbroker/broker-client-go@v1.0.0, as it has a dependency on k8s.io/client-go with constraint release-6.0, which has no overlap with existing constraint kubernetes-1.9.1 from (root)
make: *** [Makefile:18: vendor] Error 1
```
I just worked around it since it wasn't obvious to me the correct way to fix it.

pkg/registries/adapters/helm_adapter_test.go
```diff
@@ -0,0 +1,132 @@
...
+	imageNames, err := ha.GetImageNames()
+	if err != nil {
+		t.Fatal("ERROR: ", err)
+	}
+	ft.AssertEqual(t, len(imageNames), 1)
+	ft.AssertEqual(t, imageNames[0], MariaDB)
+	ft.AssertNotNil(t, imageNames)
+
+	// Override Chart URL to point to our http server
+	ha.Charts[MariaDB][0].URLs[0] = url.String() + MariaDBPath
```
djzager: @jmrodri Thank you for telling me about `testdata` that made this much easier.

djzager:
Will create another PR for just the documentation piece. Probably need to have registry configuration documentation in the `bundle-lib` also.

PR #848, opened 2018-03-22
Use bundle lib

shawn-hurley:
changes to allow for use of bundle lib in this project


shawn-hurley:
* moving pkg/apb, pkg/clients, pkg/origin, pkg/registries, pkg/registries/adapters, pkg/runtime to bundle-lib
* vendor bundle-lib
* move CRD conversion logic in DAO to use bundle-lib
* update `NewRuntime` call in `app.go` to use new configuration struct

PR #850, opened 2018-03-22
Bug 1558472 - Update openshift registry adapter to use www-auth redirect

dymurray:
Updates `openshift` registry  to intelligently grab the auth endpoint from the `www-authenticate` header.

 - Hit <registry>/v2/ to get `www-authenticate` header
 - Follow redirect to get Bearer token
 - Now attempt to get manifests

### DIFF
--- a/pkg/registries/adapters/openshift_adapter.go
+++ b/pkg/registries/adapters/openshift_adapter.go
@@ -27,9 +27,8 @@ import (
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 )
 
-const openShiftName = "registry.connect.redhat.com"
-const openShiftAuthURL = "https://registry.connect.redhat.com/auth/realms/rhc4tp/protocol/redhat-docker-v2/auth?service=docker-registry"
-const openShiftManifestURL = "https://registry.connect.redhat.com/v2/%v/manifests/%v"
+const openShiftName = "openshift"
+const openShiftManifestURL = "%v/v2/%v/manifests/%v"
 
 // OpenShiftAdapter - Docker Hub Adapter
 type OpenShiftAdapter struct {
@@ -135,7 +134,7 @@ func (r OpenShiftAdapter) loadSpec(imageName string) (*apb.Spec, error) {
 	if r.Config.Tag == "" {
 		r.Config.Tag = "latest"
 	}
-	req, err := http.NewRequest("GET", fmt.Sprintf(openShiftManifestURL, imageName, r.Config.Tag), nil)
+	req, err := http.NewRequest("GET", fmt.Sprintf(openShiftManifestURL, r.Config.URL, imageName, r.Config.Tag), nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/registries/adapters/openshift_adapter.go
```diff
@@ -18,18 +18,17 @@ package adapters
...
-	b64 "encoding/base64"
+	"strings"
 
 	logging "github.com/op/go-logging"
 	"github.com/openshift/ansible-service-broker/pkg/apb"
 )
 
-const openShiftName = "registry.connect.redhat.com"
-const openShiftAuthURL = "https://sso.redhat.com/auth/realms/rhc4tp/protocol/docker-v2/auth?service=docker-registry"
-const openShiftManifestURL = "https://registry.connect.redhat.com/v2/%v/manifests/%v"
```
shawn-hurley: 👍 I like this is configurable

pkg/registries/adapters/openshift_adapter.go
```diff
@@ -83,17 +82,40 @@ func (r OpenShiftAdapter) getOpenShiftAuthToken() (string, error) {
...
+		return "", err
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	// Ensure that response holds data we expect
+	if resp.Header["Www-Authenticate"] == nil {
```
shawn-hurley: I would prefer to acesss the header value using the helper method [here](https://golang.org/pkg/net/http/#Header.Get)

```go
if resp.Header.Get("Www-Authenticate") == "" {
```

pkg/registries/adapters/openshift_adapter.go
```diff
@@ -83,17 +82,40 @@ func (r OpenShiftAdapter) getOpenShiftAuthToken() (string, error) {
...
+		return "", errors.New("failed to find www-authenticate header from response")
+	}
 
-	req, err := http.NewRequest("GET", openShiftAuthURL, nil)
+	authChallenge := resp.Header["Www-Authenticate"][0]
```
shawn-hurley: Same as above, in fact, you might be able to remove the above check if you use the helper method.

pkg/registries/adapters/openshift_adapter.go
```diff
@@ -83,17 +82,40 @@ func (r OpenShiftAdapter) getOpenShiftAuthToken() (string, error) {
...
+	if strings.Index(authChallenge, "realm=\"") == -1 {
+		return "", errors.New("failed to find realm in www-authenticate header")
+	}
+	authRealm := strings.Split(strings.Split(authChallenge, "realm=\"")[1], "\"")[0]
+	authOptions := strings.Split(authChallenge, ",")[1]
```
shawn-hurley: I don't think there is a check for `,` in the auth challenge. this might panic if we got bad data for the header

dymurray:
Thanks Shawn, will make these changes and change the PR against release-1.1 branch

PR #857, opened 2018-03-26
Remind the user to give the developer cluster-admin

rthallisey:
Lots of folks use ```run_latest_build.sh``` with the apb tool.  But, when ```run_latest_build.sh``` completes, the script suggests the user login to the developer user in order to use the apb tool.  The developer user by default doesn't have permission to contact the broker in the ansible-service-broker namespace so the apb tool isn't going to work.

 - Print a command to give the developer user cluster-admin

### DIFF
--- a/scripts/run_latest_build.sh
+++ b/scripts/run_latest_build.sh
@@ -131,3 +131,5 @@ fi
 #
 
 echo 'NOTE: You are currently logged in as "system:admin", if you intend to use the apb tool, is is required you log in as a user with a token. "developer" is recommended.'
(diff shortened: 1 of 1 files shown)


eriknelson:
`cluster-admin` seems okay for this script, since it was never intended to support regular apb devlopment and use (it was "hey, try out the broker really quick!"). Unfortunately, **a lot** of people continue to rely on it.

FYI, this template is used by minishift to permission `developer` with all of its rules that it requires and nothing more. https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/templates/openshift-permissions.template.yaml

PR #860, opened 2018-03-27
removing config package from bundle lib. bump to new release

shawn-hurley:
* bump automationbroker/bundle-lib to 0.1.1


pkg/app/app.go
```diff
@@ -201,7 +201,25 @@ func CreateApp() App {
...
+			Type:       config.GetString("type"),
+			Name:       config.GetString("name"),
+			Images:     config.GetSliceOfStrings("images"),
+			Namespaces: config.GetSliceOfStrings("namespaces"),
+			Fail:       config.GetBool("fail_on_error"),
+			WhiteList:  config.GetSliceOfStrings("white_list"),
+			BlackList:  config.GetSliceOfStrings("black_list"),
+			AuthType:   config.GetString("auth_type"),
+			AuthName:   config.GetString("auth_name"),
+			Runner:     config.GetString("runner"),
```
shawn-hurley: @djzager Making sure this is what you were expecting.

Gopkg.lock
```diff
@@ -7,6 +7,12 @@
   packages = ["."]
   revision = "75cd24fc2f2c2a2088577d12123ddee5f54e0675"
 
+[[projects]]
+  name = "github.com/Masterminds/semver"
+  packages = ["."]
+  revision = "15d8430ab86497c5c0da827b748823945e1cf1e1"
+  version = "v1.4.0"
+
```
jmrodri: Why does bumping bundle-lib bring in a new semver? I never understood why we can't just bring in a single updated library.

pkg/app/app.go
jmrodri: Why are we copying configs everywhere now? Feels like I missed something.

Gopkg.lock
shawn-hurley: because semver was added to bundle lib for @djzager helm adapter.

shawn-hurley:
We are moving all uses of the `automationbroker/config` package back into the automation broker, and making all the configs for bundle-lib typed structs.

PR #862, opened 2018-03-28
Fix docs - enabling RBAC for minikube

jwerak:
Updates documentation for starting minikube:
- enabling RBAC for minikube 
- removing specified version (minikube has only 1.9.0 or latest, current 1.9.3 didn't work)

only docs changes above 

no

none

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Run the following from the root of the cloned git repository.
 
 # Adjust the version to your liking. Follow installation docs
 # at https://github.com/kubernetes/minikube.
-minikube start --kubernetes-version v1.9.3
+minikube start --extra-config=apiserver.Authorization.Mode=RBAC
 
 # Install helm and tiller. See documentation for obtaining the helm
 # binary. https://docs.helm.sh/using_helm/#install-helm
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Run the following from the root of the cloned git repository.
 
 # Adjust the version to your liking. Follow installation docs
 # at https://github.com/kubernetes/minikube.
-minikube start --extra-config=apiserver.Authorization.Mode=RBAC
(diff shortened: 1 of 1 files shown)


djzager:
Thank you for the submission @jwerak .

mhrivnak:
Hold on with this. I'll explain shortly.

djzager:
@mhrivnak Do you have a link to how to use the `kubeadm` bootstrapper?

djzager:
Found it :sunglasses:

jwerak:
@mhrivnak Thanks, I've had no idea about custom bootstrap in minikube, good to know, thanks 👍 

I've tested following is working for me now:
```
minikube start --bootstrapper kubeadm
# AND 
minikube start --bootstrapper kubeadm --kubernetes-version v1.9.4
```

Should I keep the version with version or without?

mhrivnak:
> Should I keep the version with version or without?

I like making it clear that setting the version is an option, and I recommend doing so as a normal practice. `minikube` makes it easy to iterate on standing up and tearing down a cluster. You'd hate to suddenly start getting 1.10 instead of 1.9, which changes something you depended on; you might not realize that a version change happened at all, and if it did break something, once you figured out that a version change was responsible, you might not know which version you were running before.

TL;DR it seems worth setting the version explicitly to get a consistent and reliable experience. But maybe that should be called out in the documentation more directly.

Does that make sense? What do you think?

jwerak:
@mhrivnak consistent and reliable makes sense to me too, new commit has landed.

README.md
```diff
@@ -217,5 +217,5 @@ make help
 ```
 
 # License
-
+git@github.com:prgcont/workshop-ASB.git
```
mhrivnak: Is this line intended to be here?

jwerak: ups, little advertisement can't hurt they say. 
Sorry for that. 
It's gone now.

PR #863, opened 2018-03-28
make more than one stage for all the tasks we want to do.

shawn-hurley:
* this should make it easier to see what failed
* should allow for easier gate debugging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
 
 jobs:
   include:
-    - stage: "CI"
+    - stage: "build"
       script:
         - ./scripts/travis.sh format
     -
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,25 +9,36 @@ env:
 
 jobs:
   include:
-    - stage: "OpenShift and Kubernetes CI"
+    - stage: "Format"
+      script:
+        - ./scripts/travis.sh format
+    - stage: "Vet"
```
djzager: Can "Format", "Vet", and "Lint" all be put under a single stage "Lint"?

.travis.yml
```diff
@@ -9,25 +9,36 @@ env:
...
+      script:
+        - ./scripts/travis.sh lint
+    - stage: "Build"
+      script:
+        - ./scripts/travis.sh build
+    - stage: "Test"
+      script:
+        - ./scripts/travis.sh build
+
+    - stage: "OpenShift CI"
```
djzager: I think this comment is outside the scope of this PR, but I was thinking it would be nice if these "OpenShift CI" and "Kubernetes CI" were separate tests belonging to the same stage "Functional Testing".

PR #870, opened 2018-04-02
Add migration command

shawn-hurley:
We need a way to migrate from ETCD to CRD for openshift ansible
 - create a new command and add to RPM and Images
 - connects to etcd and crd dao's and will convert etcd to crd

Trello: https://trello.com/c/KlFLeFI0

example job.yaml:

```yaml
---
apiVersion: batch/v1
kind: Job
metadata:
  name: asb-etcd-migration
spec:
  parallelism: 1
  completions: 1
  template:
    metadata:
      name: asb-etcd-migration
    spec:
      containers:
        - name: asb
          image: 'docker.io/shurley/origin-ansible-service-broker:migration'
          imagePullPolicy: Always
          # TODO: Need to update to proper entrypoint once it exists
          command:
            - '/usr/bin/migration'
          args:
            - '-host=asb-etcd.ansible-service-broker.svc'
            - '-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt'
            - '-client-cert=/var/run/asb-etcd-auth/client.crt'
            - '-client-key=/var/run/asb-etcd-auth/client.key'
            - '-namespace=ansible-service-broker'
          volumeMounts:
          - name: config-volume
            mountPath: /etc/ansible-service-broker
          - name: asb-tls
            mountPath: /etc/tls/private
          - name: asb-etcd-auth
            mountPath: /var/run/asb-etcd-auth
          env:
            - name: BROKER_CONFIG
              value: /etc/ansible-service-broker/config.yaml
            - name: HTTP_PROXY
              value: ""
            - name: HTTPS_PROXY
              value: ""
      volumes:
        - name: config-volume
          configMap:
            name: broker-config
            items:
              - key: broker-config
                path: config.yaml
        - name: asb-tls
          secret:
            secretName: asb-tls
        - name: asb-etcd-auth
          secret:
            defaultMode: 420
            secretName: broker-etcd-auth-secret
      restartPolicy: Never
      serviceAccount: asb
      serviceAccountName: asb
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ vendor: ## Install or update project dependencies
 
 broker: $(SOURCES) ## Build the broker
 	go build -i -ldflags="-s -w" ./cmd/broker
+	go build -i -ldflags="-s -w" ./cmd/migration
 
 build: broker ## Build binary from source
 	@echo > /dev/null
@@ -69,6 +70,7 @@ prep-local: ## Prepares the local dev environment
 
 build-image: ## Build a docker image with the broker binary
 	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/broker ./cmd/broker
+	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/migration ./cmd/migration
 	docker build -f ${BUILD_DIR}/Dockerfile-localdev -t ${BROKER_IMAGE}:${TAG} ${BUILD_DIR}
 	@echo
 	@echo "Remember you need to push your image before calling make deploy"
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,8 @@ vendor: ## Install or update project dependencies
 
 broker: $(SOURCES) ## Build the broker
 	go build -i -ldflags="-s -w" ./cmd/broker
+
+migration: $(SOURCES)
 	go build -i -ldflags="-s -w" ./cmd/migration
 
 build: broker ## Build binary from source
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,9 @@ push:
 
 clean: ## Clean up your working environment
 	@rm -f broker
+	@rm -f migration
 	@rm -f build/broker
+	@rm -f build/migration
 	@rm -f adapters.out apb.out app.out auth.out broker.out coverage-all.out coverage.out handler.out registries.out validation.out
 
 really-clean: clean cleanup-ci ## Really clean up the working environment
--- a/ansible-service-broker.spec
+++ b/ansible-service-broker.spec
@@ -150,6 +150,7 @@ cp -r pkg src/github.com/openshift/ansible-service-broker
 %build
 export GOPATH=$(pwd):%{gopath}
 go build -tags "seccomp selinux" -ldflags "-s -w" ./cmd/broker
+go build -tags "seccomp selinux" -ldflags "-s -w" ./cmd/migration
 
 #Build selinux modules
 # create selinux-friendly version from VR and replace it inplace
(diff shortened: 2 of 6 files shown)


cmd/migration/main.go
```diff
@@ -0,0 +1,209 @@
...
+	clients.InitEtcdConfig(con)
+	logrus.Infof("etcd configuration: %v", con)
+	var err error
+	etcdDao, err = etcd.NewDao()
+	if err != nil {
+		panic(fmt.Sprintf("Unable to connect to etcd - %v", err))
+	}
+	crdDao, err = crd.NewDao(options.MigrationNamespace)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to connect to crd - %v", err))
```
rthallisey: NIT: I don't think this connects to the crd resource.  It only creates the client.   Maybe a different error message should be here.

Makefile
```diff
@@ -19,6 +19,7 @@ vendor: ## Install or update project dependencies
 
 broker: $(SOURCES) ## Build the broker
 	go build -i -ldflags="-s -w" ./cmd/broker
+	go build -i -ldflags="-s -w" ./cmd/migration
```
jmrodri: this doesn't feel right. The `broker` Makefile target is the ACTUAL broker binary. This target says build the `broker` binary if any of the `$(SOURCES)` have changed.

I would expect the following:

```Makefile
migration: $(SOURCES) ## Build the migration tool
    go build -i -ldflags="-s -w" ./cmd/migration
```

Makefile
```diff
@@ -69,6 +70,7 @@ prep-local: ## Prepares the local dev environment
 
 build-image: ## Build a docker image with the broker binary
 	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/broker ./cmd/broker
+	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/migration ./cmd/migration
```
jmrodri: :ok:

ansible-service-broker.spec
```diff
@@ -178,6 +179,7 @@ rm -rf src
 %install
 install -d -p %{buildroot}%{_bindir}
 install -p -m 755 broker %{buildroot}%{_bindir}/asbd
+install -p -m 755 migration %{buildroot}%{_bindir}/migration
```
jmrodri: you need a line under the `%files` section for the migration
```
%{_bindir}/migration
``

Similar to `asbd`
https://github.com/shawn-hurley/ansible-service-broker/blob/11b100561dc09b374a2a76aef5e99a0364e77b8d/ansible-service-broker.spec#L283

File: build/Dockerfile-canary [hunk 5]
```diff
@@ -43,6 +43,7 @@ WORKDIR /go/src/github.com/openshift/ansible-service-broker
 
 RUN go build -i -ldflags="-s -w" ./cmd/broker
 RUN mv broker /usr/bin/asbd
+RUN mv migration /usr/bin/migration
```
jmrodri: :ok_hand:

build/Dockerfile-localdev
```diff
@@ -23,6 +23,7 @@ RUN mkdir /var/log/ansible-service-broker \
 
 COPY entrypoint.sh /usr/bin/
 COPY broker /usr/bin/asbd
+COPY migration /usr/bin/migration
```
jmrodri: :100:

cmd/migration/main.go
```diff
@@ -0,0 +1,209 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
```
jmrodri: stock `flag` is probably okay for this. But it is weird to have a different one than we are using for the broker:

```golang
 "github.com/jessevdk/go-flags"
```

Not a blocker just pointing it out.

cmd/migration/main.go
shawn-hurley: Waiting on more reviews before updating this one

rthallisey: @shawn-hurley might also be worth a precheck on the bundle resource. Maybe something like this: 
```go
_, err := resources.Bundle.Bundles(c.config.Namespace).List(metav1.ListOptions{})
if err != nil {
....
}
```

jmrodri:
The rest looks okay.

cmd/migration/main.go
shawn-hurley: If the bundles already exist it will error and the migration will fail, I think it would be better to attempt to migrate IMO.

jmrodri:
Add migration to the clean target `@rm -f broker migration`. Once that is done, merge it yourself.

PR #879, opened 2018-04-06
Add a table to track release dates

rthallisey:
Add release dates to the README.


shawn-hurley:
👍 on the international date format but I like if we add this!

Issue #818, opened 2018-03-06
Pass provision credentials to the deprovision playbook even if APB is not bindable


maleck13:
## Feature:

Not sure if this is a feature or a bug, but putting it under feature at the moment. 

As part of a provision or binding, we create certain resources within the application (think a realm and user in keycloak for example)
We have a usecase that developers may not want to provision a new instance of a service like keycloak every time. So we allow developers to specify using an existing keycloak as part of the provision. In this case the playbook will not create new pods etc but instead just create the new realm and user and drop the details into the credentials. 
When deprovisioning, we need to access these credentials so that we can tear down the resources we created within the service as in this case the original keycloak server will remain running.

**Solution**
Pass through the any extracted credentials created at provision time to the deprovision action. 

I happy to provide a PR for this.

vchepeli:
When we save fileds for binding/unbind
```
asb_encode_binding:
  fields:
    name: "<some-name>"
    resource: "some-resource"
    uri: "<some-url>"
```
We want also save fields in provision phase and pass them to deprovision as `_apb_provision_creds`.PARAM_NAME
Currently `_apb_provision_creds` is `undefined` when its used in deprovision phase

maleck13:
This can be assigned to me

dymurray:
I am still hitting this issue with the `canary` broker image. I am reopening this issue to track. I can also confirm that folks on the Cloudforms team have tested with `canary` as well and are unable to access `_apb_provision_creds`.

dymurray:
This is actually due to the fact that the broker does not pass these credentials to the broker unless you make `bindable` True in `apb.yml`. We should be passing in the provision creds at all times so I am going to change this issue to reflect it.

dymurray:
I would expect the broker to always run the `GetExtractedCredentials` function (Or maybe `GetBind`?) and if it doesn't find credentials, simply pass an empty object for `_apb_provision_creds` to deprovision.

maleck13:
I will take a look at this

maleck13:
This seems reasonable to pass an empty value but always pass the arg. 👀

maleck13:
Although it does raise the question of whether this should be the case for all arguments. As in we will always pass a certain set of arguments even if there is no value for them

maleck13:
Perhaps something like
```
// EnsureDefaults returns the default set of parameters with default values
func (p Parameters) EnsureDefaults() {
[...]

PR #881, opened 2018-04-09
Fix bug 1564542

maleck13:
This fixes bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1564542

 - add an EnsureDefaults() to the parameters type that sets a default value for _apb_provision_creds var. More could be added in the future.

fixes https://github.com/openshift/ansible-service-broker/issues/818


shawn-hurley:
Thanks for the bug fix @maleck13

PR #884, opened 2018-04-10
Lock version in template

djzager:
Not super important, but this makes it so that the simple-broker-template on the `release-1.1` branch is using the correct container images.


djzager:
Not associating a bug with this as it's just a template change for upstream.

PR #885, opened 2018-04-10
Issue 882 - Simple broker template should use release-1.1 instead of a sprint tag

### LINKED ISSUE #882: Cannot deploy ansible service broker because of missing docker image tag
## Bug:

**What happened**:
Error when deploying ansible-service-broker from simple-broker-template
Caused by non existing docker images with `sprint142` tag:
https://github.com/openshift/ansible-service-broker/blob/master/templates/simple-broker-template.yaml#L149

Broker deployed successfully

**How to reproduce it**:
```
oc new-project ansible-service-broker
curl -s https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/simple-broker-template.yaml | oc process -n "ansible-service-broker" -f - | oc create -f -
```

Tag `latest` fixed the issue

djzager:

Update the images used by the simple-broker-template to be release-1.1 branch.

```
$ docker run -it --entrypoint asbd docker.io/ansibleplaybookbundle/origin-ansible-service-broker:sprint142 --version
1.1.3

$ docker run -it --entrypoint asbd docker.io/ansibleplaybookbundle/origin-ansible-service-broker:release-1.1 --version
1.1.15
```

fixes #882

### DIFF
--- a/templates/simple-broker-template.yaml
+++ b/templates/simple-broker-template.yaml
@@ -146,7 +146,7 @@ objects:
       spec:
         serviceAccount: asb
         containers:
-        - image: ansibleplaybookbundle/origin-ansible-service-broker:sprint142
+        - image: ansibleplaybookbundle/origin-ansible-service-broker:release-1.1
           name: asb
           imagePullPolicy: IfNotPresent
           volumeMounts:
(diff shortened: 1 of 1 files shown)


djzager:
Not associating a bug with this as it's just a template change for upstream.

jmrodri:
@djzager why should it use release-1.1? Also the commit message should be "Issue 882 - Simple broker template should use release-1.1 instead of a sprint tag"

djzager:
I believe it should be using `release-1.1` as the `sprint142` tag 1) is not working 2) `release-1.1` is a tag we would potentially update with fixes.

PR #887, opened 2018-04-12
pass ROUTING_SUFFIX parameter

SaravanaStorageNetwork:
pass ROUTING_SUFFIX parameter to deploy-ansible-service-broker template

ROUTING_SUFFIX is having default value in deploy-ansible-service-broker template.
It may have different value if user wishes.

So, pass the same as parameter.

 - Pass ROUTING_SUFFIX as a parameter to deploy-ansible-service-broker template


djzager:
https://github.com/openshift/ansible-service-broker/pull/939 removes the use of templates (and actually deprecates) `run_latest_build.sh`. I think we should close this PR in favor of #939

rthallisey:
https://github.com/openshift/ansible-service-broker/pull/939

PR #891, opened 2018-04-12
Update scripts to use the broker apb

djzager:
Update scripts that deploy the broker to use the [`automation-broker-apb`](https://github.com/automationbroker/automation-broker-apb).

### DIFF
--- a/scripts/run_latest_build.sh
+++ b/scripts/run_latest_build.sh
@@ -45,6 +45,7 @@ APB_NAME=${APB_NAME:-"automation-broker-apb"}
 APB_IMAGE=${APB_IMAGE:-"docker.io/automationbroker/automation-broker-apb:latest"}
 BROKER_NAME=${BROKER_NAME:-"ansible-service-broker"}
 BROKER_NAMESPACE=${BROKER_NAMESPACE:-"ansible-service-broker"}
+HELM=${HELM:-"false"}
 
 version=$(oc version | head -1)
 client_version=$(echo $version | egrep -o 'v[0-9]+(\.[0-9]+)+' | tr -d v.)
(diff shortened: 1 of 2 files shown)


scripts/run_latest_k8s_build.sh
```diff
@@ -1,58 +1,31 @@
...
-
-    context=$(kubectl config current-context)
-    cluster=$(kubectl config get-contexts $context --no-headers | awk '{ print $3 }')
-
-    kubectl config set-context $context --cluster=$cluster --namespace=ansible-service-broker
-    kubectl create -f "/tmp/k8s-ansible-service-broker.yaml"
-
-    create-broker-resource
+    kubectl create ns $BROKER_NAMESPACE
+    kubectl create serviceaccount $APB_NAME --namespace $BROKER_NAMESPACE
```
shawn-hurley: Can we delete this svcacct after this runs? 

Also could this just run as the user running the script? do we want to do this extra work?

scripts/run_latest_build.sh
```diff
@@ -41,65 +41,66 @@ HOSTNAME=${PUBLIC_IP}.nip.io
...
-
-curl -s $TEMPLATE_URL \
-  | oc process \
-  -n ansible-service-broker \
-  -p DOCKERHUB_ORG="$DOCKERHUB_ORG" \
-  -p ENABLE_BASIC_AUTH="$ENABLE_BASIC_AUTH" \
-  -p NAMESPACE=ansible-service-broker \
-  $VARS -f - | oc create -f -
+# Use the automation-broker-apb to deploy the broker
+oc create serviceaccount $APB_NAME --namespace $BROKER_NAMESPACE
```
shawn-hurley: Same question as below re: do we need the service account

djzager: I actually was smart and delete that stuff in this script but forgot to add that to the k8s version.

Ref: https://github.com/openshift/ansible-service-broker/pull/891/files#diff-34011cced263e7ac1afc542ee01bf933R101

scripts/run_latest_build.sh
```diff
@@ -41,65 +41,65 @@ HOSTNAME=${PUBLIC_IP}.nip.io
...
+    --serviceaccount=$APB_NAME \
+    -- provision -e broker_name=$BROKER_NAME
 if [ "$?" -ne 0 ]; then
-  echo "Error processing template and creating deployment"
+  echo "Error deploying broker"
   exit
 fi
+oc delete pod -n $BROKER_NAMESPACE $APB_NAME
+oc delete clusterrolebinding $APB_NAME
+oc delete serviceaccount $APB_NAME
```
shawn-hurley: I think your going to want to specify the namespace for this deletion as well.

PR #897, opened 2018-04-19
Bug 1569220 - Add dashboard redirector feature

eriknelson:
This PR adds support for async provisioned dashboard urls. This is [currently not possible with the OSB spec](https://github.com/openservicebrokerapi/servicebroker/issues/4980), so our solution is to add a redirect service that runs alongside the broker. The broker immediately returns a URL to the redirector service as a sort of "future" url, that will ultimately redirect to an APB's specified `dashboard_url` location at the conclusion of its provision run.

A few notes:
* This feature requires an APB to be built on top of an `apb-base` that carries the dashboard redirector ansible module.
* The broker author must specify in an `alpha` section of their APB spec that they would like to enable this feature. [Example APB](https://github.com/eriknelson/dashboard-url-apb/blob/master/apb.yml#L14).
* The deployment of the broker must have the route to the dashboard redirector service specified in the broker's configmap. [We have a sensible default configured in the template for this](https://github.com/openshift/ansible-service-broker/blob/feature-add-dashboard-url/templates/deploy-ansible-service-broker.template.yaml#L287).
* This is entirely opt in; there should be no change to existing behavior unless the feature is explicitly opted into by an APB author.

* Update bundle-lib and broker-client-go dependencies to bring in `DashboardURL` and `Alpha` feature gate support.
* Update broker to check if conditions are right to use the feature.
* Returns a `DashboardURL` on initial async provision response that points to the dashboard redirector service if the feature is enabled.
* Adds dashboard-redirector binary and modifies the build files to bake it into the broker image.
* Updates template to deploy the dashboard-redirector alongside the broker.


cmd/dashboard-redirector/main.go
```diff
@@ -0,0 +1,109 @@
...
+		errMsg = fmt.Sprintf("No DashboardURL set for requested instance! %v", id)
+		logrus.Infof("%s, returning 404", errMsg)
+		http.Error(w, errMsg, http.StatusNotFound)
+		return
+	}
+
+	logrus.Infof("DashboardURL found: %s, 301 redirecting", si.DashboardURL)
+	var redirectURL string
+	if !strings.HasPrefix("http", si.DashboardURL) {
+		redirectURL = fmt.Sprintf("http://%s", si.DashboardURL)
```
mhrivnak: What's the use case behind this? Is it possible to just force whatever external entity is supplying the URL to supply a URL with a valid scheme?

Aside from the risk of assuming non-tls http, it's possible a service would have a management dashboard available through other non-http means, such as ssh.

If there is a use case that requires making this assumption, it would be helpful to add an in-line comment about it.

cmd/dashboard-redirector/main.go
```diff
@@ -0,0 +1,109 @@
...
+		errMsg := "Did not find service instance id as a query param!"
+		logrus.Error(errMsg)
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	logrus.Infof("Got request for service instance %s, looking up dashboard_url", id)
+
+	si, err := crdDao.GetServiceInstance(id)
+	if err != nil {
```
mhrivnak: What happens if the service instance doesn't exist? Would that result in `err` having a value here? Or would `si` just have the default zero value?

If it's the former, I think the `StatusInternalServerError` below is incorrect in that case and should be a 404, since the instance would have literally been "not found".

If it's the latter, then the log message "Successfully loaded" would be misleading.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -3,6 +3,23 @@ kind: Template
 metadata:
   name: ansible-service-broker
 objects:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: dr
```
mhrivnak: For someone looking at this Service and wondering what it is, could this be a touch more descriptive? It'll save us from answering the question "What is the 'dr' service?" a hundred times. ;)

cmd/dashboard-redirector/main.go
```diff
@@ -0,0 +1,109 @@
...
+		logrus.Fatal("ListenAndServe: ", err)
+	}
+}
+
+func redirect(w http.ResponseWriter, r *http.Request) {
+	var errMsg string
+	logrus.Info("Checking for form")
+
+	id := r.FormValue("id")
```
mhrivnak: Is there any security risk in handing out the dashboard URL to anyone who knows the instance ID? Or is there some implied authorization happening before the request makes it this far?

cmd/dashboard-redirector/main.go
eriknelson: > What's the use case behind this? Is it possible to just force whatever external entity is supplying the URL to supply a URL with a valid scheme?

When an APB passes back something without a scheme; this is more friendly and allows that case to pass and assume http rather than explicitly fail.

> Aside from the risk of assuming non-tls http

It seems to me the worst case scenario here is that http is assumed when not supported by the service. Many will force a redirect to https.

> it's possible a service would have a management dashboard available through other non-http means, such as ssh.

Interesting point.

I'm anticipating there will be APB authors that unintentionally do not provide a scheme and the intention is to help them with a reasonable assumption. Alternatively, we can lock this down, warn, and default to not providing the dashboard url if the scheme is missing. As I'm thinking about it, the latter might not be a bad idea since the majority of authors will probably get that service URL dynamically from some API and store it in a var to later provide to our module, greater increasing the likelyhood of having the scheme handled by the remote system.

cmd/dashboard-redirector/main.go
eriknelson: Discussed in person, decided to file an issue here to address if/when the feature graduates from alpha.

cmd/dashboard-redirector/main.go
eriknelson: Updated to check if it's actually an is not found error and will return the correct code in that case, otherwise will return an internal error.

eriknelson:
@mhrivnak Think I addressed your requests for change, let me know if I missed anything!

pkg/broker/broker.go
```diff
@@ -655,10 +657,44 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
...
+			return false
+		}
+		if val, ok = spec.Alpha["dashboard_redirect"]; !ok {
+			return false
+		}
+		if dr, ok = val.(bool); !ok {
+			return false
+		}
+		if a.brokerConfig.DashboardRedirector == "" {
+			warnMsg := fmt.Sprintf("Attempting to provision %v", spec.FQName) +
```
shawn-hurley: I personally prefer log.Warningf and then you can remove the string concatenation.

pkg/broker/broker.go
```diff
@@ -655,10 +657,44 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
...
 		}
 	}
-	// TODO: What data needs to be sent back on a response?
-	// Not clear what dashboardURL means in an AnsibleApp context
-	// operation should be the task id from the work_engine
-	return &ProvisionResponse{Operation: token}, nil
+
+	var response *ProvisionResponse
+
+	dashboardRedirectEnabled := func() bool {
```
shawn-hurley: should we make this it's own helper function below and not a function defined in a function? 

Also do we need the dashboard url for [GetServiceInstance](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#fetching-a-service-instance)

pkg/broker/job_state_subscriber.go
```diff
@@ -44,6 +44,22 @@ func (jss *JobStateSubscriber) Notify(msg JobMsg) {
 			return
 		}
 	}
+	//TODO: Need to get the service instance
+	//TODO: NOTE: THIS NEEDS TO BREAK OUT TO OWN SUBSCRIBER
```
shawn-hurley: I think if we have the time, this might be something we should do.

templates/deploy-ansible-service-broker.template.yaml
```diff
@@ -167,6 +184,36 @@ objects:
             secret:
               secretName: asb-auth-secret
 
+- apiVersion: v1
```
shawn-hurley: I think before we merge we should get this working as a container inside the asb pod IMO

cmd/dashboard-redirector/README.md
```diff
@@ -0,0 +1,26 @@
+# Dashboard Redirector
+
+**IMPORTANT** This is an alpha feature, meaning it can change at any time!
+
+Currently, the OSB spec does not allow for the brokers to return a `dashboard_url`
+at the conclusion of an async provision. This is problematic because brokers
+(and bundles, in our case) may not actually know their `dashboard_url` if its
```
mhrivnak: s/its/it's/

cmd/dashboard-redirector/README.md
```diff
@@ -0,0 +1,26 @@
...
+a path that's dynamically determined while the instance is being provisioned.
+
+The issue has been documented in the spec [here](https://github.com/openservicebrokerapi/servicebroker/issues/498).
+
+As a workaround, our broker runs this "dashboard-redirector" as a simple
+service that accepts a service instance ID, and looks up the service instance
+in our DAO. If the `DashboardURL` is present on the service instance, the
+dashboard-redirector will return a 302 redirect to that location. Otherwise,
```
mhrivnak: s/302/301/

cmd/dashboard-redirector/main.go
mhrivnak: :+1: I'd change the value for `errMsg` too just to make it clear that all the software worked correctly, and your instance just doesn't exist. But that's a minor suggestion.

eriknelson: updated

pkg/broker/broker.go
eriknelson: Do you have a preferred alternative to adding the strings? I really can't stand 130 char error messages, and that's all I care about :)

templates/deploy-ansible-service-broker.template.yaml
djzager: My belief is that this is so we don't have to expose a public facing route in order for the broker to be able to talk to it. Is that right?

eriknelson: @djzager that's is true, but the broker in this case never talks to the redirector. It's for external clients (OpenShift UI).

pkg/broker/broker.go
shawn-hurley: >but I'm not sure that there is actually anything calling it?
👍 

>  I prefer inline function defs like this because I'm trying to define what it means to be enabled at a higher level when the conditionals start to get more complex than a simple boolean expression in the conditional. I prefer inline because it's close to the call site.

I personally worry about to many inline functions. I believe that at a certain point they make it harder to follow the function. With that said I this one makes sense if it is all encapsulated in another helper function 👍 from me

dymurray:
Still trying to get the DR container in the same pod as ASB.... routes are not cooperating but will push here when its done.

pkg/broker/broker.go
jmrodri: You could do one of the following, use tick marks:

```golang
	warnMsg := `Attempting to provision %v, which has dashboard redirect enabled,
but no dashboard_redirector route was found in the broker's configmap.
Deploying without a dashboard_url.`

	log.Warningf(warnMsg, spec.FQName)
```
You could keep the concatenation but remove the sprintf using the Warningf version:

```golang
	warnMsg1 := "Attempting to provision %v, which has dashboard redirect enabled" +
		"but no dashboard_redirector route was found in the broker's " +
		"configmap. Deploying without a dashboard_url."

	log.Warningf(warnMsg1, "FQName")
```

Or you could reword the message to be shorter:

```golang

	// short message, no fqname
	log.Warning("Deploying without a dashboard_url, no dashboard_redirector route found")

	// short message with fqname
	log.Warningf("Deploying without a dashboard_url, no dashboard_redirector route found for %v", spec.FQName)
```

cmd/dashboard-redirector/main.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("Unable to create crd client - %v", err))
+	}
+
+	logrus.Info("Trying to load batch specs as a sanity check...")
+	specs, err := crdDao.BatchGetSpecs("")
+	if err != nil {
+		panic(fmt.Sprintf("Sanity check failed! -> %v", err))
+	} else {
+		logrus.Info("Sanity check passed! Loaded specs: %v", specs)
```
jmrodri: What does this print out?

eriknelson: @jmrodri don't the ticks include the newlines?

jmrodri: @eriknelson yeah the tick marks will include the new lines.

eriknelson:
@mhrivnak Think I hit got your second round of comments.

@shawn-hurley I'm not sure I understand why we need another subscriber? Could you elaborate on that?

eriknelson:
@fabianvf pinging you; think these template additions probably need to make their way into openshift-ansible.

fabianvf:
@eriknelson could you break the template changes out to a 3.10 aligned Trello card?

pkg/broker/job_state_subscriber.go
jmrodri: I don't think it needs a different subscriber, but it could definitely use a helper function. It feels weird where it is now.

```golang

func (jss *JobStateSubscriber) handleDashboardURL(msg JobMsg) {
	// Update with dashboard URL.
	if msg.DashboardURL != "" {
		instance, err := jss.dao.GetServiceInstance(msg.InstanceUUID)
		if err != nil {
			log.Errorf("Error after job succeeded : %v", err)
			return
		}
		instance.DashboardURL = msg.DashboardURL
		err = jss.dao.SetServiceInstance(msg.InstanceUUID, instance)
		if err != nil {
			log.Errorf("Error after job succeeded : %v", err)
			return
		}
}
```

The `Notify` would then look cleaner:
```golang
// Notify external API to notify this subscriber of a change in the Job
func (jss *JobStateSubscriber) Notify(msg JobMsg) {
	log.Debugf("JobStateSubscriber Notify : msg state %v ", msg.State)
	id := msg.InstanceUUID
	if isBinding(msg) {
		id = msg.BindingUUID
	}
	if _, err := jss.dao.SetState(id, msg.State); err != nil {
		log.Errorf("Error JobStateSubscriber failed to set state after action %v completed with state %s err: %v", msg.State.Method, msg.State.State, err)
		return
	}
	if msg.State.State == apb.StateSucceeded {
		if err := jss.handleSucceeded(msg); err != nil {
			log.Errorf("Error after job succeeded : %v", err)
			return
		}
	}

	handleDashboardURL(msg)

	return
}
```

dymurray:
https://github.com/openshift/openshift-ansible/pull/8086

cmd/dashboard-redirector/main.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		var errMsg string
+		if crdDao.IsNotFoundError(err) {
+			errMsg = fmt.Sprintf("Requested service instance with id %s not found.", id)
+			http.Error(w, errMsg, http.StatusNotFound)
+		} else {
+			errMsg = fmt.Sprintf("Something went wrong trying to load service instance [%s] -> %s", id, err)
+			http.Error(w, errMsg, http.StatusInternalServerError)
+		}
+		logrus.Errorf(errMsg, id, err.Error())
```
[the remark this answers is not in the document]
eriknelson: Updated!

PR #898, opened 2018-04-19
Bug 1533425 - protect against simultaneous updates

jmrodri:
protect against simultaneous updates

Simultaneous calls to SetState or SetBindingInstance would fail with an already exists error instead of graciously updating the existing instances.

Also, the broker.Bind was incorrectly trying to set the state of a binding job using the instance id as the key, vs use bindingid for state in bindings

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -951,7 +951,7 @@ func (a AnsibleBroker) Bind(instance apb.ServiceInstance, bindingUUID uuid.UUID,
 			return nil, false, err
 		}
 
-		stateKey, err := a.dao.SetState(instance.ID.String(), apb.JobState{
+		stateKey, err := a.dao.SetState(bindingUUID.String(), apb.JobState{
 			Token:  token,
 			State:  apb.StateInProgress,
 			Method: apb.JobMethodBind,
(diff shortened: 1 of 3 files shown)


pkg/dao/crd/dao.go
```diff
@@ -266,11 +280,27 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
 	}
 
+	// Create the actual JobState CRD
 	_, err = d.client.JobStates(d.namespace).Create(&js)
-	if err != nil {
+	if err != nil && apierrors.IsAlreadyExists(err) {
+		// looks like we already have this state, probably created by
+		// another goroutine. Let's try to update the existing one instead.
+		if js, err := d.client.JobStates(d.namespace).Get(state.Token, metav1.GetOptions{}); err == nil {
+			js.Spec = j
```
shawn-hurley: Should this be in a helper function?

pkg/broker/broker.go
```diff
@@ -1073,13 +1074,13 @@ func (a AnsibleBroker) Unbind(
...
 		}
 
-		if _, err := a.dao.SetState(serviceInstance.ID.String(), apb.JobState{
-			Token:  token,
-			State:  apb.StateInProgress,
-			Method: apb.JobMethodUnbind,
-		}); err != nil {
-			log.Errorf("failed to set initial jobstate for %v, %v", token, err.Error())
-		}
+		// if _, err := a.dao.SetState(bindInstance.ID.String(), apb.JobState{
```
shawn-hurley: Just my opinion but if the code is not running can we delete it? 

We can also go back in history with git and re-add it back?

jmrodri:
@shawn-hurley  refactored update code into a method. Let me know.

PR #900, opened 2018-04-19
Add PUSH_IMAGE var to allow image to be pushed

jmrodri:
There are quite a few times I run make build-image ORG=jmrodri and I
immediately call docker push afterwards. This allows me to let the
makefile do it if I want it to.


Makefile
```diff
@@ -2,6 +2,7 @@ REGISTRY         ?= docker.io
 ORG              ?= ansibleplaybookbundle
 TAG              ?= latest
```
djzager: I think if this is something we want then we have to make it more dummy proof. At a minimum I think you must set `TAG` to be something like:

`TAG ?= $(shell git rev-parse short HEAD)`

That will prevent people from accidentally using the makefile to break the "latest" broker.

PR #61, opened 2017-04-04
Endpoint validation

rthallisey:
Validate we are connected to endpoints before we start the broker.

PR #904, opened 2018-04-23
fix statements with formatting

jmrodri:
Making broker consistent with the log statements in bundle-lib, i.e. use
the *f version when logging formatting strings. It is most important
when using logrus, go-logging seems to handle it okay. But this PR will
make it easier to move to logrus in the future.

See PR #61 from bundle-lib for more information:

https://github.com/automationbroker/bundle-lib/pull/61

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -42,7 +42,7 @@ import (
 	"github.com/automationbroker/bundle-lib/registries"
 	agnosticruntime "github.com/automationbroker/bundle-lib/runtime"
 	"github.com/automationbroker/config"
-	"github.com/jessevdk/go-flags"
+	flags "github.com/jessevdk/go-flags"
 	"github.com/openshift/ansible-service-broker/pkg/auth"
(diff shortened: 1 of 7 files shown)


pkg/app/app.go
```diff
@@ -42,7 +42,7 @@ import (
 	"github.com/automationbroker/bundle-lib/registries"
 	agnosticruntime "github.com/automationbroker/bundle-lib/runtime"
 	"github.com/automationbroker/config"
-	"github.com/jessevdk/go-flags"
+	flags "github.com/jessevdk/go-flags"
```
shawn-hurley: Does this not cause issues?

jmrodri: interesting goimports changed this, On an old version of goimports, it completely removed it, and on the other machine it added the above. I'll fix it using gedit :) But no it didn't seem to cause any issues since we're not also importing the system flags.

jmrodri: So goimports won't touch it now that I fixed it. It's weird that `go-flags` is actually used as `flags.`

PR #905, opened 2018-04-24
Build migration into canary image

djzager:
Actually build the migration command before trying to move it.


djzager:
Merging w/o waiting for CI as this is not source related.

PR #906, opened 2018-04-24
Use git clone for canary

djzager:
Turns out that dockerhub doesn't allow you to configure the
build-context (unlike cloud.docker). However, since this is an openshift
repository, it isn't straightforward how to configure automated builds
for cloud.docker. Instead of wasting time trying to figure that out.
Just revert back to the git clone method for now.


djzager:
Again. This doesn't need CI as it's only build related.

PR #907, opened 2018-04-25
Bug 1566924 - renaming crd resources

shawn-hurley:
 - removes job state CRD
 - Moves state to Bundle Instance and Bundle Binding
 -

depends-on <PR>

fixes <issue_number>


shawn-hurley:
@jmrodri @eriknelson This appears to be failing due to config changes between automationbroker-apb and the image created from this PR. Feels like the first time that the APB is hitting this problem.

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+	// get the binding based on instance ID //update the job based on the token.
+	var job v1.Job
+	bi, err := d.client.BundleBindings(d.namespace).Get(id, metav1.GetOptions{})
+	if err != nil && !d.IsNotFoundError(err) {
+		log.Errorf("Unable to update the job state: %v - %v", token, err)
+		return apb.JobState{}, fmt.Errorf("unable to find job state %v", token)
+	} else if d.IsNotFoundError(err) {
+		si, err := d.client.BundleInstances(d.namespace).Get(id, metav1.GetOptions{})
+		if err != nil {
+			log.Errorf("Unable to update the job state: %v - %v", token, err)
```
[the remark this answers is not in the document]
eriknelson: Also, err != nil and si.Status.Jobs branches here look the same? Could combine them?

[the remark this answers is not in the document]
eriknelson: Is this the BundleInstance not found case?

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+		Error:       job.Error,
+	}, nil
 }
 
 // GetStateByKey - Retrieve a job state from the kvp API for a job key
 func (d *Dao) GetStateByKey(key string) (apb.JobState, error) {
-	return d.GetState("", key)
+	bi, err := d.client.BundleBindings(d.namespace).Get(key, metav1.GetOptions{})
+	if err != nil {
+		log.Errorf("Unable to update the job state: %v - %v", key, err)
```
eriknelson: Logs should read not found?

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+		Token:       token,
+		State:       crd.ConvertStateToAPB(job.State),
+		Error:       job.Error,
+	}, nil
 }
 
 // GetStateByKey - Retrieve a job state from the kvp API for a job key
 func (d *Dao) GetStateByKey(key string) (apb.JobState, error) {
-	return d.GetState("", key)
+	bi, err := d.client.BundleBindings(d.namespace).Get(key, metav1.GetOptions{})
```
eriknelson: Why does `GetStateByKey` only care about bindings? The usage in the broker only cares about that case?

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+	if err != nil {
+		log.Errorf("Unable to update the job state: %v - %v", key, err)
+		return apb.JobState{}, err
+	}
+	for token, j := range bi.Status.Jobs {
+		// Assuming a single bind job happens per binding instance.
```
eriknelson: :+1: seems like a fair assumption.

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+	}
+	return apb.JobState{}, &apierrors.StatusError{ErrStatus: metav1.Status{
+		Status: metav1.StatusFailure,
+		Code:   http.StatusNotFound,
+		Reason: metav1.StatusReasonNotFound,
+	}}
 }
 
 // FindJobStateByState - Retrieve all the jobs that match the specified state
 func (d *Dao) FindJobStateByState(state apb.State) ([]apb.RecoverStatus, error) {
```
eriknelson: :+1: looks good.

pkg/dao/crd/dao.go
```diff
@@ -311,78 +315,163 @@ func (d *Dao) SetState(instanceID string, state apb.JobState) (string, error) {
...
+				}})
+			}
+		}
 	}
-	return rs, nil
+	return rss, nil
 }
 
 // GetSvcInstJobsByState - Lookup all jobs of a given state for a specific instance
 func (d *Dao) GetSvcInstJobsByState(ID string, state apb.State) ([]apb.JobState, error) {
```
eriknelson: It's a little confusing seeing a "get service instance jobs by state" that also is looking for bindings? Does this imply that I can pass this a binding ID and it will give me back relevant jobs?

pkg/dao/crd/dao.go
shawn-hurley: Yes that is the case, the only thing that has a key is the binding as well. apb.BindIntance

pkg/dao/crd/dao.go
shawn-hurley: Yes that is the case. I am making the functionality equivalent here but I agree with your point. I think this interface could use a second pair of eyes.

PR #908, opened 2018-04-26
[wip] Use run_latest_build with existing clusters

rthallisey:
Remove the check on a failed oc cluster up so that folks can run this script on a pre-existing cluster.

### DIFF
--- a/scripts/run_latest_build.sh
+++ b/scripts/run_latest_build.sh
@@ -70,10 +70,7 @@ else
         --public-hostname=${HOSTNAME}
 fi
 
-if [ "$?" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


rthallisey:
marking as WIP since there's some discussion around this on the ML.

djzager:
https://github.com/openshift/ansible-service-broker/pull/939 removes the use of templates (and actually deprecates) `run_latest_build.sh`. I think we should close this PR in favor of adding the broker, and associated automation, to the project.

rthallisey:
thanks @djzager. Closing in favor of https://github.com/openshift/ansible-service-broker/pull/939

PR #909, opened 2018-04-26
Ensure initial state set when job started. Reduce repition in job definitions. Allow jobs to be mocked out at broker level

maleck13:
I did not create an issue for this change. If one is needed that is no problem. Also as did this out of scope work, if it is felt that it is not needed then I happy to remove the PR.  The things here that is likely important is the initialization of the state. Also I know there is a freeze right now so not expecting this to land immediately

While working on the state changes. I was looking into the jobs and work engine. This pr creates a mockable interface laying some of the ground work to allow us to potentially create e2e tests that run against the API ensuring the correct responses without the need to actually spin up APBS but still execute the key paths in the code. It also reduces duplication on in the jobs code.

 -  consolidate the job types and reduce the need for separate method definitions on each job type 
 - Add an interface for getting work in the broker ```WorkFactory```
 - Ensure to set the initial job state before handing control back to the caller to avoid the state not being set and possible race condition (more than one job being created and service catalog asking for last operation update before state has been set).
- generate a mock dao


pkg/app/app.go
```diff
@@ -282,8 +282,11 @@ func CreateApp(args Args, regs []registries.Registry) App {
 		KeepNamespaceOnError: app.config.GetBool("openshift.keep_namespace_on_error"),
 	}
 	apb.InitializeClusterConfig(clusterConfig)
+	brokerNS := app.config.GetString("openshift.namespace")
+	// initialize the work factory
+	workFactory := broker.NewWorkFactory()
```
jmrodri: I'm not sure of the need for the `WorkFactory`.

pkg/broker/jobs.go
```diff
@@ -48,6 +48,17 @@ type apbJob struct {
...
+	if (j.method == apb.JobMethodUnbind || j.method == apb.JobMethodBind) && j.bindingID != nil {
+		return *j.bindingID
+	}
+	return j.serviceInstanceID
+}
+
+func (j *apbJob) Method() apb.JobMethod {
+	return j.method
+}
+
```
jmrodri: `ID` and `Method` are good additions.

pkg/broker/jobs.go
```diff
@@ -138,120 +149,134 @@ func (j *apbJob) createJobMsg(
...
-		serviceInstanceID:      j.serviceInstance.ID.String(),
-		specID:                 j.serviceInstance.Spec.ID,
-		method:                 apb.JobMethodProvision,
-		metricsJobStartHook:    metrics.ProvisionJobStarted,
-		metricsJobFinishedHook: metrics.ProvisionJobFinished,
-		skipExecution:          false,
-		run: func(exec apb.Executor) <-chan apb.StatusMessage {
-			return exec.Provision(j.serviceInstance)
+// ProvisionJob will setup a Work implementation that will perform the provision work
+func (wf *workFactory) ProvisionJob(si *apb.ServiceInstance) Work {
```
jmrodri: I'm not a fan of using the old struct name as a function now. It feels confusing to see something that looks like a "struct" vs something that is a method. For example when I see `ProvisionJob` I'd expect to see `{}` after it like: `ProvisionJob{id: "foo", blah blah blah}`.

If we're going to keep these I prefer making them look like constructor methods: `NewProvisionJob`, `NewBindJob`, etc.

pkg/broker/work_engine.go
```diff
@@ -77,7 +87,14 @@ func waitForNotify(ctx context.Context, sub WorkSubscriber, msg JobMsg, signal c
...
 
-func (engine *WorkEngine) startJob(token string, work Work, topic WorkTopic) {
+func (engine *WorkEngine) setupJob(token string, work Work) error {
+	if _, err := engine.dao.SetState(work.ID(), apb.JobState{Token: token, State: apb.StateNotYetStarted, Method: work.Method()}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (engine *WorkEngine) runJob(token string, work Work, topic WorkTopic) {
```
jmrodri: +1 to `setupJob` and `runJob` vs `startJob`.

pkg/broker/jobs.go
maleck13: Sure happy to change these

pkg/app/app.go
maleck13: The intention here was creating an interface that would encapsulate the creation of work and allow a mock to be used later down the line to do tests against a running broker without it needing to actually start pods etc. Initializing the stuct directly in the code makes it difficult to control the behavior. Really just saw it as a step in the direction of a more unit testable broker.

maleck13:
@jmrodri changes made as requested

PR #913, opened 2018-04-30
Fix RBAC and CRDs in k8s template

pst:

This PR fixes two issues with the K8s template:
 * it adds the `jobstates` resource to the `asb-auth` ClusterRole
 * it fixes the copy&paste error in the `bundlebindings` and `bundleinstances` CRDs

### DIFF
--- a/templates/k8s-ansible-service-broker.yaml.j2
+++ b/templates/k8s-ansible-service-broker.yaml.j2
@@ -29,7 +29,7 @@ rules:
   verbs: ["create", "delete"]
 - apiGroups: ["automationbroker.io"]
   attributeRestrictions: null
-  resources: ["bundles", "bundlebindings", "bundleinstances"]
+  resources: ["bundles", "bundlebindings", "bundleinstances", "jobstates"]
   verbs: ["*"]
 
 ---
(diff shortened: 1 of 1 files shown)


templates/k8s-ansible-service-broker.yaml.j2
```diff
@@ -29,7 +29,7 @@ rules:
   verbs: ["create", "delete"]
 - apiGroups: ["automationbroker.io"]
   attributeRestrictions: null
-  resources: ["bundles", "bundlebindings", "bundleinstances"]
+  resources: ["bundles", "bundlebindings", "bundleinstances", "jobstates"]
```
shawn-hurley: If your using bundle binding/bundle instances then job states should not be installed in the cluster. I think we should remove this IMO.

templates/k8s-ansible-service-broker.yaml.j2
```diff
@@ -257,9 +257,9 @@ spec:
   version: v1alpha1
   scope: Namespaced
   names:
-    plural: servicebindings
-    singular: servicebinding
-    kind: ServiceBinding
+    plural: bundlebindings
```
shawn-hurley: 👍 thanks for fixing this!

djzager:
@pst There is one change that I'm aware needs to be made, removing `jobstates`. Once that is made we should be able to merge this.

templates/k8s-ansible-service-broker.yaml.j2
jmrodri: agreed let's remove the jobstates.

PR #915, opened 2018-04-30
Lock helm version for ci

djzager:
Make helm useful for travis until helm latest is functional.


jmrodri:
Never assume an ack :laughing:

PR #916, opened 2018-05-01
Create broker-apb dir, add install yaml

djzager:
Originally, I had intended to bring in the entire broker-apb into this project so that I would be able to finish the versioning strategy blog post. All that really needs to exist for the blog post is the `install.yaml`. This PR adds that file where the broker-apb will live in the future without having to do all the work up front of moving the project, updating travis, and updating docker build automation.

### DIFF
--- a/broker-apb/install.yaml
+++ b/broker-apb/install.yaml
@@ -1,0 +1,43 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: automation-broker
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: automation-broker-apb
+  namespace: automation-broker
+
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: automation-broker-apb
+roleRef:
+  name: cluster-admin
+  kind: ClusterRole
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: automation-broker-apb
+  namespace: automation-broker
+
+---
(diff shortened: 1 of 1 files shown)


djzager:
Assuming that this is an acceptable strategy. I'll be adding similar files to the `release-1.0` and `release-1.1` branches.

eriknelson:
LGTM to me too; so this is a set of k8s objects and not an openshift template? Also, do we want a branded "ansible-broker" version? I guess I'm wondering...what do we call ourselves upstream...

djzager:
> LGTM to me too; so this is a set of k8s objects and not an openshift template?

Set of k8s objects that can just be used...`kubectl create -f install.yaml` to get a broker.

> Also, do we want a branded "ansible-broker" version?

I can update the yaml to make everything `ansible-service-broker` (except the APB image) if that helps.

> I guess I'm wondering...what do we call ourselves upstream, because we're in a weird spot right now where our home is openshift/ansible-service-broker, and now we're starting to use yet-another-name. @jmrodri thoughts?

I think this is a fair point, in [Jason's PR to add the broker to origin](https://github.com/jmontleon/origin/blob/c02076b77508cd75198aed91e7d96449ea2476a4/install/automationservicebroker/deploy-automation-broker-apb.yaml#L39) you will see `automation-service-broker`.

djzager:
@jmrodri My feeling is the best way, and the way that would make the template go away, would be for the whole broker-apb to be brought into this project. That would allow us to update the Makefile to 1) build the broker 2) build the broker-apb 3) provision with the broker-apb. I'm not certain that we would rely on this `install.yaml` for that, but that's how I believe we would deploy the broker in development.

Do we still intend to run the broker locally?

Issue #912, opened 2018-04-30
Incorrect version check in run_latest_build.sh


beekhof:
## Bug:

**What happened**:

# ./run_latest_build.sh
++ oc version
++ head -1
+ version='oc v3.7.23'
++ echo oc v3.7.23
++ egrep -o 'v[0-9]+(\.[0-9]+)+'
++ tr -d v.
+ client_version=3723
+ '[' latest '!=' latest ']'
+ ((  3723 >= 3100  ))
+ oc cluster up --image=docker.io/openshift/origin --tag=latest --enable=service-catalog,template-service-broker,router,registry,web-console --routing-suffix=172.16.1.70.nip.io --public-hostname=172.16.1.70.nip.io
Error: unknown flag: --tag

[...]
+ '[' 1 -ne 0 ']'
+ echo 'Error starting cluster'
Error starting cluster
+ exit

**What you expected to happen**:

Script uses the correct flags

**How to reproduce it**:

Run ./run_latest_build.sh with oc v3.7.23

jmrodri:
@beekhof  thanks we'll look at this one soon. For now, you could update the version check to be 3723 to let you run with the 3.7.23 until we have a patch for this script.

jmrodri:
The current rpmvercmp.c file
https://github.com/rpm-software-management/rpm/blob/master/lib/rpmvercmp.c

jmrodri:
I think in order to make this fool proof we're going to have to parse the string field by field. Otherwise, we're always going to run into a problem.

djzager:
I believe the _best_ resolution to this is for us to simply remove the `oc cluster up` from the `run_latest_build`.

PR #920, opened 2018-05-01
Merge run_latest_k8s_build into run_latest_build script

djzager:
Fixes: #912 #931


scripts/run_latest_build.sh
```diff
@@ -1,93 +1,49 @@
...
+echo "========================================================================"
+echo ""
+echo " This script expects a running Kubernetes|OpenShift cluster and a service-catalog."
+echo ""
+echo " OpenShift:"
+echo "   Setup OpenShift: https://github.com/openshift/origin/#installation"
+echo "   You must also enable the service-catalog. This depends on the installed origin"
+echo "   client version."
+echo "     Origin client version < 3.10 \`--service-catalog=true\`."
+echo "     Origin client version >= 3.10 \`--enable=service-catalog\`."
```
shawn-hurley: Question: Do we need PVs installed as well, will it work if you only have just service-catalog?

djzager: I'm going to try this on an oc cluster up without pvs to check. My initial reaction is no.

djzager:
Closing in favor of #939

PR #922, opened 2018-05-02
Add install.yaml for release-1.1 broker

djzager:
Simply add the install.yaml file to a known location in release branch.


shawn-hurley:
This looks to be failing CI for k8s because of helm, and of course, catasb/oc is failing for openshift. 
I don't think we should require these to be fixed for this PR to go in IMO

PR #895, opened 2018-04-17
Bug 1565545 - Adding migration of extracted credentials

shawn-hurley:
#### Describe what this PR does and why we need it:
Currently we do not migrate extracted credentials to the new mode of saving/retrieving/updating/deleting credentials.

#### Changes proposed in this pull request
 - add to migration ability to get all extracted creds for a service instance, and binding
 - save the extracted creds as a secret using the helper methods
 
#### Which issue this PR fixes (This will close that issue when PR gets merged)
https://bugzilla.redhat.com/show_bug.cgi?id=1565545

PR #925, opened 2018-05-03
Remove dashboard-redirector binary

PhilipGough:
Remove binary that was mistakingly added in #895 

cc @eriknelson


eriknelson:
Thanks Philip!

PR #926, opened 2018-05-03
Support building a debuggable image for development

PhilipGough:
Change to help with the development process from a contributor point of view. Allows building of a remotely debuggable image so a debugger can be attached when needed.

Feel free to close this if it is seen to have no value but I found it useful when working on recent issues so thought I would share.

On a side note the docs to run a local broker did not work for me on OSx on either `minishift` or `oc cluster up` but perhaps I'm missing something (broker started but errors out).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ SVC_ACCT_DIR     := /var/run/secrets/kubernetes.io/serviceaccount
 KUBERNETES_FILES := $(addprefix $(SVC_ACCT_DIR)/,ca.crt token tls.crt tls.key)
 COVERAGE_SVC     := travis-ci
 .DEFAULT_GOAL    := build
+ASB_DEBUG_PORT   := 9000
 
 vendor: ## Install or update project dependencies
 	@dep ensure
@@ -76,16 +77,20 @@ prep-local: ## Prepares the local dev environment
 	@./scripts/prep_local_devel_env.sh
 
 build-dev: ## Build a docker image with the broker binary for development
-	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/broker ./cmd/broker
+	docker run --rm -v ${BUILD_DIR}:/tmp/artifact docker.io/philipgough/dlv:centos cp /go/bin/dlv /tmp/artifact
+	env GOOS=linux go build -i -gcflags="-N -l" -o ${BUILD_DIR}/broker ./cmd/broker
 	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/migration ./cmd/migration
 	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/dashboard-redirector ./cmd/dashboard-redirector
-	docker build -f ${BUILD_DIR}/Dockerfile-localdev -t ${BROKER_IMAGE} ${BUILD_DIR}
-	@echo
-	@echo "Remember you need to push your image before calling make deploy"
+	docker build -f ${BUILD_DIR}/Dockerfile-localdev -t ${BROKER_IMAGE} ${BUILD_DIR} --build-arg DEBUG_PORT=${ASB_DEBUG_PORT}
+	@echo ""
+	@echo "Remember you need to push your image before calling make deploy or updating deployment config"
 	@echo "    docker push ${BROKER_IMAGE}"
+	@echo ""
+	@echo "To remotely debug the container, update the deployment config and run the following before connecting the debugger"
+	@echo "    ./scripts/prep_debug_env.sh ${ASB_DEBUG_PORT} <broker-namespace> <broker-deployment-name> "
 
 build-image: ## Build the broker (from canary)
-	docker build -f ${BUILD_DIR}/Dockerfile-canary --build-arg VERSION=${TAG} -t ${BROKER_IMAGE} .
+	docker build -f ${BUILD_DIR}/Dockerfile-canary --build-arg VERSION=${TAG} --build-arg DEBUG_PORT=${ASB_DEBUG_PORT} -t ${BROKER_IMAGE} .
 
 build-apb: ## Build the broker apb
 ifeq ($(TAG),canary)
--- a/scripts/clean_debug_env.sh
+++ b/scripts/clean_debug_env.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+BROKER_NAMESPACE=$1
+PROJECT=${BROKER_NAMESPACE:-"automation-broker"}
+BROKER_DEPLOYMENT=$2
+DEPLOYMENT_NAME=${BROKER_DEPLOYMENT:-"automation-broker"}
+
+oc rollout pause dc/${DEPLOYMENT_NAME} -n ${PROJECT}
+oc env dc/${DEPLOYMENT_NAME} DEBUG_ENABLED=False --overwrite=true -n ${PROJECT}
+oc set probe dc/${DEPLOYMENT_NAME} --readiness --get-url=https://:1338/healthz \
+    --initial-delay-seconds=15 --success-threshold=1 --timeout-seconds=1 --period-seconds=10 --failure-threshold=3 \
+    --liveness --get-url=https://:1338/healthz \
+    --initial-delay-seconds=15 --success-threshold=1 --timeout-seconds=1 --period-seconds=10 --failure-threshold=3 \
+    -n ${PROJECT}
+oc patch -n ${PROJECT} svc ${DEPLOYMENT_NAME} --type=json -p='[
+    {"op": "test",
+     "path": "/spec/ports/0/name",
+     "value": "debug"
+    },
+    {"op": "remove",
+     "path": "/spec/ports/0"
+    }
+]'
+oc rollout resume dc/${DEPLOYMENT_NAME} -n ${PROJECT}
--- a/scripts/prep_debug_env.sh
+++ b/scripts/prep_debug_env.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+DEBUG_PORT=$1
+BROKER_NAMESPACE=$2
+BROKER_DEPLOYMENT=$3
+PROJECT=${BROKER_NAMESPACE:-"automation-broker"}
+DEPLOYMENT_NAME=${BROKER_DEPLOYMENT:-"automation-broker"}
+
+oc rollout pause dc/${DEPLOYMENT_NAME} -n ${PROJECT}
+oc set probe dc/${DEPLOYMENT_NAME} --remove --readiness --liveness -n ${PROJECT}
+oc env dc/${DEPLOYMENT_NAME} DEBUG_ENABLED=True --overwrite=true -n ${PROJECT}
+oc rollout resume dc/${DEPLOYMENT_NAME} -n ${PROJECT}
+oc patch svc ${DEPLOYMENT_NAME} --patch='{"spec":{"ports":[{"name":"debug", "port":'${DEBUG_PORT}',"targetPort":'${DEBUG_PORT}}']}}' -n ${PROJECT}
+sleep 5
+POD_NAME="$(oc get po -o jsonpath='{.items[?(@.metadata.labels.service=="'${DEPLOYMENT_NAME}'")].metadata.name}')"
+until oc get po ${POD_NAME} -o jsonpath='{.status.containerStatuses[0].ready}' | grep "true" >/dev/null 2>&1 ; do sleep 1; done
+
+oc port-forward ${POD_NAME} ${DEBUG_PORT} -n ${PROJECT}
+
+echo "To revert these changes, run ./scripts/clean_debug_env.sh ${PROJECT} ${DEPLOYMENT_NAME}"
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ build/broker
 build/migration
 /dashboard-redirector
 build/dashboard-redirector
+build/dlv
 
 # python artifacts
 .python-version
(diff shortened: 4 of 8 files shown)


jmrodri:
AWESOME! Thanks for the contribution.

build/entrypoint.sh
```diff
@@ -16,4 +16,8 @@ if [ ! -f "$BROKER_CONFIG" ] ; then
 fi
 echo "Using config file mounted to $BROKER_CONFIG"
 
-exec asbd -c $BROKER_CONFIG $FLAGS
+if [ ${DEBUG_ENABLED:-False} == "True" ]; then
+    dlv  --listen=0.0.0.0:${DEBUG_PORT} --headless=true --api-version=2 --log=true exec asbd -- -c $BROKER_CONFIG $FLAGS
```
djzager: Could you skip modifying the entrypoint and simply add this to the debug Dockerfile?

rthallisey:
@PhilipGough can you add some docs around this?

build/entrypoint.sh
djzager: Adding this logic to the entrypoint production broker's use (including downstream) scares me.

build/Dockerfile-debug
```diff
@@ -0,0 +1,49 @@
+FROM philipgough/go-centos:1.10.1 AS build-env
```
djzager: Is there a library image we can use here?

djzager: My initial thought with this is that it should mirror the existing build images as best as possible. Even better if you could modify the `Dockerfile-canary` with build arguments to optionally build it for debugging.

Makefile
```diff
@@ -104,6 +118,9 @@ clean: ## Clean up your working environment
 really-clean: clean cleanup-ci ## Really clean up the working environment
 	@rm -f $(KUBERNETES_FILES)
 
+clean-debug: ## Revert changes made to support debug image
```
djzager: This is a big IMO, so other reviewers can downvote me if I'm wrong. I don't believe that we need this make target (or the associated script), with the move to a broker-apb. My reason for this is that, with our move to a broker-apb, you could simply deprovision the broker and start fresh. I also normally view `clean` make targets to be cleaning up stuff on the host...I'm not super concerned about the cluster.

PhilipGough:
@rthallisey Yep, will do just wanted to see if it was valid first

build/Dockerfile-debug
djzager: This is the first time I've seen a Dockerfile like this :+1:

scripts/prep_debug_env.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+DEBUG_PORT=$1
+PROJECT=${ASB_PROJECT:="ansible-service-broker"}
+
+oc set probe dc/asb --remove --readiness --liveness -n ${PROJECT}
```
djzager: We have the https://github.com/automationbroker/automation-broker-apb maybe this is something that could be added there (ie. you provision a debuggable broker).

build/entrypoint.sh
PhilipGough: Yes can do I wasn't sure if this entrypoint.sh was used in production rpm

build/Dockerfile-debug
PhilipGough: > Is there a library image we can use here?

Couldn't find one

> Even better if you could modify the Dockerfile-canary with build arguments to optionally build it for debugging.

That could be done but not without the modified entrypoint script right?

Makefile
PhilipGough: I'm fine with that, really just wanted this t get state back to where it was in case there were any concerns in that regard.

build/Dockerfile-debug
PhilipGough: The main problem here is what we would need to have go installed in the broker base, which is currently just based off centos to achieve this, where as at the moment we build the binaries and copy them in. I think anything other than a build container is problematic if that makes sense?

scripts/prep_debug_env.sh
PhilipGough: Sure I can take a look there also

build/Dockerfile-debug
djzager: My primary reason for starting this conversation is, if we can avoid having another Dockerfile, that would best. Both the `Dockerfile-latest` and `Dockerfile-nightly` use rpms so we can't do anything there. `Dockerfile-localdev` does the copying but `Dockerfile-canary`, as you can see [here](https://github.com/openshift/ansible-service-broker/blob/master/build/Dockerfile-canary#L44) builds the broker from source.

I would be curious: 

1. how many contributors are using the localdev variant, if that's even well supported anymore I don't know.
1. if we could simply build `dlv` in the `Dockerfile-canary` and update the broker-apb to deploy the broker with the appropriate `command` and `args` in the deployment(config) for debugging.

Makefile
```diff
@@ -88,6 +89,19 @@ release-image:
 	@echo "Remember you need to push your image before calling make deploy"
 	@echo "    make push"
 
+debug-image: ## Build a docker image with the broker binary started in debug mode
+	env GOOS=linux go build -i -gcflags='-N -l' -o ${BUILD_DIR}/broker ./cmd/broker
+	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/migration ./cmd/migration
+	env GOOS=linux go build -i -ldflags="-s -s" -o ${BUILD_DIR}/dashboard-redirector ./cmd/dashboard-redirector
+	docker build -f ${BUILD_DIR}/Dockerfile-debug -t ${BROKER_IMAGE}:${TAG} ${BUILD_DIR} --build-arg DEBUG_PORT=${ASB_DEBUG_PORT}
+	@./scripts/prep_debug_env.sh ${ASB_DEBUG_PORT}
```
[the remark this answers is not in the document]
jmrodri: For now, I'd probably change this to echo the command they should run:

```
@echo "Please run scripts/prep_debug_env.sh ${ASB_DEBUG_PORT}"
```

Or something along those lines.

build/Dockerfile-debug
```diff
@@ -0,0 +1,49 @@
...
+
+RUN git clone https://github.com/derekparker/delve.git $GOPATH/src/github.com/derekparker/delve && \
+    cd $GOPATH/src/github.com/derekparker/delve && \
+    make install
```
jmrodri: I would do a `go get -u github.com/derekparker/delve/cmd/dlv`

build/entrypoint.sh
```diff
@@ -16,4 +16,8 @@ if [ ! -f "$BROKER_CONFIG" ] ; then
...
+else
+    exec asbd -c $BROKER_CONFIG $FLAGS
+fi
```
jmrodri: entrypoint looks good.

scripts/clean_debug_env.sh
```diff
@@ -0,0 +1,19 @@
...
+
+oc patch -n ${PROJECT} svc asb --type=json -p='[
+    {"op": "test",
+     "path": "/spec/ports/0/name",
+     "value": "debug"
+    },
+    {"op": "remove",
+     "path": "/spec/ports/0"
+    }
+]'
```
jmrodri: clean_debug_env script looks good.

scripts/prep_debug_env.sh
```diff
@@ -0,0 +1,7 @@
...
+oc patch svc asb --patch='{"spec":{"ports":[{"name":"debug", "port":'${DEBUG_PORT}',"targetPort":'${DEBUG_PORT}}']}}'
```
jmrodri: prep_debug_env looks good as well.

Makefile
```diff
@@ -88,6 +89,19 @@ release-image:
...
+	@echo
+	@echo "Remember you need to push your image before updating deployment config"
+	@echo "    docker push ${BROKER_IMAGE}:${TAG}"
```
jmrodri: add an `@echo ""` above the docker push line

Makefile
```diff
@@ -88,6 +89,19 @@ release-image:
...
+	@echo
+	@echo "Run the following before connecting the debugger once the container is running"
+	@echo "oc port-forward <oc get po -o jsonpath='{.items[?(@.metadata.labels.service=="asb")].metadata.name}')> ${ASB_DEBUG_PORT}"
```
jmrodri: add an `@echo ""` above the `oc port-forward` line also add 4 spaces in front of the `oc` command so that the output lines up with the `docker push` from above.

The current output looks like this:
```
Remember you need to push your image before updating deployment config
    docker push docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest

Run the following before connecting the debugger once the container is running
oc port-forward <oc get po -o jsonpath='{.items[?(@.metadata.labels.service==asb)].metadata.name}')> 9000
```

I believe this would look better, as it makes the commands more prominent:

```
Remember you need to push your image before updating deployment config

    docker push docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest

Run the following before connecting the debugger once the container is running

    oc port-forward <oc get po -o jsonpath='{.items[?(@.metadata.labels.service==asb)].metadata.name}')> 9000
```

build/Dockerfile-debug
jmrodri: @djzager @PhilipGough I use the `Dockerfile-localdev` ALL THE TIME :) my workflow recently has been to do a `make build-image ORG=jmrodri TAG=demo` and pushing that to my org, then redeploying with catasb. I would be okay movign the debug stuff to the localdev dockerfile

Makefile
```diff
@@ -88,6 +89,19 @@ release-image:
 	@echo "Remember you need to push your image before calling make deploy"
 	@echo "    make push"
 
+debug-image: ## Build a docker image with the broker binary started in debug mode
+	docker run --rm -v ${BUILD_DIR}:/tmp/artifact docker.io/philipgough/dlv:centos cp /go/bin/dlv /tmp/artifact
```
PhilipGough: I'm happy to maintain this image. This is a workaround to support those who cannot do muti stage builds

PhilipGough:
@jmrodri I've made some changes here which fix the issues discussed. If you are happy with these changes I would like to move the debug work into the the `make build-image` target and debugging would be enabled in a single image by updating an env var

eriknelson:
I feel this is ready to merge, but I'm not going to offer any support if it breaks.

PhilipGough:
@eriknelson 

> I feel this is ready to merge, but I'm not going to offer any support if it breaks.

I'd like to get @jmrodri approval on this one because I want to put this stuff into the local image and reduce the overall maintenance. Debugging will be switched on with a simple env var setting at that point. If not we can merge as is.

Makefile
djzager: Is there a reason you can't `go get` it?

PhilipGough:
@rthallisey Sure, I'll land docs with this as soon as its finalised

@djzager > Is there a reason you can't go get it?

The reason I have taken this approach is because it is faster. A `go get` would need to be called on each image build otherwise, slowing down the development cycle.

djzager:
@PhilipGough wouldn't the `go get` really only the first time `make debug-image` was called and just be there on subsequent calls?

PhilipGough:
@djzager Hey, not sure I'm totally following. I have baked the delve binary into the image here and am copying from the container to local disk via a `docker run` command. If the docker run command does a `go get` instead on a centos image that has go installed, I think that will run each time and don't see anything being cached? Maybe I am missing something or misunderstanding. 

The only alternative I see is to maintain an additional Dockerfile on this repo and have a build done from that which would then be cached but I don't like that approach as much. I have done a `go get` to build this image referenced in the make file.

I see the approach I've taken as a workaround essentially for those that cannot do multi stage Docker builds.

scripts/prep_debug_env.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+DEBUG_PORT=$1
+PROJECT=${ASB_PROJECT:="ansible-service-broker"}
+
+oc rollout pause dc/asb -n ${PROJECT}
```
djzager: The broker's deployment config, when deployed via the broker-apb, will not be named `asb`. In https://github.com/openshift/ansible-service-broker/pull/939 every script that I could find that will deploy the broker will use the apb and https://github.com/fusor/catasb/pull/242 changes catasb to enable the broker (which also uses an apb).

djzager:
> Hey, not sure I'm totally following. I have baked the delve binary into the image here and am copying from the container to local disk via a docker run command. If the docker run command does a go get instead on a centos image that has go installed, I think that will run each time and don't see anything being cached? Maybe I am missing something or misunderstanding.

Here is how I changed the `Dockerfile-canary`:

```diff
--- a/build/Dockerfile-canary
+++ b/build/Dockerfile-canary
@@ -41,6 +41,7 @@ RUN mkdir -p ${BROKER_PATH}
 RUN git clone -b $VERSION https://github.com/openshift/ansible-service-broker ${BROKER_PATH}
 WORKDIR ${BROKER_PATH}

+RUN go get -u github.com/derekparker/delve/cmd/dlv
 RUN go build -i -ldflags="-s -w" ./cmd/broker && mv broker /usr/bin/asbd
 RUN go build -i -ldflags="-s -w" ./cmd/migration && mv migration /usr/bin/migration
```

On the second run of `docker build -t broker:canary -f build/Dockerfile-canary build`:

```
Step 18/26 : RUN go get -u github.com/derekparker/delve/cmd/dlv
 ---> Using cache
 ---> a25234b25eca
```

PhilipGough:
@djzager Ah ok, so we were talking about different things :) I intended to merge this work into the local dev image only which does not have golang env available, so the command in the make file makes sense in that respect.

But yeah, if you guys are happy to get this stuff into the canary image also then I'll make the required changes there tomorrow, update the docs and hopefully wrap this up at that point :)

djzager:
> But yeah, if you guys are happy to get this stuff into the canary image also then I'll make the required changes there tomorrow, update the docs and hopefully wrap this up at that point :)

I would like to get this wrapped up too. Sorry to be slowing it down.

PR #928, opened 2018-05-04
Add parsing of param dependencies

### LINKED ISSUE #859: need a way to conditionally render fields
## Feature: hide some fields from the Service Bundle from the UI

As seen by the submission of PR #834, there is a desire to have the ability to specify when a parameter should be displayed based on some condition or set of conditions. There is also a desire to have a more robust parameter set in the `apb.yml` to include objects etc. INSERT FOLLOW ISSUE HERE. This conditional feature needs to be included in the larger parameter discussion.

PhilipGough:

Ready for review

Change to parse the dependencies defined by apb.yml and map them to valid json schemas

depends-on https://github.com/automationbroker/bundle-lib/pull/70

fixes #859


maleck13:
@jmrodri example of this working with the catalog pr linked above 
https://youtu.be/MX-G0TjaP7E

dymurray:
Please see my comments  in bundle-lib#70 about documentation updates and incrementing of the bundle spec to `1.1`. I would also encourage some changes to this PR that would do a check on the version or at least be backwards compatible so that the version is more just a representation of the spec than actually changing code paths.

PhilipGough:
cc @maleck13 Still work to be done here but be nice to get some early feedback.

pkg/broker/util_test.go
```diff
@@ -474,3 +474,102 @@ func TestInitMetadataCopy(t *testing.T) {
...
+			Title: "Example Also Shown When True",
+			Dependencies: []apb.Dependency{
+				{
+					Key: "eg_boolean_param",
+				},
+			},
+		},
+	}
+
+	expectDependencyJson := `
```
shawn-hurley: Lint is complaining should be `expectDependencyJSON`

PhilipGough:
@shawn-hurley I've update the bundle-lib dependency as part of this PR FYI

PR #929, opened 2018-05-04
Add fix for get si npe

eriknelson:
@shawn-hurley and I discovered a panic this morning. If for some reason the catalog has a bunch of instances that the broker is unaware of, the broker will try to lookup the instance, fail to find it, and panic when trying to deref the spec off a nil value. This correctly checks the error.


shawn-hurley:
Pretty soon we can start to write tests for the broker if we want. We have mocked for dao and we just need to figure out the mocks for apb.Executor/async engine. Just a thought for later

Issue #934, opened 2018-05-08
cannot update optional parameter


odra:
## Bug:

**What happened**:

I've created an optional updatable parameter (apb file definition):

```yaml
- name: ANDROID_LICENSE_AGREEMENT
      default: ""
      pattern: ^yes$
      type: string
      description: "Type 'yes' above to accept the Android SDK License Agreement. https://developer.android.com/studio/terms.html"
      title: Android SDK License Agreement
      required: True
      display_group: Android SDK
    - name: ANDROID_SDK_AUTO_CONFIG
      default: 'Yes'
      type: string
      display_type: enum
      enum: ["Yes", "No"]
      description: "Installs the default android sdk packages"
      title: Android SDK Auto Package Install
      required: False
      display_group: Android SDK
    - name: ANDROID_SDK_CONFIG
      default: ""
      type: string
      display_type: textarea
      description: "Android SDK config file content"
      title: Android SDK package config
      required: True
      updatable: True
      display_group: Android SDK
```

I tried using `required: False` but it didn't make a difference.

I get the following error when updating the field:

```
[2018-05-08T13:05:32.152Z] [DEBUG] - Request Params: map[ANDROID_LICENSE_AGREEMENT:yes ANDROID_SDK_AUTO_CONFIG:No ANDROID_SDK_CONFIG:[base]\ntools\nplatform-tools\n\n[platforms]\nandroid-24\nandroid-25\nandroid-26\n\n[build-tools]\n25.0.1\n25.0.3\n\n[extras:google]\ngoogle_play_services\nm2repository\n\n[extras:android]\nm2repository\n\n[keystore]\nalias=AndroidDebugKey\nname=AG\nunit=AeroGear\norg=RedHat\nloc=Waterford\nstate=WD\ncountry=IRL\nstorepass=android\nkeypass=android"]
--
  | [2018-05-08T13:05:32.152Z] [DEBUG] - Previous Params: map[ANDROID_SDK_AUTO_CONFIG:No _apb_plan_id:default _apb_service_class_id:4319a63a322ba4ddcc0d855908ac4f1f _apb_service_instance_id:085d2de7-0b21-4cce-91e5-6cc64adc717b ANDROID_LICENSE_AGREEMENT:yes]
  | [2018-05-08T13:05:32.152Z] [NOTICE] - Requested update parameter: [%!v(MISSING)] with the value of [%!v(MISSING)] was found in an update request, but it did not exist on the previous service instance.
[...]

PR #935, opened 2018-05-08
Add requested parameters when not in existing service instance

djzager:
Add the requested update parameter to the changed parameters when not found in the existing serviceinstance.

Fixes: #934


pkg/broker/broker.go
```diff
@@ -1398,6 +1398,7 @@ func (a AnsibleBroker) validateRequestedUpdateParams(
 			// can make sense of them.
 			log.Noticef("Requested update parameter: [%v] with the value of [%v] was found " +
 				"in an update request, but it did not exist on the previous service instance.")
+			changedParams[reqParam] = reqVal
```
shawn-hurley: If this is all we are going to do then we should just remove the entire if/else chain and always update IMO

djzager: I'm not against that.

djzager:
This should go against master once `release-1.2` branch is created.

PR #940, opened 2018-05-09
doc to show how to use broker via vendor

jmrodri:


docs/using_as_library.md
```diff
@@ -0,0 +1,205 @@
+# Using as a library
+
+The Automation Broker implements the Open Service Broker API. As a broker it
+provides quite a bit of behavior that might want to be used by other broker
+authors, for example, asynchronous job support, ...
```
djzager: nit: `asynchronous job support, etc.`

docs/using_as_library.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Vendoring
+
+To get the Automation Broker as a dependency, here is an example of the
+`Gopkg.toml` used by `dep`. You can use any of the types supported by the `dep`
```
djzager: Here is an excerpt from `Gopkg.toml` used by [golang/dep](https://golang.github.io/dep/) that adds the Automation Broker as a dependency.

docs/using_as_library.md
```diff
@@ -0,0 +1,205 @@
...
+```toml
+[[constraint]]
+  branch = "master"
+  name = "github.com/openshift/ansible-service-broker"
+```
+
+## Extending
+
+Once we have the dependency installed in our project, how do we use it? The
+Automation Broker will do most if not all of the heavy lifting, but allows the
```
djzager: nit: will do most, if not all, ...

docs/using_as_library.md
jmrodri: addressed

docs/using_as_library.md
jmrodri: addressed

docs/using_as_library.md
jmrodri: addressed

PR #947, opened 2018-05-11
Bug 1576728 - Fixes multiple requests to a CR

shawn-hurley:
This makes the dao calls happen 1 at a time for a dao. Attempts to handle the case when two service instances are gotten with two bindings. both are deleted, but because we only remove each of one, and when we update we just set the value, we were adding back a binding.

Depends on https://github.com/automationbroker/bundle-lib/pull/80

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -44,7 +44,6 @@
   version = "0.1.2"
 
 [[projects]]
-  branch = "bug-1576728"
   name = "github.com/automationbroker/bundle-lib"
   packages = [
     "apb",
@@ -57,8 +56,8 @@
     "registries/adapters",
     "runtime"
   ]
-  revision = "4162f5a4f0a5bac5e074b1473b738c4c20779b40"
-  source = "github.com/shawn-hurley/bundle-lib"
+  revision = "d1bf04cf29b67929313e998b83d07c807f772c33"
+  version = "0.1.6"
 
 [[projects]]
   branch = "master"
@@ -925,6 +924,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "b33df39ec11fc620bb6629c612363cae755b3e4843a7e56a8ce4545f86bd7507"
+  inputs-digest = "ec45f8aae9030f4422019e0af5f9a39793f532beb467d9953baf8e8e9316847a"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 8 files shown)


Gopkg.lock
```diff
@@ -44,6 +44,7 @@
   version = "0.1.2"
 
 [[projects]]
+  branch = "bug-1576728"
```
jmrodri: will this remain when we merge?

Gopkg.toml
```diff
@@ -37,7 +37,9 @@
   version = "~1.3.0"
 
 [[constraint]]
-  version = "~0.1.0"
+  #version = "~0.1.0"
+  branch = "bug-1576728"
+  source = "github.com/shawn-hurley/bundle-lib"
```
jmrodri: are we doing this?

pkg/dao/crd/dao.go
```diff
@@ -47,13 +48,20 @@ const (
...
 }
 
 // NewDao - Create a new Dao object
 func NewDao(namespace string) (*Dao, error) {
-	dao := Dao{namespace: namespace}
+	dao := Dao{namespace: namespace,
+		bundleLock:   sync.Mutex{},
+		bindingLock:  sync.Mutex{},
+		instanceLock: sync.Mutex{},
+	}
```
jmrodri: CRDs suck as a database :)

Gopkg.toml
shawn-hurley: No, this depends on this https://github.com/automationbroker/bundle-lib/pull/80

pkg/dao/crd/dao.go
```diff
@@ -47,13 +48,20 @@ const (
 
 // Dao - object to interface with the data store.
 type Dao struct {
-	client    automationbrokerv1.AutomationbrokerV1alpha1Interface
-	namespace string
+	client       automationbrokerv1.AutomationbrokerV1alpha1Interface
+	namespace    string
+	bundleLock   sync.Mutex
```
eriknelson: :+1:

rthallisey:
https://github.com/automationbroker/bundle-lib/pull/80 just merged. Should be able to remove  ```branch = "bug-1576728"``` everywhere

eriknelson:
@rthallisey I think we probably want to update the Gopkg stuff to reference a release tag of bundle-lib.

eriknelson:
This is ready to merge with the bundle lib release.

PR #953, opened 2018-05-15
Bug 1572129 - recover in-progress mediawiki provision and deprovision jobs

jmrodri:


shawn-hurley:
@jmrodri did we create a issue to track the refactor of recover? 

I was planning on putting together a proposal soon for this but I think it would be worth linking this PR to that issue?

PR #954, opened 2018-05-16
update obj form definition

odra:

I opened this issue a couple of days ago: https://github.com/openshift/origin-web-console/issues/2985 because the `update` form definition object was always empty so I made this tiny change into the service broker to include all updatable fields into the service instance object.

I am sending this PR to ask for feedback about this change, if this makes sense at all.

 - Fills the update form definition object up with  a list of updatable fields (if any)

This PR breaks the service instance ui (when listing read-only parameters) which will need a change as well if this change makes any sense.

### DIFF
--- a/pkg/broker/util.go
+++ b/pkg/broker/util.go
@@ -138,9 +138,8 @@ func initMetadataCopy(original map[string]interface{}) (map[string]interface{}, 
 	return dst, nil
 }
 
-
-func createUpdateFormDefinition(params []apb.ParameterDescriptor) []interface{} {
-	var updateParams []apb.ParameterDescriptor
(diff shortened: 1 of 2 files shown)


odra:
This is how the service instance params are shown with this change (service detail view):

service-detail-secret-1
service-detail-secret-2

This is how those properties should be displayed:

screenshot from 2018-05-16 10-04-45
screenshot from 2018-05-16 10-04-50

shawn-hurley:
I think this is a good change. Just a heads up that this will be on hold until master branch opens up. Currently, only bug fixes that have a bugzilla for openshift 3.10 release can be brought in. We are trying to make a release branch so the master branch can open up soon.

jmrodri:
@shawn-hurley is this still on hold? I assume your comment is no longer valid.

PR #962, opened 2018-05-19
dep prune vendor

jmrodri:
dep now can remove tests and non-go files from dependencies which bloat
the repo. There is no reason to have these files in the repo as they do
not get compiled.

The BEST part is the vendor tree shrinks dramatically.

### Before the prune:
```
$ du -sh vendor
258M    vendor
```

### After the prune:
```
$ du -sh vendor
46M     vendor
```


jmrodri:
https://golang.github.io/dep/docs/Gopkg.toml.html#prune

Gopkg.toml
```diff
@@ -75,3 +75,8 @@
 [[constraint]]
   name = "k8s.io/kubernetes"
   version = "1.9.1"
+
+[prune]
+  unused-packages = true
+  go-tests = true
```
shawn-hurley: Should we add `non-go = true` to this as well it will reduce the size to 40M?

PR #965, opened 2018-05-23
Bug 1581580 - skip if bind instance does not have credentials

shawn-hurley:


cmd/migration/main.go
```diff
@@ -239,6 +239,11 @@ func main() {
 	for _, bi := range biSaved {
 		labels := map[string]string{"apbAction": "bind"}
 		extcredsJSON, err := etcdDao.GetRaw(fmt.Sprintf("/extracted_credentials/%v", bi.ID))
+		// If no extracted credential for the binding, then it is using the service instance's
+		// extracted credentials and we should assume that this is the case.
```
djzager: are we basically just saying, it's okay if a binding has no extracted credentials?

PR #972, opened 2018-05-30
Bug 1583064 - missing action causes many sandboxes

jmrodri:
If the unbind fails we really want to let them know it failed because it
is probably something the APB developer needs to fix. But if the action
is non-existent we really need to prevent the multiple sandboxes.

If we can't find the action, there is no point in continuing.
marking job as SUCCEEDED but setting the message to be the actual error
string. This will only happen for the optional actions: bind & unbind.

Update isJobInProgress to take a string to allow the proper ID to be
passed in.

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -614,7 +614,7 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest, 
 	// away from []byte it can still be evaluated.
 	if si != nil && uuid.Equal(si.ID, serviceInstance.ID) {
 		if reflect.DeepEqual(si.Parameters, serviceInstance.Parameters) {
-			alreadyInProgress, jobToken, err := a.isJobInProgress(serviceInstance, apb.JobMethodProvision)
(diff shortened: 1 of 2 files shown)


pkg/broker/broker.go
```diff
@@ -732,7 +732,7 @@ func (a AnsibleBroker) Deprovision(
 		return nil, err
 	}
 
-	alreadyInProgress, jobToken, err := a.isJobInProgress(&instance, apb.JobMethodDeprovision)
+	alreadyInProgress, jobToken, err := a.isJobInProgress(instance.ID.String(), apb.JobMethodDeprovision)
```
shawn-hurley: 👍

pkg/broker/job_state_subscriber.go
```diff
@@ -35,6 +36,28 @@ func (jss *JobStateSubscriber) Notify(msg JobMsg) {
 	if isBinding(msg) {
 		id = msg.BindingUUID
 	}
+
+	log.Debugf("message: method: [%v] isNotFound? [%v] state error [%v]", msg.State.Method, msg.State.Error == runtime.ErrorActionNotFound.Error(), msg.State.Error)
```
shawn-hurley: NIT: I think this might have been for testing is it still needed?

PR #976, opened 2018-06-04
Bug 1585649 - wait group should be added to before go function

shawn-hurley:


djzager:
Does this depend on the fix for mediawiki 
https://github.com/ansibleplaybookbundle/mediawiki-apb/pull/36 so that it will succeed on deprovision?

jmontleon:
Probably want https://github.com/ansibleplaybookbundle/mediawiki-apb/pull/35 rather than 36 if this fixes the root issue. 36 was just an additional workaround.

PR #983, opened 2018-06-11
adding unit tests for broker package

shawn-hurley:
Example of how to implement tests for the broker package.

 - test cases for job_state_subscriber moves to 81.7%
 - test cases for jobs moves to 70%
 - broker.go moves to 7.7%


pkg/broker/broker_test.go
```diff
@@ -17,24 +17,211 @@
...
+func TestGetServiceInstance(t *testing.T) {
+	u := uuid.NewUUID()
+	errIsNotFound := fmt.Errorf("is not found")
+	unknownError := fmt.Errorf("is not found")
+	testCases := []struct {
+		name            string
+		dao             *mocks.Dao
+		config          Config
+		addExpectations func(*mocks.Dao)
+		shouldErr       error
```
maleck13: nit: I think most of tests like this use ```ExpectError``` not a big deal though

maleck13: ignore it is name the same usage

pkg/broker/broker_test.go
```diff
@@ -17,24 +17,211 @@
...
+					},
+				}, nil)
+			},
+		},
+		{
+			name:   "get service instance error is not found",
+			dao:    new(mocks.Dao),
+			config: Config{},
+			addExpectations: func(d *mocks.Dao) {
+				d.On("GetServiceInstance", u.String()).Return(nil, unknownError)
```
maleck13: just trying to understand. Is this specifying the behavior when the GetServiceInstance is called or is it the expected result?

pkg/broker/jobs_test.go
```diff
@@ -269,6 +271,190 @@ func TestApbJobRun(t *testing.T) {
...
+				if len(messages) != 2 {
+					return fmt.Errorf("expected 2 job messages")
+				}
+				first := messages[0]
+				if first.State.State != apb.StateInProgress {
+					return fmt.Errorf("unexpected first message contents")
+				}
+
+				second := messages[1]
+				log.Infof("second: %#v", second)
```
maleck13: do we want to keep the log message?

pkg/broker/jobs_test.go
```diff
@@ -269,6 +271,190 @@ func TestApbJobRun(t *testing.T) {
...
```
maleck13: same might want to remove

pkg/broker/broker_test.go
shawn-hurley: I think the answer is both.

https://github.com/stretchr/testify#mock-package 

```go
// TestSomething is an example of how to use our test object to
// make assertions about some target code we are testing.
func TestSomething(t *testing.T) {

  // create an instance of our test object
  testObj := new(MyMockedObject)

  // setup expectations
  testObj.On("DoSomething", 123).Return(true, nil)

  // call the code we are testing
  targetFuncThatDoesSomethingWithObj(testObj)

  // assert that the expectations were met
  testObj.AssertExpectations(t)
```

jmrodri:
broker test coverage increased from 27.9% to 32.5%

PR #984, opened 2018-06-11
Use test playbook to test broker in ci

djzager:
Update travis to use the broker-apb to test the broker via the `test` playbook.


djzager:
Even if we like these changes. We should hold this to know what the impact would be to Jenkins before merging.

djzager:
I have more to work out here than I thought. I'm closing this for now.

PR #989, opened 2018-06-14
adds docs and plumbing for the skip_verify_tls setting on the rhcc adapter

mhrivnak:

This enables testing the broker with stage registries that use a self-signed CA. It could also be useful for other scenarios. It is a companion PR to https://github.com/automationbroker/bundle-lib/pull/98

 - surfaces a setting that exposes a feature of bundle-lib that can skip TLS cert verification

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -143,7 +143,10 @@ registry:
     url: <rhcc url>
     white_list:
       - ".*-apb$"
+    skip_verify_tls: false
 ```
+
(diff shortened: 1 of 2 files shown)


djzager:
Is this something we should add to the APB?

https://github.com/openshift/ansible-service-broker/blob/master/apb/templates/broker.configmap.yaml

+ something like `broker_skip_verify_tls: false` in https://github.com/openshift/ansible-service-broker/blob/master/apb/defaults/main.yml

PR #990, opened 2018-06-14
Use broker-apb test playbook to test the broker

djzager:


djzager:
Before this goes in, because it will break Jenkins, I want to have the changes for Jenkins lined up.

PR #991, opened 2018-06-15
Pass in namespace to runtime

maleck13:
the namespace for the broker can change. So ensure it is passed through to the runtime package at startup.

While testing some state changes, I noticed state was not being saved this was due to the fact it was expecting to save to the ns ansible-service-broker.

While running make check I noticed that several tests were failing due to compilation errors so I fixed these also.


maleck13:
ping @shawn-hurley

shawn-hurley:
I think we need to merge this @djzager @dymurray @rthallisey you guys mind checking this out?

PR #995, opened 2018-06-21
removing subject rules review auth and using aggregated rules.

shawn-hurley:
Implementation of https://trello.com/c/fiOSsq1Q
 - remove origin package
 - remove rules review
 - use bundle-lib authorizer to determine if user  ca

depends-on <PR>
https://github.com/automationbroker/bundle-lib/pull/112


Gopkg.toml
```diff
@@ -37,8 +37,10 @@
   version = "~1.3.0"
 
 [[constraint]]
-  version = "~0.2.0"
+  #version = "~0.2.0"
```
dymurray: I'm assuming we don't want these changes in the PR.

apb/templates/broker-user-auth.yaml
```diff
@@ -0,0 +1,17 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: asb-user-access
```
djzager: I think this name, since it's cluster scope, should be related to the `{{ broker_name }}`

apb/templates/broker-user-auth.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    {% if broker_sandbox_role == 'admin' %}
```
djzager: I think it needs to be formatted more like:

```
labels:
{% if broker_sandbox_role == 'admin' %}
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
{% endif %}
...
```

Otherwise, I think the whitespace is included.

apb/templates/broker-user-auth.yaml
djzager: or...maybe not, but at least not `asb-user-access` :sunglasses:

apb/templates/broker-user-auth.yaml
```diff
@@ -0,0 +1,17 @@
+---
+kind: ClusterRole
```
djzager: The only thing I don't see is this template being used in deployment.

djzager:
If you are using aggregated reviews, does that impact our need for https://github.com/shawn-hurley/ansible-service-broker/blob/a85d2a365ac1fb293dae0951ee61320038649523/apb/tasks/main.yml#L8 ?

shawn-hurley:
Yes, I think the default should be authorization for both k8s and
openshift.

Gopkg.toml
shawn-hurley: This depends on the bundle lib PR so will update once we get a release of bundle lib

apb/templates/broker-user-auth.yaml
shawn-hurley: I am very confused by this statement sorry can you explain what I need to do?

djzager: I would expect to see `broker-user-auth.yaml` somewhere in this list (https://github.com/shawn-hurley/ansible-service-broker/blob/a85d2a365ac1fb293dae0951ee61320038649523/apb/tasks/main.yml#L54) in order for the `broker-user-auth ClusterRole` to be created.

PR #997, opened 2018-06-26
Reconciliation Proposal

alaypatel07:

Add spec-reconciliation.md

### DIFF
--- a/docs/proposals/spec-reconciliation.md
+++ b/docs/proposals/spec-reconciliation.md
@@ -1,0 +1,91 @@
+# Spec Reconciliation Proposal
+
+## Introduction
+Add a way for broker to sync the list of specs it has with registries while handling the following corner cases:
+ 1. Temporary loss of connection with registry
+ 2. A spec removed from the registry but has provisioned instance/s
+
+## Problem Description
+
+The current approach of syncing the spec from registries is as follows:
+ 1. Delete all the specs currently in the datastore
+ 2. Fetch new specs
+ 3. Add all the fetched specs to the datastore
+ 
+This leads to a bunch of bugs [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345)
+
+
+Instead, the broker can take the following actions:
+####Approach#1
+ 1. Fetch the specs from registry
+    1. If the fetch fails increment a failed connection counter. If the failed connection counter reaches a limit delete all the specs that do not have a service instance
+    2. If the fetch succeeds:
+        1. add new specs to the datastore
+        2. delete that specs that are removed from datastore and do not have a service instance.
+        3. mark the specs that are removed from datastore and has a service instance
+ 2. During deprovision, if the instance getting deprovisioned is the last one delete the spec from datastore
+ 3. Serve the catalog request with specs that are not marked
+  
+This will solve the problem but the task of deleting the specs is spread across time (when fetching specs and when deprovisioning)
+
+####Approach#2
+ 1. Fetch the specs from registry
+    1. If the fetch fails mark all the specs for deletion
+    2. If the fetch succeeds,
+       1. add new specs to the datastore
+       2. mark the specs that are removed from the registry.
+ 3. Run a background task to delete all the marked specs that does not have a service instance currently provisioned. The interval of running this task can be configurable
+ 4. Serve the catalog request with specs that are not marked
+ 
+This solves the problem but introduces a new parameter of configuration. 
+
+If in future, other conditions of deleting a spec crop up, just mark it for deletion and background task will take care of deleting it safely.
+
+
+Either approach will take care of the all objectives and bugs.
+
+##Other ideas
+
+####APB Versioning
+
+Currently, tags are the only way an apb developer can version an apb. If the developer is given a way to do an explicit versioning of the apb irrespective of the tags, there might never be a need to allow multiple tags
+
+A rough idea of what can be done:
+
+ 1. Add a new field to apb.yml call bundleVersion
+ 2. During provision save the value of "bundleVersion" with the service instance
+ 3. During de-provision pass the value of "bundleVersion" in the args.
+ 
+This allows:
+ 1. the developer to have a versioning scheme that is independent of the tags.
+ 2. the developer can push v1.1 of that apb even if v1.0 is provisioned. The version number passed during binding, unbinding and deprovision would still be v1.0
+ 3. the admin can know what version of apb is provisioned in the cluster
+ 
+
+####Multiple tags
+
+The requirement supporting multiple tags of apb is largely because versioning of an apb is tied to the tag right now. There needs to be a way in which admins can have multiple versions of the same apb. If the apb versioning scheme mentioned above is implemented, there might not be a need to support multiple tags
+
+If there is, the registry section of configmap of the broker can be modified as follows:
+
+    registry:
+      - type: dockerhub
+        name: dh
+        url: https://registry.hub.docker.com
+        org: ansibleplaybookbundle
+        tag:
+          default: latest
+          named: 
+            - hello-apb: 
+              - v1.0
+              - v2.0
+            - postgres-apb: 
+              - v3.9
+        white_list:
+          - '.*-apb$'
+
+
+This would require modification to the bootstrap function where after fetching all the specs, the broker needs to iterate through the list of named specs and fetch specific tags from the registry
+
+Adding a method to adaptor interface that takes two inputs, image name and image tag, and fetches its spec. The broker will then call this method iteratively for all named pair of specs and tags
+          
--- a/docs/proposals/spec-reconciliation.md
+++ b/docs/proposals/spec-reconciliation.md
@@ -4,6 +4,8 @@
 Add a way for broker to sync the list of specs it has with registries while handling the following corner cases:
  1. Temporary loss of connection with registry
  2. A spec removed from the registry but has provisioned instance/s
+ 3. The service catalog sends a provision request to the broker after the broker has started bootstrapping but before it has completed bootstrapping (untimely provision request)
+ 4. The service catalog sends a provision request to the broker for specs that has been marked deleted in the broker (out of sync provision request)
 
 ## Problem Description
 
@@ -12,11 +14,17 @@ The current approach of syncing the spec from registries is as follows:
  2. Fetch new specs
  3. Add all the fetched specs to the datastore
  
-This leads to a bunch of bugs [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345)
+This leads to a bunch of bugs:
+ - [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495)
+ - [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495) 
+ - [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345)
 
+The underlying problem with all those bugs is that the broker is deleting specs even if it is provisioned. When the catalog tries to deprovision, bind, unbind with these dangling instances (instances with no specs) it will fail. Hence the broker needs a way to safely delete the specs.
 
-Instead, the broker can take the following actions:
-####Approach#1
+### Approaches to safely deleting specs
+
+The broker can take the following actions to safely delete the specs:
+#### Approach#1
  1. Fetch the specs from registry
     1. If the fetch fails increment a failed connection counter. If the failed connection counter reaches a limit delete all the specs that do not have a service instance
     2. If the fetch succeeds:
@@ -25,10 +33,10 @@ Instead, the broker can take the following actions:
         3. mark the specs that are removed from datastore and has a service instance
  2. During deprovision, if the instance getting deprovisioned is the last one delete the spec from datastore
  3. Serve the catalog request with specs that are not marked
-  
+ 
 This will solve the problem but the task of deleting the specs is spread across time (when fetching specs and when deprovisioning)
 
-####Approach#2
+#### Approach#2
  1. Fetch the specs from registry
     1. If the fetch fails mark all the specs for deletion
     2. If the fetch succeeds,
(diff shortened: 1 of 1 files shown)


docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,91 @@
...
+ 2. A spec removed from the registry but has provisioned instance/s
+
+## Problem Description
+
+The current approach of syncing the spec from registries is as follows:
+ 1. Delete all the specs currently in the datastore
+ 2. Fetch new specs
+ 3. Add all the fetched specs to the datastore
+ 
+This leads to a bunch of bugs [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495), [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345)
```
djzager: I think these would be better added as a list with a small summary of the problem.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+Instead, the broker can take the following actions:
+####Approach#1
```
djzager: I think this should be `### Approach 1`

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,91 @@
...
+ 4. Serve the catalog request with specs that are not marked
+ 
+This solves the problem but introduces a new parameter of configuration. 
+
+If in future, other conditions of deleting a spec crop up, just mark it for deletion and background task will take care of deleting it safely.
+
+
+Either approach will take care of the all objectives and bugs.
+
+##Other ideas
```
djzager: My gut feeling is that these should be separate proposals and are immediately related to the reconciliation problem.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,91 @@
...
+
+Instead, the broker can take the following actions:
+####Approach#1
+ 1. Fetch the specs from registry
+    1. If the fetch fails increment a failed connection counter. If the failed connection counter reaches a limit delete all the specs that do not have a service instance
+    2. If the fetch succeeds:
+        1. add new specs to the datastore
+        2. delete that specs that are removed from datastore and do not have a service instance.
+        3. mark the specs that are removed from datastore and has a service instance
+ 2. During deprovision, if the instance getting deprovisioned is the last one delete the spec from datastore
```
djzager: How would you know this is the last service instance associated with a particular bundle spec?

alaypatel07: Either maintain a counter of the number of instances provisioned in bundle.Spec or compare all the instance specs with the spec of the instance being deprovisioned. I think the former would be better

docs/proposals/spec-reconciliation.md
alaypatel07: ok sure

docs/proposals/spec-reconciliation.md
alaypatel07: yes I can add a summary describing the problem

alaypatel07:
@djzager both the approaches have pros and cons which approach do you think I should pick to refine?

And yes, I started looking into the service catalog reconciliation but the biggest difference in their environment is, they are getting all the changes as a callback. So, I stopped looking, thinking their approach would not be applicable in our environment as we use polling to sync the specs.

docs/proposals/spec-reconciliation.md
djzager: I think this one needs to be updated `[Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495)`

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+ - [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495)
+ - [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495) 
+ - [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345)
+
+The underlying problem with all those bugs is that the broker is deleting specs even if it is provisioned. When the catalog tries to deprovision, bind, unbind with these dangling instances (instances with no specs) it will fail. Hence the broker needs a way to safely delete the specs.
+
+### Approaches to safely deleting specs
+
+The broker can take the following actions to safely delete the specs:
+#### Approach#1
```
shawn-hurley: NIT: space between `#` and `1`

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+ 1. Fetch the specs from registry
+    1. If the fetch fails increment a failed connection counter. If the failed connection counter reaches a limit delete all the specs that do not have a service instance
+    2. If the fetch succeeds:
+        1. add new specs to the datastore
+        2. delete that specs that are removed from datastore and do not have a service instance.
```
shawn-hurley: I think that we should be looking at bundle instances and not service instances

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Approaches to handle untimely provision request
+
+
+To to handle a case where provision is requested after the broker has started bootstrapping but before it has completed bootstrapping, the broker can take the following approaches:
+
+#### Approach#1
+The broker always sends an error code if it is busy provisioning. OSB [spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-2) does not have an appropriate error code for busy.
+
+#### Approach#2
```
shawn-hurley: I think that a combination of 2 and 3 is an option. in this scenario we start the provision with some soft of waiting mechanism, that will get notified when bootstrap finishes. While this is happening async provision last operation request's can be called just like they are now and will error if the spec is deleted.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Approach#3
+It can send a 202 Accepted status, suggesting that provisioning is in progress. The service catalog will when keep polling to check if the service instance is fully provisioned. The [last operation](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-1) has a mechanism where broker can tell if the provision is success of failure.
+
+
+## Other ideas
+
+These ideas are not directly related to the objectives of this proposal but are mentioned here and might become independent proposals of their own when addressing those issues.
+
+#### APB Versioning
```
shawn-hurley: I think this section is a different PR.

docs/proposals/spec-reconciliation.md
jmrodri: I like Approach 1 best but I'd like to change it a bit.
1. Fetch the specs from a registry
2. Fetch the specs from the datastore
3. Delete any specs in the datastore currently marked for deletion, assuming there is no instance that has a reference to this spec which is marked for deletion.
   i. if there is a reference, leave the spec alone
  ii. if there is no reference, delete the spec
4. For each spec from the registry, get the spec from the datastore
   i. if found, update it, and save to datastore
   ii. if not found, add it to the datastore
5. For each spec in the datastore, NOT in registry list, mark it for deletion.

Connection failure
   i. if connection fails, retry 10 times backing of 10 seconds each time, then give up until the next bootstrap loop runs
   if the bootstrap loop is < our retry time, I'd abort and wait for the next bootstrap loop to run.

Catalog payload should change such that it returns all specs that are NOT marked for deletion

I would **not** do reference counting, I would simply look for any instances that have a reference to a spec that was marked for deletion. If so, then we do not delete it.

docs/proposals/spec-reconciliation.md
eriknelson: The bundle spec should only get deleted in the case that it had been previously marked for deletion, but was *not* deleted because it had an outstanding instance, correct?

docs/proposals/spec-reconciliation.md
eriknelson: Agreed with the above comments. I don't think 1 is reasonable, 2 shouldn't block. It would be semantically correct to respond with a 202 saying "okay, we've accepted your request and we're working on it". The fact that we're waiting on the bootstrap to complete is a broker implementation detail; it has nothing to do with the OSB spec.

Once the bootstrap completes, the job waiting should then investigate the current state of the specs. If the requested instance refers to a spec that has been deleted, we can mark the job failed. If the spec still exists, proceed as normal.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Approach #3
+It can send a 202 Accepted status, suggesting that provisioning is in progress. The service catalog will when keep polling to check if the service instance is fully provisioned. The [last operation](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-1) has a mechanism where broker can tell if the provision is success of failure.
+
+### Approaches to handle out of sync provision request
+
+To handle the case where service catalog sends a request to provision an instance of a spec that has been marked for deletion the broker can take the following approaches:
+
+#### Approach#1
+Successfully serves the provision request (very similar to scenario #1): This approach might lead to unexpected problems during execution.
```
eriknelson: I'm not sure what it means to "successfully serve the provision request" here, but I don't think this is the correct approach. If we know the spec has been marked for deletion, IMO it should effectively not exist anymore, and therefore we should respond as if it doesn't exist.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Approach#2
+Error out the provision request. It is clear with the polling approach of the Service Catalog to update the [specs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-2), that this scenario will eventually popup, but the specs doesnt have an error code specifically meant for this scenario.
+
+#### Approach#3
+It can send a 202 Accepted status, suggesting that provisioning is in progress. The service catalog will when keep polling to check if the service instance is fully provisioned. The [last operation](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-1) has a mechanism where broker can tell if the provision is success of failure.
```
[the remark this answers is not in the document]
eriknelson: I'm not convinced we should respond with a 202 on the initial response if we know the spec has been marked for deletion. It's effectively gone at that point, and I don't think we should accepting any more requests for instances (I think I'm advocating a bad request response here).

docs/proposals/spec-reconciliation.md
eriknelson: +1

jmrodri:
So I had a private conversation with Alay earlier today, and another one with Erik & Shawn. Figured I'll post my comments here:

> not a huge reason, if reaper was preferred by majority, I'd be okay with it. I just wouldn't do that if I was implementing this.

> a reaper will have to be yet another "goroutine" running that interacts with a bootstrap "goroutine" and each provision/deprovision, etc is another set of "goroutines".

> I don't think deprovision should ever delete a spec. If something is marked for deletion and a deprovision comes in, the spec should remain marked for deletion until the bootstrap reruns. Basically my in place bootstrap loop is basically a reaper process as well. So when the bootstrap runs again, first thing it should do is remove the specs that were marked for deletion and no longer provisioned.

> and I don't think provision or deprovision should ever have to have logic about a marked for deletion spec. When they ask for a spec the method should filter it for them.

> though I might be glossing over some details, I'd have to look at the code again to see how we get them for provision, deprovision. /me assumes it's just a getSpec or something like that
```
    if spec, err = a.dao.GetSpec(specID); err != nil {
```

docs/proposals/spec-reconciliation.md
dymurray: I agree with @shawn-hurley, we already frequently hit 504s with the broker route when bootstrapping I would prefer to not add too much computation to the bootstrap loop.

docs/proposals/spec-reconciliation.md
dymurray: I like approach 3 the best and agree with the above comments.

docs/proposals/spec-reconciliation.md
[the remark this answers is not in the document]
dymurray: +1 to Erik's comment. Feels weird for this to be a 202.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+ 2. During provision save the value of "bundleVersion" with the service instance
+ 3. During de-provision pass the value of "bundleVersion" in the args.
+ 
+This allows:
+ 1. the developer to have a versioning scheme that is independent of the tags.
+ 2. the developer can push v1.1 of that apb even if v1.0 is provisioned. The version number passed during binding, unbinding and deprovision would still be v1.0
+ 3. the admin can know what version of apb is provisioned in the cluster
+ 
+
+#### Multiple tags
```
dymurray: This should be a separate proposal as well IMO.

docs/proposals/spec-reconciliation.md
```diff
@@ -0,0 +1,130 @@
...
+              - v2.0
+            - postgres-apb: 
+              - v3.9
+        white_list:
+          - '.*-apb$'
+
+
+This would require modification to the bootstrap function where after fetching all the specs, the broker needs to iterate through the list of named specs and fetch specific tags from the registry
+
+Adding a method to adaptor interface that takes two inputs, image name and image tag, and fetches its spec. The broker will then call this method iteratively for all named pair of specs and tags
```
alaypatel07: @jmrodri @eriknelson @shawn-hurley I think the ***in-place merge*** (approach 1) could still be a good way to achieve reconciliation. My modified approach is as follows:

1. Maintain a ***bundle instance counter*** along with each spec in the CRD.
2. Compare the list of registrySpecs with crdSpecs.
3. Add all the specs that present in registrySpecList but absent in crdSpecList, with bundle instance counter initialized to 0
4. Remove all specs present in crdSpecList but absent in registrySpecList and with bundle instance counter value 0

Increment the spec bundle instance counter during provision
Decrement the spec bundle instance counter during deprovision

I think step 2 and 3 in the above algorithm would be the ones that would take most of the time. But those steps would be required even if the ***reaper approach*** is taken, to mark the specs for deletion. Maintaining the counter value would help save the time consuming computation required to determine whether a spec has a bundle instance. This will also avoid all the concurrency problems that complicates the ***reaper approach***

jmrodri: I don't like maintaining a separate counter because that is information we can calculate by looking for instances that have references to the specs. While that's not trivial, it is accurate. My gut tells me those counters are going to get out of date and create another issue. But that's just me, I'll let others chime in

alaypatel07: @jmrodri @eriknelson @shawn-hurley I think modifying the reaper approach as follows could be one way to handle reconciliation with some simplicity:

1. There is a background process reaper which listens for data from bootstrap. Bootstrap runs periodically as it does now.
2. The bootstrap process is responsible for determining which spec should be added and adding them.
3. The bootstrap process will then determine which spec to be deleted and send them to reaper(one by one as it determines).
4. The reaper always runs in the background waiting to get a spec from bootstrap. When the reaper gets a spec, it deletes the spec if there are no bundle instances, else it will ignore the spec.

This simplified approach gets rid of the **mark for deletion** field. The reaper now tries to **safely delete** the specs it gets from bootstrap. If it is not safely able to delete, it will ignore it. Since the adding of specs is done by bootstrap, and safe removal is done by reaper, they no longer compete for a shared resource. This will always ensure a desirable state of the broker.

eriknelson: Unfortunately the delete is not the expensive operation, it's the iteration of outstanding service instances to determine which instances correspond to specs, so it's not desirable to have the bootstrap process determine the specs that are safe to delete. The purpose of the reaper is to delegate this work to the background.

jmrodri:
I met with @eriknelson  & @alaypatel07  to discuss the bootstrap. The main issue is deleting ALL specs causes many bugs, primarily in the cases where they are going to be added again.

Each of the 3 bugs identified in the proposal was case where we deleted the specs, before we were able to process the specs from the registry. Either, we didn't add the spec yet or there was an error
reaching the registry. Both scenarios exhibited the same behavior because our process was to delete ALL specs first.

Given the above, we came to the conclusion that the simplest solution to this problem is to simply not delete all of the specs, but to process them inline. Basically back to approach #1 as outlined by:

https://github.com/openshift/ansible-service-broker/pull/997#discussion_r198575908

  1. Fetch the specs from a registry
  1. Fetch the specs from the datastore
  1. Delete any specs in the datastore currently marked for deletion, assuming there is no instance that has a reference to this spec which is marked for deletion.
     i. if there is a reference, leave the spec alone
     ii. if there is no reference, delete the spec
  1. For each spec from the registry, get the spec from the datastore
     i. if found, update it, and save to datastore
     ii. if not found, add it to the datastore
  1. For each spec in the datastore, NOT in registry list, mark it for deletion.

This removes all of the concurrency issues we were trying to "solve" in all of the calls. It's easy to understand. Because we will be doing the handling of the delete (#3 above) in-place this will cause the
bootstrap to take longer.

We can make the bootstrap endpoint async so that it won't be affected by how long the bootstrap takes to reconcile. The topic of adding a last_operation like endpoint for bootstrap came up, but for now, we are forgoing that portion of the solution. The last_operation is only of interest to clients that interact with the broker directly, for instance, sbcli. We can implement the same polling mechanism that the catalog uses for the standard OSB methods.

The above will fix the original bugs we outlined, without too much intrusion. We didn't go too deep into backwards compatibility with existing clients, which seems to only affect apb tool. Apb tool is being replaced by sbcli anyway.

The plan is as follows:

1. send out this email
1. add the above solution as a comment to the proposal PR
1. close proposal PR
1. create trello card with above steps
1. @alaypatel07  will implement the solution from the trello card

If anyone has any strong objections please speak up now.

PR #1001, opened 2018-07-03
Add IRC meeting notes

jmrodri:
We were capturing meeting minutes when we started then we just stopped. And now that we are officially holding weekly IRC meetings, seemed like a good idea to share the logs in the repo.


jmrodri:
These are RAW files generated by supybot. The intent is to simply be able to copy them here and submit a PR for future meetings, not to format them as markdown.

PR #1002, opened 2018-07-06
update for quay adapter

johnkim76:
required for quay adapter in [bundle-lib](https://github.com/automationbroker/bundle-lib/pull/75)

 - added a `token` field in the `registry` section of the `broker_config`

depends-on [bundle-lib PR](https://github.com/automationbroker/bundle-lib/pull/75)

[Trello Card](https://trello.com/c/EgOcdkNp)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -240,10 +240,32 @@ registry:
       - ".*-apb$"
 ```
 
+
 If the registry requires authentication for pulling images, this can be achieved by running the following command on every node in your existing cluster:
 
 ```bash
 docker --config=/var/lib/origin/.docker login -u <registry-user> -p <registry-password> <registry_url>
+```
+
+### Quay Registry
+
+Using the Quay registry will allow you to load APBs that are published to the [CoreOS Quay Registry](https://quay.io/about/).  If an authentication `token` is provided, private repositories (*which the token is configured to access*) will also load as well as any public repositories in the specified `org`.
+
+```yaml
+registry:
+  - name: quay_reg
+    type: quay
+    url:  https://quay.io
+    token: <for_private_repos>
+    org: <your_org>
+    white_list:
+      - ".*-apb$"
+```
+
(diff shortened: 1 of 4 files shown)


docs/config.md
```diff
@@ -207,6 +207,27 @@ The partner registry requires authentication for pulling images.  This can be ac
...
+  - name: quay_reg
+    type: quay
+    url:  https://quay.io
+    token: <for_private_repos>
+    org: <your_org>
+    white_list:
+      - ".*-apb$"
+```
+
+The quay registry requires authentication for pulling images.  This can be achieved by running the following command on every node in your existing OpenShift cluster:
```
jmrodri: Feels like this should be `Quay` instead of quay.

docs/config.md
```diff
@@ -207,6 +207,27 @@ The partner registry requires authentication for pulling images.  This can be ac
 docker --config=/var/lib/origin/.docker login -u <registry-user> -p <registry-password> registry.connect.redhat.com
 ```
 
+### Quay Registry
+
+Using the Quay registry will allow you to load APBs that are published to the [CoreOS Quay Registry](https://quay.io/about/).  If an authentication `token` is provided, private repositories (*which the token is configured to access*) will also load as well as any public repositories in the specified `org`.
+
+```yaml
+registry:
+  - name: quay_reg
```
djzager: This looks like something you could add to the broker-apb if you wish, kind of like we do for dockerhub [here](https://github.com/openshift/ansible-service-broker/blob/master/apb/defaults/main.yml#L58-L63). You could also default it to `false` like we do for helm [here](https://github.com/openshift/ansible-service-broker/blob/master/apb/defaults/main.yml#L79).

johnkim76: @djzager I can do this in another PR.  This way, I can include other recently-added-adapters to that list as well.

johnkim76: @djzager actually, I've just updated it here.  It doesn't make sense to add the others as default

PR #1003, opened 2018-07-06
Document API v2 adapter usage 

johnkim76:
The [bundle-lib PR](https://github.com/automationbroker/bundle-lib/pull/109) allows the user to configure an API v2 compatible registry via the `apiv2` adapter

depends-on [bundle-lib PR 109](https://github.com/automationbroker/bundle-lib/pull/109)


docs/config.md
```diff
@@ -207,6 +207,27 @@ The partner registry requires authentication for pulling images.  This can be ac
 docker --config=/var/lib/origin/.docker login -u <registry-user> -p <registry-password> registry.connect.redhat.com
 ```
 
+### API V2 Registry
+
+Registries that implements the Docker [Registry HTTP API V2](https://docs.docker.com/registry/spec/api/) protocol can be configured with the `apiv2` registry adapter as follows
+
+```yaml
+registry:
+  - name: jfrog
```
mhrivnak: Rather than suggest a specific name, I'd make this a pseudo-template like the values below. That would make it clear that the user can supply their own value.

docs/config.md
```diff
@@ -207,6 +207,27 @@ The partner registry requires authentication for pulling images.  This can be ac
...
+    type: apiv2
+    url:  <registry_url>
+    user: <registry-user>
+    pass: <registry-password>
+    white_list:
+      - ".*-apb$"
+```
+
+If the registry requires authentication for pulling images.  This can be achieved by running the following command on every node in your existing OpenShift cluster:
```
mhrivnak: s/.  This/, this/

mhrivnak: Is this OpenShift-specific? Or would this apply to other kubernetes clusters too?

johnkim76:
@mhrivnak I've made the changes you've suggested

PR #1008, opened 2018-07-11
Enhance spec reconciliation, using async bootstrap

alaypatel07:
The broker now performs reconciliation of specs as per [this](https://www.redhat.com/archives/ansible-service-broker/2018-July/msg00003.html)

 - Change Bootstrap function to implement reconciliation alrorithm
 - Change Catalog function to implement filter out specs which are marked for deletion
 - Change Provision function to error out specs which are marked for deletion

#### The PR Depends on:
-  [#137](https://github.com/automationbroker/bundle-lib/pull/137)

 - [Bug 1583495](https://bugzilla.redhat.com/show_bug.cgi?id=1583495): can not perform certain actions when asb loses connection to registry
 - [Bug 1577810](https://bugzilla.redhat.com/show_bug.cgi?id=1583495):  can not perform certain actions when asb loses connection to registry
 - [Bug 1586345](https://bugzilla.redhat.com/show_bug.cgi?id=1586345): applications provisioned with APB cannot update with 3.9 apbs


pkg/dao/crd/dao.go
```diff
@@ -100,6 +100,25 @@ func (d *Dao) SetSpec(id string, spec *bundle.Spec) error {
 	return err
 }
 
+// updateSpec - update a spec for an id in the kvp API and overwrite the
+// existing spec.
+func (d *Dao) updateSpec(id string, spec *bundle.Spec) error {
```
shawn-hurley: Regarding the other comment about not having the update, I think that this should be apart of SetSpec.

We already override SetXXXX to include updates with both instances. I think that we should probably just do the same thing here.

pkg/dao/dao.go
```diff
@@ -49,9 +49,15 @@ type Dao interface {
 	// BatchSetSpecs - set specs based on SpecManifest in the kvp API.
 	BatchSetSpecs(bundle.SpecManifest) error
 
+	// BatchUpdateSpecs - update specs based on SpecManifest in kvp API
+	BatchUpdateSpecs(bundle.SpecManifest) error
```
shawn-hurley: I think that we should not add this method to the dao interface.

pkg/handler/handler.go
```diff
@@ -198,8 +198,13 @@ func NewHandler(b broker.Broker, brokerConfig *config.Config, prefix string,
 func (h handler) bootstrap(w http.ResponseWriter, r *http.Request, params map[string]string) {
 	defer r.Body.Close()
 	h.printRequest(r)
-	resp, err := h.broker.Bootstrap()
-	writeDefaultResponse(w, http.StatusOK, resp, err)
+	go func() {
+		h.broker.Bootstrap()
```
shawn-hurley: IMO we should at least log the error here if bootstrap fails

pkg/broker/broker.go
```diff
@@ -261,16 +305,57 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	bundleInstances, err := a.dao.BatchGetBundleInstances()
+	if err != nil {
+		log.Errorf("error getting bundle instances '%+v'", err)
+		// returning nil because the instead of checking the error,
+		// the broker can simply ignore to delete the specs and hope
+		// that the error wont repeat during next bootstrap cycle
+		return nil
+	}
+	log.Debugf("markedSpecs: %+v\n", markedSpecs)
+	for _, bundleInstance := range bundleInstances {
```
shawn-hurley: We could be going through this loop even after all the specs have been removed from the map. 

Could we add a check at the bottom that could then break out of the loop?

pkg/broker/broker.go
```diff
@@ -241,18 +244,59 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+		if _, ok := specManifest[specID]; !ok {
+			log.Infof("spec '%v' marked for deletion", specID)
+			spec.MarkDelete()
+			markedSpecs[specID] = spec
+		}
+	}
+	return markedSpecs
+}
+
+func getNewAndUpdatedSpecs(daoSpecs map[string]*bundle.Spec, specs []*bundle.Spec) (bundle.SpecManifest,
```
shawn-hurley: We shouldn't need this function if we remove the update specs from the DAO

pkg/broker/broker.go
```diff
@@ -194,33 +194,35 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
 	var specs []*bundle.Spec
 	var imageCount int
 
-	// Remove all non apb-push sourced specs that have been saved.
-	pushedSpecs := []*bundle.Spec{}
```
shawn-hurley: @dymurray Can we make a follow up issue to delete the apb endpoints? Can we offically deprecate and delete them as we move to the new CLI?

pkg/broker/broker.go
```diff
@@ -238,21 +240,62 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	}
+	return markedSpecs
+}
+
+func getSpecManifest(daoSpecs map[string]*bundle.Spec, specs []*bundle.Spec) bundle.SpecManifest {
+	specManifest := make(map[string]*bundle.Spec)
 
 	for _, s := range specs {
-		specManifest[s.ID] = s
+		// If condition is just for logging. It is useful information
```
shawn-hurley: 👍

pkg/broker/broker.go
```diff
@@ -927,20 +1021,20 @@ func (a AnsibleBroker) Bind(instance bundle.ServiceInstance, bindingUUID uuid.UU
 			// unknown error
 			case err != nil && !a.dao.IsNotFoundError(err):
 				return nil, false, err
-			// If there is a job in "succeeded" state, or no job at all, or
-			// the referenced job no longer exists (we assume it got
-			// cleaned up eventually), assume everything is complete.
+				// If there is a job in "succeeded" state, or no job at all, or
```
shawn-hurley: I think that this should be inline with the case statement

pkg/dao/mock_dao.go
```diff
@@ -47,6 +47,11 @@ func (_m *MockDao) BatchGetSpecs(_a0 string) ([]*apb.Spec, error) {
 	return r0, r1
 }
 
+// BatchGetBundleInstances a noop
+func (_m *MockDao) BatchGetBundleInstances() ([]*apb.ServiceInstance, error) {
+	return nil, nil
```
shawn-hurley: I think that this is not correct. I think you may need to re-generate this mock. https://github.com/vektra/mockery

pkg/metrics/metrics.go
```diff
@@ -33,19 +33,21 @@ var (
...
-			Help:      "Specs loaded from registries, partitioned by registry name.",
-		}, []string{"registry_name"})
-
-	specsReset = prometheus.NewCounter(
-		prometheus.CounterOpts{
-			Subsystem: subsystem,
-			Name:      "specs_reset",
-			Help:      "Counter of how many times the specs have been reset.",
-		})
+			Name:      "source",
```
shawn-hurley: I think the name should be something like `specs_total`. https://prometheus.io/docs/practices/naming/

pkg/broker/broker.go
```diff
@@ -261,16 +304,57 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	return specManifest
+}
+
+// getSafeToDeleteSpecs - will check if any bundle instance spe
+func getSafeToDeleteSpecs(a AnsibleBroker, markedSpecs map[string]*bundle.Spec) []*bundle.Spec {
+	safeToDeleteSpecs := make([]*bundle.Spec, 0)
+	bundleInstances, err := a.dao.BatchGetBundleInstances()
+	if err != nil {
+		log.Errorf("error getting bundle instances '%+v'", err)
+		// returning nil because the instead of checking the error,
```
jmrodri: rephrase: 'returning nil because the instead' should read 'returning nil instead of ...'

pkg/broker/broker.go
```diff
@@ -550,6 +636,11 @@ func (a AnsibleBroker) Provision(instanceUUID uuid.UUID, req *ProvisionRequest,
 		// otherwise unknown error bubble it up
 		return nil, err
 	}
+	// If the spec is marked for Deletion, send the same
+	// error as Spec Not Found
+	if spec.Delete {
+		return nil, ErrorNotFound
```
jmrodri: +1

pkg/dao/crd/dao.go
```diff
@@ -143,6 +160,32 @@ func (d *Dao) BatchGetSpecs(dir string) ([]*bundle.Spec, error) {
 	return specs, nil
 }
 
+//BatchGetBundleInstances - get list of bundleinstances
```
jmrodri: should have a space after `//`

pkg/dao/etcd/dao.go
```diff
@@ -149,6 +151,28 @@ func (d *Dao) BatchGetSpecs(dir string) ([]*bundle.Spec, error) {
 	return specs, nil
 }
 
+//BatchGetBundleInstances - get list of bundleinstances
```
jmrodri: space after `//`

pkg/dao/mock_dao.go
jmrodri: run `make generate` and it will do it iirc :)

alaypatel07: I have generated tests using mockery in subsequent [commit](https://github.com/openshift/ansible-service-broker/pull/1008/commits/35929277d4fa1e8539c592f9add0727dcf07f44e)

pkg/broker/broker.go
alaypatel07: I think that too, I think my IDE changed that while formatting code, will re-change that and hate ide just a little bit...

pkg/metrics/metrics.go
alaypatel07: Good catch @shawn-hurley, I did not intend it to be this way. Also, I think, we are staying consistent in giving variable name and CounterOpts Name param, I am renaming variable `specs` to `specsTotal` (along with the change you suggested ofcourse)

pkg/broker/broker.go
```diff
@@ -194,33 +194,35 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	if err := a.dao.BatchDeleteSpecs(unwantedSpecs); err != nil {
+		log.Errorf("Something went real bad trying to delete batch specs... - %v", err)
+		return nil, err
 	}
 
-	err = a.dao.BatchDeleteSpecs(specs)
+	log.Infof("%v specs deleted", len(unwantedSpecs))
+	metrics.SpecsDeleted(len(unwantedSpecs))
+
+	// Getting specs again so that deleted specs does not end up in further comparisons
```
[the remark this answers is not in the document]
alaypatel07: done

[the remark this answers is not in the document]
alaypatel07: done

pkg/broker/broker.go
```diff
@@ -238,21 +240,62 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+	markedSpecs = markSpecsForDeletion(daoSpecs, specManifest)
+
+	metrics.SpecsMarkedForDeletion(len(markedSpecs))
+
+	// Update specs in data-store
+	if err := a.dao.BatchSetSpecs(specManifest); err != nil {
+		return nil, err
+	}
+
+	//Update specs that are marked for deletion
```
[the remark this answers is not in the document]
alaypatel07: done

pkg/broker/broker.go
```diff
@@ -261,16 +304,57 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
 			}
-			planNameManifest[p.ID] = p.Name
 		}
 	}
-	if err := a.dao.BatchSetSpecs(specManifest); err != nil {
-		return nil, err
+	return specManifest
+}
+
+// getSafeToDeleteSpecs - will check if any bundle instance spe
```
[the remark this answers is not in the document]
alaypatel07: changed

pkg/broker/broker.go
alaypatel07: done

pkg/broker/broker.go
```diff
@@ -261,16 +304,57 @@ func (a AnsibleBroker) Bootstrap() (*BootstrapResponse, error) {
...
+func getSafeToDeleteSpecs(a AnsibleBroker, markedSpecs map[string]*bundle.Spec) []*bundle.Spec {
+	safeToDeleteSpecs := make([]*bundle.Spec, 0)
+	bundleInstances, err := a.dao.BatchGetBundleInstances()
+	if err != nil {
+		log.Errorf("error getting bundle instances '%+v'", err)
+		// returning nil because the instead of checking the error,
+		// the broker can simply ignore to delete the specs and hope
+		// that the error wont repeat during next bootstrap cycle
```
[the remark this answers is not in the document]
alaypatel07: done

pkg/broker/broker.go
alaypatel07: Added a condition to break out if the markedSpec map is empty.

pkg/dao/crd/dao.go
alaypatel07: done

pkg/dao/etcd/dao.go
alaypatel07: done

pkg/dao/mock_dao.go
alaypatel07: done. @jmrodri I ran `make generate` but it imported `apb` as `bundle`, renamed that to `apb` to reduce number of lines in git diff.

pkg/metrics/metrics.go
```diff
@@ -109,42 +111,22 @@ func recoverMetricPanic() {
...
-// SandboxDeleted - Counter for how many sandbox deleted.
-func SandboxDeleted() {
-	defer recoverMetricPanic()
-	sandbox.Dec()
-}
-
-// SpecsLoaded - Will add the count of specs. (The value can be negative,
-// resulting in a decrease of the specs loaded).
+// SpecsLoaded - Will add the count of specs.
 func SpecsLoaded(registryName string, specCount int) {
```
shawn-hurley: Should we consider re-naming this function to match the new name? `specsTotal`?

shawn-hurley:
One last question @alaypatel07 have you left a broker with this change running for a long period of time? I don't see any reason it would be a problem but think it might be a test worth running?

alaypatel07:
@shawn-hurley, the instance I have running right now is 3 days old and is still healthy i.e. it is reconciling specs as expected.

alaypatel07:
@eriknelson thanks for the suggestion, changed the PR name 👍

pkg/broker/broker.go
dymurray: +1

PR #1011, opened 2018-07-12
remove golang macros

jmontleon:
Downstream we have golang 1.10 in the buildroot. In 1.10 on Fedora the go macros were broken out into a separate package and golang has a dependency on this package (go-srpm-macros). Downstream in our buildroot this package does not exist and golang has no such dependency so we need to define/remove macros ourselves to get this built.


djzager:
The test that was cancelled was allowed to fail. This LGTM.

PR #1015, opened 2018-07-19
Add ns broker support to the broker's apb

eriknelson:
Adds a switch that will install the broker as a namespaced broker into the `broker_namespace` specified. This will only function if the catalog running has the feature gate `NamespacedServiceBroker` set to true.

### DIFF
--- a/apb/templates/broker-access.clusterrole.yaml
+++ b/apb/templates/broker-access.clusterrole.yaml
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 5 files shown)


apb/templates/broker-access.clusterrole.yaml
```diff
@@ -3,11 +3,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: access-{{ broker_name }}-role
+  name: access-{{ broker_namespace }}-role
```
djzager: This is the only change that scares me.

PR #1024, opened 2018-08-02
Bug 1585951 - Delete provision credentials on deprovision from ASB namespace

eriknelson:
If credentials are extracted from a provision, a secret is created with their contents in the ASB namespace. This secret is not being cleaned up when the corresponding instance is deprovisioned.

Delete the secret containing provision credentials if it exists during a deprovision.

### DIFF
--- a/pkg/broker/broker.go
+++ b/pkg/broker/broker.go
@@ -865,7 +865,15 @@ func (a AnsibleBroker) Deprovision(
 		if err := a.engine.StartNewSyncJob(token, dpjob, DeprovisionTopic); err != nil {
 			return nil, err
 		}
+
+		// Attempt to delete extracted credentials created from provision. The creds
+		// should only be deleted if the instance was successfully deleted, otherwise
+		// we will end up in a corrupt state.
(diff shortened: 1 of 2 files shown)


eriknelson:
Not sure how coverage went up here :smile:, but will look into adding a test.

pkg/broker/broker.go
```diff
@@ -865,7 +865,15 @@ func (a AnsibleBroker) Deprovision(
 		if err := a.engine.StartNewSyncJob(token, dpjob, DeprovisionTopic); err != nil {
 			return nil, err
 		}
+
+		// Attempt to delete extracted credentials created from provision. The creds
+		// should only be deleted if the instance was successfully deleted, otherwise
+		// we will end up in a corrupt state.
+		if err := bundle.DeleteExtractedCredentials(instance.ID.String()); err != nil {
+			log.Infof("Attempted to delete extracted credentials from a provision but could not: %s", err.Error())
```
jmrodri: That's very elegant wording :)

PR #1032, opened 2018-08-07
Bug 1583503 Always run updates

jmontleon:
Changes updates to always run, regardless of whether parameters have changed.

This behavior change is in response to BZ 1583503. We previously blindly assumed that if the parameters have not changed since the last request that the last request succeeded, which we simply cannot guarantee.

Since David has done work to ensure our APB updates are idempotent it should be acceptable to rerun updates without changes. But it will probably make it a hard requirement rather than a polite suggestion that update tasks be idempotent.

fixes BZ 1583503


jmontleon:
added do-not-merge label until I can do a little more testing.

jmontleon:
Worked doing a normal update, a no change update, and about as well as hoped with an erroring update.

PR #1038, opened 2018-08-09
fix coverage to fail when tests fail

jmrodri:
Tests were broken for a while now, but the Makefile was hiding the error. This will now let it show failures. We had a similar problem with bundle-lib a while back. Fixed that with PR https://github.com/automationbroker/bundle-lib/pull/145


jmrodri:
This PR will always stay failed until PR #1039 gets in.

PR #1043, opened 2018-08-14
Fix broker route to line up with service

djzager:


jmontleon:
ACK

apb/templates/broker.route.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
 spec:
```
dymurray: I think we need to update line 12 to `broker` as well for the label.

PR #1044, opened 2018-08-14
Remove APB spec from broker-apb

djzager:


apb/apb.yml
```diff
@@ -1,34 +0,0 @@
...
-  - name: main
-    description: Manage the automation broker in a cluster
-    free: "True"
-    metadata:
-      displayName: Default
-      longDescription:
-        This plan deploys the automation broker
-      cost: $0.00
-    parameters:
-      - name: broker_name
```
dymurray: Additonal vars:
```
broker_kind
```
remove
```
broker_namespace -> namespace
```

PR #1045, opened 2018-08-14
Update bundle-lib

djzager:
Bring a bunch of bundle-lib bugfixes into the broker.


jmrodri:
once travis is green we can merge.

PR #1046, opened 2018-08-15
Fix broker install yaml for release-1.2

djzager:


apb/install.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       serviceAccount: automation-broker-apb
       containers:
         - name: apb
-          image: docker.io/automationbroker/automation-broker-apb:latest
+          image: docker.io/automationbroker/automation-broker-apb:release-1.2
```
matzew: +1

PR #1048, opened 2018-08-16
Bug 1615787 - Add APB Spec to Broker APB

djzager:


downstream.patch
```diff
@@ -0,0 +1,44 @@
...
+   providerDisplayName: "Red Hat, Inc."
+ plans:
+diff --git a/apb/defaults/main.yml b/apb/defaults/main.yml
+index 0ff227b3..7cf17b66 100644
+--- a/apb/defaults/main.yml
++++ b/apb/defaults/main.yml
+@@ -12,7 +12,7 @@ wait_for_broker: false # By default, we do not wait for the broker to be availab
+ 
+ # The image to use for the broker
+ broker_image_tag: latest
```
jmontleon: I think the desire is to have this match the release (v3.11, for instance) rather than latest so as additional releases come along they get a broker in line with the correct release.

PR #1050, opened 2018-08-17
Bug 1610714 - pass bindingid to unbind jobs

jmrodri:
We were not passing the bindingid to the unbind job, so it would default
to the instance id and couldn't find the bindings to delete.

Also added a test to ensure that the unbindjob gets the binding id


jmrodri:
Fixes Bug https://bugzilla.redhat.com/show_bug.cgi?id=1610714

PR #1053, opened 2018-08-21
Fix broker rpm spec

djzager:
I had duplicated prep sections.

### DIFF
--- a/ansible-service-broker.spec
+++ b/ansible-service-broker.spec
@@ -15,11 +15,11 @@
 %if 0%{?with_debug}
 %global _dwz_low_mem_die_limit 0
 %else
-%global	debug_package	%{nil}
+%global debug_package %{nil}
 %endif
 
-%global	provider github
-%global	provider_tld com
+%global provider github
+%global provider_tld com
 %global project openshift
 %global repo ansible-service-broker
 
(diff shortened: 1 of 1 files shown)


ansible-service-broker.spec
```diff
@@ -155,6 +149,11 @@ unit-test for %{name}
 %endif
 
 %prep
+%setup -q -n %{name}-%{version}
```
dymurray: Seems strange to me that we call this line here, and again at line 157

djzager:
Before:

```
$ rpmlint -i ansible-service-broker.spec
ansible-service-broker.spec:50: W: macro-in-comment %{arm}
There is a unescaped macro after a shell style comment in the specfile. Macros
are expanded everywhere, so check if it can cause a problem in this case and
escape the macro with another leading % if appropriate.

ansible-service-broker.spec:217: W: mixed-use-of-spaces-and-tabs (spaces: line 217, tab: line 18)
The specfile mixes use of spaces and tabs for indentation, which is a cosmetic
annoyance.  Use either spaces or tabs for indentation, not both.

ansible-service-broker.spec: E: specfile-error warning: Macro expanded in comment on line 50: %{arm}}
This error occurred when rpmlint used rpm to query the specfile.  The error is
output by rpm and the message should contain more information.

ansible-service-broker.spec: E: specfile-error
This error occurred when rpmlint used rpm to query the specfile.  The error is
output by rpm and the message should contain more information.

ansible-service-broker.spec: E: specfile-error error: line 157: second %prep
This error occurred when rpmlint used rpm to query the specfile.  The error is
output by rpm and the message should contain more information.

ansible-service-broker.spec: E: specfile-error error: query of specfile ansible-service-broker.spec failed, can't parse
This error occurred when rpmlint used rpm to query the specfile.  The error is
output by rpm and the message should contain more information.

0 packages and 1 specfiles checked; 4 errors, 2 warnings.
```

After:

```
$ rpmlint -i ansible-service-broker.spec
0 packages and 1 specfiles checked; 0 errors, 0 warnings.
```

jmrodri:
Looks okay, it's a little weird that `%prep` is so far down in the spec file, but it seems to work.

PR #1057, opened 2018-08-22
Bug 1615787 - Make a registry configurable via apb

djzager:


djzager:
I want to test this but this is my attempt at making one registry configurable via the APB spec.

PR #1058, opened 2018-08-22
Make downstream patch applicable

djzager:


djzager:
I'm guessing that the changes that I made for the apb.yml support made it so the downstream.patch wouldn't apply.

Issue #1056, opened 2018-08-22
Cluster admin role is required to run any APB on Minishift / OKD


ibuziuk:
## Bug: Cluster admin role is required to run any APB on Minishift / OKD

**What happened**: When trying to run any APB e.g. mongo / postgreSQL provisioned services are failed to start with:

`The service is not yet ready. Error provisioning ServiceInstance of ClusterServiceClass (K8S: "e9c042c4925dd0c7c25ceca4f5179e1c" ExternalName: "dh-mongodb-apb") at ClusterServiceBroker "openshift-automation-service-broker": Status: 403; ErrorMessage: <nil>; Description: User does not have sufficient permissions; ResponseError: <nil>`

![image](https://user-images.githubusercontent.com/1461122/44454958-23c19c00-a5fd-11e8-9612-d1aa753ea55a.png)

And only after granting cluster admin rights APB is started succesfully:

`oc adm policy add-cluster-role-to-user cluster-admin developer`

**What you expected to happen**: it is expected that the default user without cluster admin priviliges would be to run APB

**How to reproduce it**: 
Originally the issue has been reproduced for che apb - https://github.com/ansibleplaybookbundle/eclipse-che-apb/issues/19
with the default configuration - https://github.com/ansibleplaybookbundle/eclipse-che-apb#requirements
`canary` tag was used in order to get the latest version of APB

eivantsov:
I can reproduce it with Origin 3.9 + ansible service broker deployed as https://github.com/openshift/ansible-service-broker/blob/master/scripts/run_latest_build.sh, and OKD 3.10 (--enable=ansible-service-broker)

ibuziuk:
@dymurray @djzager could you please comment, if it is expected that cluster admin role is required for user to run any APB? AFAIK, it used to work with the default privileges

djzager:
@ibuziuk the short answer is that the `cluster-admin` role is not required for running APBs via the service catalog UI.

ibuziuk:
@djzager thank you for looking into this:

1. minishift v1.20.0+53c500a
2. 3.10 / @eivantsov tried 3.9 and it also works only with cluster admin rights
3. 

> export MINISHIFT_ENABLE_EXPERIMENTAL=y && \
> minishift config set iso-url centos && \
> minishift start --memory=8gb --cpus=4 --disk-size=50g \
>                 --network-nameserver 8.8.8.8 \
>                 --openshift-version 3.10.0-rc.0 \
[...]

PR #1059, opened 2018-08-23
Attempt to remove dependency on cluster version in broker apb

djzager:
Fixes: #1056


apb/defaults/main.yml
```diff
@@ -44,7 +44,7 @@ broker_refresh_interval: "600s"
 broker_launch_apb_on_bind: "false"
 broker_output_request: "true"
 broker_recovery: "true"
-broker_auto_escalate: ""
+broker_auto_escalate: "true"
```
jwmatthews: Before we merge this I would like to ensure team wants this to 'true' for upstream.
I saw the downstream patch is setting it to 'false'.

In general I don't like deviating behavior from downstream/upstream unless we have a clear reason for doing so.

CC:  @jmrodri @shawn-hurley @eriknelson @rthallisey

djzager: Just to add the reason why I am defaulting upstream to false revolves around the fact that the service catalog has the origin identity header behind a feature gate. This feature gate is not enabled by default in their helm chart. Furthermore, the service catalog has no interrogation mechanism (ie. Ask if that feature is enabled).

djzager: Just to add the reason why I am defaulting upstream to false revolves around the fact that the service catalog has the origin identity header behind a feature gate. This feature gate is not enabled by default in their helm chart. Furthermore, the service catalog has no interrogation mechanism (ie. Ask if that feature is enabled).

jmrodri: Maybe we just document that if Service Catalog does not have the feature gate enabled they MUST edit the configmap to enable auto-escalate.

djzager: I think if we really want to leave auto_escalate set to false, then we update the log messages to say something like "hey, we didn't get an origin identity header. Either enable it in the service catalog or set auto_escalate to true in the broker config."

PR #1060, opened 2018-08-27
Check for capabilities instead of openshift during APB install

jmontleon:
Fixes: #1056 / BZ 1621909


apb/tasks/main.yml
```diff
@@ -1,19 +1,11 @@
...
-# These can always be overridden at the command line
-# http://docs.ansible.com/ansible/devel/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable
 - name: 'Set facts'
   set_fact:
-    cluster: "{{ _cluster_type }}"
-    broker_auto_escalate: "{{ 'true' if _cluster_type == 'kubernetes' else 'false' }}"
-    broker_local_openshift_enabled: "{{ 'true' if _cluster_type == 'openshift' else 'false' }}"
-  vars:
-    _cluster_type: '{{ "openshift" if "openshift" in lookup("k8s", cluster_info="version") else "kubernetes" }}'
+    broker_auto_escalate: "{{ 'true' if 'project.openshift.io' not in lookup('k8s', cluster_info='api_groups') else 'false' }}"
```
jwmatthews: Based on https://github.com/kubernetes-incubator/service-catalog/commit/97f79ce9088b55cfe153328d171b7239f5464a58

Even in k8s the helm chart will enable the originating identity header now:
https://github.com/kubernetes-incubator/service-catalog/blob/master/charts/catalog/values.yaml#L165

So should we just default this to 'false' for both cases now?

eriknelson: > So should we just default this to 'false' for both cases now?

This is what I was thinking, also loudly warning if that header is missing and auto_escalate isn't specified.

apb/defaults/main.yml
```diff
@@ -44,7 +44,7 @@ broker_refresh_interval: "600s"
 broker_launch_apb_on_bind: "false"
 broker_output_request: "true"
 broker_recovery: "true"
-broker_auto_escalate: ""
+broker_auto_escalate: "false"
```
jwmatthews: +1

jwmatthews:
Based on a recent change to upstream service-catalog (https://github.com/kubernetes-incubator/service-catalog/commit/97f79ce9088b55cfe153328d171b7239f5464a58) the originating user header will be enabled by default now.

So we will plan to set broker_auto_escalate to false and not bother to check if it's openshift or k8s

dymurray:
@jmontleon Tried to fix the merge conflict but wasn't sure about indexes in patch files... will sync up in the morning.

PR #1064, opened 2018-08-28
use the new osb prefix

jmrodri:
PR #1029 standardized the urls to be `osb/` so this updates the test scripts to use the same paths.


eriknelson:
Merging this; the ci failure is unrelated and these are just test script updates.

PR #1078, opened 2018-09-11
Add broker operator support

eriknelson:
Adds the ansible operator to the broker's repo, mostly based on @fabianvf's initial work in #1047 

- [X] Updates the broker's ansible code to a portable role with some refactoring that will make it easier to deploy alongside an ansible-operator. The APB should continue to use this role as well.
- [X] Adds additional files required for deploying the broker via operator


eriknelson:
@fabianvf think I touched on your points. I'm able to actually deploy a healthy broker running create on `ansible_role/{crd,rbac,operator,cr.example}.yaml`, so that's a pretty  good start. The ansible-operator seems to be ignoring the update when I edit the CR though. It doesn't seem to trigger reconcile, unless the 5m resync interval needs to kick in. Will catch up with you guys tomorrow about that.

fabianvf:
@eriknelson, are you editing the config map directly or editing the broker CR?

Makefile
```diff
@@ -95,15 +100,18 @@ build-image: ## Build the broker (from canary)
...
-	docker build -f ${APB_DIR}/Dockerfile --build-arg VERSION=${TAG} --build-arg APB=${TAG} -t ${APB_IMAGE} ${APB_DIR}
+	docker build -f ${APB_DIR}/Dockerfile --build-arg VERSION=${TAG} --build-arg APB=${TAG} -t ${APB_IMAGE} ${ANSIBLE_ROLE_DIR}
 else
-	docker build -f ${APB_DIR}/Dockerfile --build-arg VERSION=${TAG} -t ${APB_IMAGE} ${APB_DIR}
+	docker build -f ${APB_DIR}/Dockerfile --build-arg VERSION=${TAG} -t ${APB_IMAGE} ${ANSIBLE_ROLE_DIR}
 endif
 
+build-operator: ## Build the broker operator image
+	docker build -f ${OPERATOR_DIR}/Dockerfile -t ${OPERATOR_IMAGE} ${ANSIBLE_ROLE_DIR}
+
```
jmrodri: Makefile looks okay. The travis logs show that the `deploy.sh` is trying to reference `apb/install.yaml` which looks like it moved.

ansible_role/operator/deploy.yaml
```diff
@@ -14,5 +14,5 @@ spec:
     spec:
       containers:
         - name: automation-broker-operator
-          image: docker.io/eriknelson/automation-broker-operator:testing
+          image: docker.io/automationbroker/automation-broker-operator
```
jmrodri: HAHAHA

ansible_role/defaults/main.yml
```diff
@@ -55,6 +66,7 @@ broker_keep_namespace: "false"
 broker_keep_namespace_on_error: "false"
 
 # Special broker config options
+broker_log_level: "debug"
```
fabianvf: Should the default log level be debug?

ansible_role/operator/Dockerfile-olm
```diff
@@ -0,0 +1,8 @@
+FROM quay.io/water-hole/ansible-operator
+
+COPY . /opt/ansible/roles/automation-broker
+COPY playbooks/operator.yml /opt/ansible/deploy.yaml
+COPY operator/watches.yaml /opt/ansible/watches.yaml
+
+RUN sed -i "s/olm_managed: false/olm_managed: true/" \
```
fabianvf: Can we parameterize this on the main dockerfile or is this the standard way to handle this?

ansible_role/operator/cr.example.yaml
```diff
@@ -0,0 +1,51 @@
...
+  createBrokerNamespace: false
+  waitForBroker: false
+################################################################################
+# Some other example overrides, in general most defaults will work.
+################################################################################
+  #brokerName: automation-broker
+  #brokerImage: "ansibleplaybookbundle/origin-ansible-service-broker"
+  #brokerClusterRole: admin
+  #brokerNodeSelector: ""
+  #brokerConfig:
```
fabianvf: I would line the comment symbols up for ease of removal

ansible_role/tasks/build_config.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
fabianvf: Ansible yaml files typically end in .yml

fabianvf: This applies to below files as well

ansible_role/tasks/main.yml
```diff
@@ -30,10 +16,52 @@
...
 - include_tasks: "generate_certificate.yaml"
   when: ('route.openshift.io' not in lookup('k8s', cluster_info='api_groups')) | bool
 
+- name: 'Set broker config map state={{ state }}'
+  k8s:
+    state: '{{ state }}'
+    definition: '{{ broker_config_map }}'
+  register: configmap_result
+
+- name: Redeploy the broker if the configmap changed
```
fabianvf: Were you able to get this working when you edited the CR?

ansible_role/operator/cr.example.yaml
eriknelson: Do you mean that they should all be in the first column?

fabianvf: That would work, I just usually make them line up somewhere so you can vertically select a chunk

ansible_role/defaults/main.yml
eriknelson: This is preserving existing behavior; seems reasonable to me for upstream to have this IMO.

ansible_role/tasks/main.yml
eriknelson: Yup, edited the CR with some new configmap vals, they propogate through and the operator bounces the pod. Works pretty well.

PR #1090, opened 2018-09-19
update rpm spec and add apb Dockerfiles

jmontleon:


ansible_role/apb/Dockerfile-v3.10
```diff
@@ -0,0 +1,74 @@
+FROM ansibleplaybookbundle/apb-base:v3.11
+
+RUN rm -f /etc/yum.repos.d/asb.repo && curl https://copr.fedorainfracloud.org/coprs/g/ansible-service-broker/ansible-service-broker-1.2.0-openshift-3.10/repo/epel-7/group_ansible-service-broker-ansible-service-broker-1.2.0-openshift-3.10-epel-7.repo -o /etc/yum.repos.d/asb.repo
```
jmontleon: I think the apb requires the v3.11 APB Base. Maybe @djzager can confirm. If not we can use v3.10 apb-base.

PR #1102, opened 2018-10-05
Bug 1613280 - only create netpolicy if one already exists

jmrodri:
If there are other services in the target namespace talking to each other with
no network policy, during APB deployments we automatically create one that
blocks the other services. This change looks to see if there are any network
policies on the target namespace. If there are none, we forgo creating the
network policy, assuming things are open and the transient namespace should be
able to talk to the target with no issues. If there are existing network
policies, then we will add ours to give the transient namespace access to the
target. There is still the chance that our network policy could still affect
things depending on the variety of existing network policies in place, too
many to verify.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -186,32 +186,45 @@ func (p provider) CreateSandbox(podName string,
 		}
 	}
 
-	// Must create a Network policy to allow for comunication from the APB pod to the target namespace.
-	networkPolicy := &networkingv1.NetworkPolicy{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: podName,
-		},
-		Spec: networkingv1.NetworkPolicySpec{
-			PodSelector: metav1.LabelSelector{},
-			Ingress: []networkingv1.NetworkPolicyIngressRule{
-				networkingv1.NetworkPolicyIngressRule{
-					From: []networkingv1.NetworkPolicyPeer{
-						networkingv1.NetworkPolicyPeer{
-							NamespaceSelector: metav1.AddLabelToSelector(&metav1.LabelSelector{}, "apb-pod-name", podName),
+	// Check to see if there are already namespaces available before
+	// creating ours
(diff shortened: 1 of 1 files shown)


pkg/runtime/runtime.go
```diff
@@ -186,32 +186,45 @@ func (p provider) CreateSandbox(podName string,
...
-		Spec: networkingv1.NetworkPolicySpec{
-			PodSelector: metav1.LabelSelector{},
-			Ingress: []networkingv1.NetworkPolicyIngressRule{
-				networkingv1.NetworkPolicyIngressRule{
-					From: []networkingv1.NetworkPolicyPeer{
-						networkingv1.NetworkPolicyPeer{
-							NamespaceSelector: metav1.AddLabelToSelector(&metav1.LabelSelector{}, "apb-pod-name", podName),
+	// Check to see if there are already namespaces available before
+	// creating ours
+	policies, err := k8scli.Client.NetworkingV1().NetworkPolicies(targets[0]).List(metav1.ListOptions{})
```
eriknelson: Doesn't look like the existing code did this, but we should make sure len targets > 0 otherwise return an error here to prevent array out of bounds panic. Also, is it sufficient just to look at the first target? Are any of the other targets relevant?

pkg/runtime/runtime.go
```diff
@@ -186,32 +186,45 @@ func (p provider) CreateSandbox(podName string,
...
-		},
-	}
+		}
 
-	log.Debugf("Creating network policy for pod: %v to grant network access to ns: %v", podName, targets[0])
-	_, err = k8scli.Client.NetworkingV1().NetworkPolicies(targets[0]).Create(networkPolicy)
-	if err != nil {
-		log.Errorf("unable to create network policy object - %v", err)
-		return "", err
+		log.Debugf("Creating network policy for pod: %v to grant network access to ns: %v", podName, targets[0])
```
eriknelson: Same question here, just wondering why the other targets aren't relevant (and why we have multiple targets at all?)

PR #1106, opened 2018-10-12
Additional OLM support

eriknelson:
Tweaks to support operators built for OLM

### DIFF
--- a/ansible_role/operator/cr.example.yaml
+++ b/ansible_role/operator/cr.example.yaml
@@ -6,8 +6,26 @@ metadata:
   namespace: automation-broker
 spec:
   brokerImagePullPolicy: IfNotPresent
+  # TODO: Need to default these to false when building for OLM
   createBrokerNamespace: false
   waitForBroker: false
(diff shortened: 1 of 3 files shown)


ansible_role/operator/cr.example.yaml
```diff
@@ -6,8 +6,26 @@ metadata:
...
+  # TODO: Need to default these to false when building for OLM
   createBrokerNamespace: false
   waitForBroker: false
+  # TODO: Currently, this must match the namespace that catasb creates the
+  # admin binding for SA in (at least until operator bundles become available
+  # with OLM
+  # TODO: Want to update the ansible to accept an env var so that the OLM
+  # deployment can actually pass this through, rather than forcing the user
+  # to specify it
+  brokerNamespace: automation-broker
```
djzager: Something like `"brokerNamespace: {{ lookup('env','NAMESPACE') | default('automation-broker', true) }}"` would work in the `ansible_role/defaults/main.yml`

PR #1112, opened 2018-10-19
Update to bundle-lib 0.2.15 to fix net-policy issue

jmrodri:
Fixes Bug 1613280 - only create netpolicy if one already exists

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,47 +3,60 @@
 
 [[projects]]
   branch = "default"
+  digest = "1:c0164b6d52877372590085b45f53a160f47c24e8a2312bb815489223044f5914"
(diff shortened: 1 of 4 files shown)


jmrodri:
I used the following `dep` version:
```
$ dep version
dep:
 version     : v0.5.0
 build date  : 2018-07-26
 git hash    : 224a564
 go version  : go1.10.3
 go compiler : gc
 platform    : linux/amd64
 features    : ImportDuringSolve=false
```

jmontleon:
https://github.com/openshift/ansible-service-broker/blob/release-1.3/apb/templates/broker-auth.clusterrole.yaml#L20 for 1.3, if the same works for master.

PR #1113, opened 2018-10-19
Update to bundle-lib 0.1.9 to fix net-policy issue

jmrodri:
Fixes Bug 1613280 - only create netpolicy if one already exists


jmontleon:
https://github.com/openshift/ansible-service-broker/blob/release-1.2/apb/templates/broker-auth.clusterrole.yaml#L20 if the same works for master.

PR #1126, opened 2018-10-31
Operator should use meta from AO

djzager:
This should make it so that the broker_name and broker_namespace are set by the Ansible Operator passing in the `meta` object as extravars.


jmontleon:
ACK

PR #1130, opened 2018-11-13
First draft of proposal for rate limiting the API

jmrodri:

 - rate limit all or specific APIs
 - keep only one failed namespace instead of all of them

### DIFF
--- a/docs/proposals/proposal-ratelimiter.md
+++ b/docs/proposals/proposal-ratelimiter.md
@@ -32,10 +32,10 @@ able to remove the namespace at the end of the action. Because the names are
 generated it isn't feasible to go searching for all "action" jobs.
 
 I think we need to add a new attribute, `LastSandbox`, to
-`BundleInstanceStatus`. It will be a string containing the last sandbox
-namespace.
+`BundleInstanceStatus` and `BundleBindingStatus`. It will be a string
+containing the last sandbox namespace.
 
-```
+```golang
   // BundleInstanceStatus status is a service instance status.
     Bindings        []LocalObjectReference `json:"bindings"`
     State           State                  `json:"state"`
@@ -43,6 +43,16 @@ namespace.
     Jobs            map[string]Job         `json:"jobs"`
     LastSandbox     string                 `json:"lastSandbox"`
   }
+```
+
+```golang
+// BundleBindingStatus - status of the bundle
+type BundleBindingStatus struct {
+    State           State          `json:"state"`
+    LastDescription string         `json:"lastDescription,omitempty"`
+    Jobs            map[string]Job `json:"jobs"`
+    LastSandbox     string         `json:"lastSandbox"`
+}
 ```
 
 When we start the action, we should set the namespace in the
--- a/docs/proposals/proposal-ratelimiter.md
+++ b/docs/proposals/proposal-ratelimiter.md
@@ -83,10 +83,80 @@ exponentially. We take n calls with no limit, once those n are exhausted we take
 every 3rd call, then we back that off to every 5th call. After m failed attempts
 we stop executing APBs and simply return failure always.
 
-New configuration items:
+At first I thought we'd have a time limit as well but the exponential backoff
+seems like a better idea to apply across the board.  A new configuration item
+includes a list of action handlers we want to limit.
 
 ```yaml
 rate_limit:
-  timeout: in seconds, default to 300
   handler: ["all", "provision", "deprovision", "bind", "unbind" ]
 ```
+
+To be clear, this rate limiter will not control how often the API is called by
+the platform, but it will control how often the broker launches an APB to
+perform the action on failed cases.
+
+#### Rate limited scenario
+
+##### Today
+
+- provision API called
+  - provision APB launched, failed
+  - return failed status to platform
+- deprovision API called first time on above instance
+  - deprovision APB launched, failed
+  - deprovision 1st namespace remains
+  - return failed status to platform
+- deprovision API called second time on above instance
+  - deprovision APB launched, failed
+  - deprovision 2nd namespace remains
+  - return failed status to platform
+  ...
+- deprovision API called nth time on above instance
+  - deprovision APB launched, failed
+  - deprovision nth namespace remains
+  - return failed status to platform
+
+This will continue forever or until the clusters resources are exhausted. The
+only way to stop this is to remove the finalizer on the serviceinstance from the
+platform so that the deprovision stops getting called.
+
+##### Proposed scenario
+
+- provision API called
+  - provision APB launched, failed
+  - return failed status to platform
+- deprovision API called first time on above instance
+  - deprovision APB launched, failed
+  - deprovision 1st namespace remains
+  - return failed status to platform
+- deprovision API called second time on above instance
+  - deprovision APB launched, failed
+  - 1st namespace is removed (based on rules from first portion of proposal)
+  - deprovision 2nd namespace remains (until next call)
+  - return failed status to platform
+- deprovision API called third time on above instance
+  - rate limit rule triggered, skip launching deprovision APB
+  - last deprovision namespace remains
+  - return failed status to platform
+- deprovision API called 4th time on above instance
+  - rate limit rule triggered, skip launching deprovision APB
+  - last deprovision namespace remains
+  - return failed status to platform
+- deprovision API called 5th time on above instance
+  - rate limit rule triggered, skip launching deprovision APB
+  - last deprovision namespace remains
+  - return failed status to platform
+- deprovision API called 6th time on above instance
+  - rate limit rule triggered, ALLOW deprovision APB to run
+  - last deprovision namespace removed
+  - return failed status to platform
+- Repeat the above scenarios a few times.
+  ...
+- deprovision API called nth time on above instance
+  - after a few rounds finally stop launching APBs
+  - rate limit ruled exhausted, skip launching deprovision APB
+  - last deprovision namespace removed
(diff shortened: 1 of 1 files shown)


jmrodri:
@shawn-hurley @djzager okay I believe I'm done with it

jmrodri:
I don't care for `PreviousNamespaces` and I'd be willing to rename it to something else `OldSandboxes`, `PreviousSandboxes`.

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+as first in first out. Newer jobs will append to the list, we simply pop the
+first ones off.
+
+```
+  // BundleInstanceStatus status is a service instance status.
+    Bindings           []LocalObjectReference `json:"bindings"`
+    State              State                  `json:"state"`
+    LastDescription    string                 `json:"lastDescription,omitempty"`
+    Jobs               map[string]Job         `json:"jobs"`
+    PreviousNamespaces []string               `json:"previousNamespaces"`
```
djzager: Could we not just track the `lastNamespace`?

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+here is to minimize wasting resources in the cluster and still allow developers
+& admins to have access to the failed action. During a deprovision & unbind,
+once we get the `ServiceInstance` we will see if there are any
+`PreviousNamespaces`. If there are old ones, we will delete the first one in the
+list, id 0. Before creating a new sandbox namespace.
+
+TBD: not sure if this should be done at the handler level or further down. I'll
+leave that as an implementation detail.
+
+NOTE: we could make the number of namespaces kept to be configurable if users
```
djzager: I don't see a huge value boost here. All the scenarios where I've had a deprovision fail, over and over and over, it was for the exact same failure.

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+The broker will respond to an API call as many times as you call it. Often this
+means launching a new APB which can result in a ton of cluster resources being
+used when the API call was a result of a previous failure. For example, if a
+provision call fails, the platform will likely call a deprovision to prevent any
+dangling resources. If the deprovision fails, the platform will continue to call
+deprovision until it works. If the `KeepNamespaceOnError` is `true`, we will
+keep the namespaces for *every* failed call around wasting precious cluster
+resources.
+
+The proposal above will mitigate the left over namespaces. But if we now a call
```
djzager: `But if we know`

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+will always fail because it has failed the last n times. There really is no
+point to continue to spawn new APBs and namespaces. We should backoff the calls
+exponentially. We take n calls with no limit, once those n are exhausted we take
+every 3rd call, then we back that off to every 5th call. After m failed attempts
+we stop executing APBs and simply return failure always.
+
+At first I thought we'd have a time limit as well but the exponential backoff
```
djzager: I like the exponential backoff.

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+seems like a better idea to apply across the board.  A new configuration item
+includes a list of action handlers we want to limit.
+
+```yaml
```
djzager: Do we need a configuration? Or could we simply have an exponential backoff for subsequent requests on something that previously failed.

Thinking about it, what if in the first go we simply rate limited deprovision. No configuration no special cases. If deprovision fails over and over we simply attempt deprovision less and less often.

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+keep the namespace around if there was an error. This happens on every error.
+We would want to keep only the latest one and get rid of the previously failed
+namespaces. We will need a way to keep track of the namespaces previously used
+for the jobs.
+
+The `Namespace` is stored on the `Context` which allows the current run to be
+able to remote the namespace at the end of the action. Because the names are
+generated it isn't feasible to go searching for all "action" jobs.
+
+I think we need to add a new attribute, `PreviousNamespaces`, to
```
shawn-hurley: Shouldn't Previous Namespace only ever be a single namespace? if we are only keeping 1 old failed namespace, the successful namespace should be deleted already. Thoughts?

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+`BundleInstanceStatus`. It will be an slice of strings. The list will be ordered
```
shawn-hurley: Don't we need the same thing for BindingInstance?

docs/proposals/proposal-ratelimiter.md
shawn-hurley: 👍

docs/proposals/proposal-ratelimiter.md
shawn-hurley: NIT: now -> know

docs/proposals/proposal-ratelimiter.md
```diff
@@ -0,0 +1,162 @@
...
+deprovision until it works. If the `KeepNamespaceOnError` is `true`, we will
+keep the namespaces for *every* failed call around wasting precious cluster
+resources.
+
+The proposal above will mitigate the left over namespaces. But if we now a call
+will always fail because it has failed the last n times. There really is no
+point to continue to spawn new APBs and namespaces. We should backoff the calls
+exponentially. We take n calls with no limit, once those n are exhausted we take
+every 3rd call, then we back that off to every 5th call. After m failed attempts
+we stop executing APBs and simply return failure always.
```
shawn-hurley: I think I would prefer a time here instead of just counting calls. The reason being, the service catalog also has some timing backoff that they are doing, This to me means that we will quickly move our exponential up, and they will move there's up, and eventually you're getting cached results even after waiting minutes. I don't think that is the behavior that we would want. thought?

docs/proposals/proposal-ratelimiter.md
jmrodri: Yeah, I think at some point the idea of offering more than one was in my head. Like I'd like to keep the last 2 or 3. But then I realized I don't think we need that. I'll make it a simple string instead.

docs/proposals/proposal-ratelimiter.md
jmrodri: Ah yes we do. I forgot it was a separate entity.

mhrivnak:
There is one potential gotcha with the proposal to only keep the most recently-failed namespace for a given operation. We keep the namespace so an admin can inspect it and figure out what went wrong. That person's job will be more difficult if namespaces keep getting deleted while the admin is inspecting them.

In theory the admin could add a finalizer to a namespace to ensure it doesn't get deleted mid-troubleshooting, but that doesn't feel like the right user experience.

Would it make sense to only keep the oldest one, and delete any newer namespaces that fail in the same way? Or maybe only keep the oldest and newest, but none in between? Or something else?

shawn-hurley:
> That person's job will be more difficult if namespaces keep getting deleted while the admin is inspecting them.

@mhrivnak I didn't consider that, yeah that makes a lot of sense to me. 

> ould it make sense to only keep the oldest one, and delete any newer namespaces that fail in the same way? Or maybe only keep the oldest and newest, but none in between? Or something else?

I vote to keep the oldest and delete all the rest. I think the only issue is if something got fixed and a new issue pops up, then it might be hard to see. We can fix that by deleting the namespace that is kept and then keeping the next run's namespace though?

docs/proposals/proposal-ratelimiter.md
jmrodri: I called it `lastSandbox` now

docs/proposals/proposal-ratelimiter.md
jmrodri: > There is one potential gotcha with the proposal to only keep the most recently-failed namespace for a given operation. We keep the namespace so an admin can inspect it and figure out what went wrong. That person's job will be more difficult if namespaces keep getting deleted while the admin is inspecting them.

I didn't think of the scenario of the admin looking at the specific namespace that may go away from them. I was kind of under the impression that it is probably failing for the same reason constantly, so looking at the most recent one would suffice. But after reading your comment, I can see how they might be looking at the last failed error and POOF, it's gone. Now they have to restart with the new one to get back to where they were then POOF, it's also gone.

> 
> In theory the admin could add a finalizer to a namespace to ensure it doesn't get deleted mid-troubleshooting, but that doesn't feel like the right user experience.

I wouldn't expect them to have to go around the broker to solve this.

> 
> Would it make sense to only keep the oldest one, and delete any newer namespaces that fail in the same way? Or maybe only keep the oldest and newest, but none in between? Or something else?

This is certainly an interesting approach. I'm seeing 2 new attributes: `FirstFailedSandbox` and `LastSandbox`. The former will contain the value of the sandbox that failed or empty if no failure has occurred. The latter will contain the last sandbox run. My first iteration had a list of sandboxes with the oldest being the first one in the list, not sure if we need to keep that feature or not.

@shawn-hurley thinks just keeping the first failed sandbox would suffice (see https://github.com/openshift/ansible-service-broker/pull/1130#issuecomment-439970003) but I think we'll need at least the first failed one and the latest.

jmrodri:
Not going to happen, closing.

PR #1146, opened 2018-12-11
Fix caBundle population for 4.0 deploys

djwhatle:
 - Make changes to ansible role so that broker can be deployed onto OpenShift 4.0 and `clusterservicebroker` resource will contain correct `caBundle: <blob>` section which is required for successful communication with the broker.

A build of this is currently pushed to _docker.io/djwhatle/asb-operator:4.0-svcat-cert_

### To test:
 - Create 4.0 cluster including OLM
 - Create `kube-service-catalog` namespace
 - Create Service Catalog subscription [link](https://github.com/djwhatle/stuff/blob/master/olm-svcat-brokers-testing/svcat-subscription.yaml)
 - Create osb-operators configmap [link](https://github.com/djwhatle/stuff/blob/master/olm-svcat-brokers-testing/osb-operators.configmap.upstream.4.0-fixed.yaml) (references my custom image above)
 - Create osb-operators catalogsource [link](https://github.com/djwhatle/stuff/blob/master/olm-svcat-brokers-testing/osb-operators.catalogsource.yaml)
 - Create `automation-broker` namespace
 - Create asb subscription [link](https://github.com/djwhatle/stuff/blob/master/olm-svcat-brokers-testing/asb-subscription.yaml)
 - Create asb CR [link](https://github.com/djwhatle/stuff/blob/master/olm-svcat-brokers-testing/asb-cr.yaml
)
 - Verify that `oc get clusterservicebrokers -o yaml` contains a `caBundle` with b64 content, and that the broker logs contain 200 responses like below

```
10.128.0.48 - - [11/Dec/2018:19:20:09 +0000] "GET /osb/v2/catalog HTTP/1.1" 200 373935
```

 - `oc get clusterservicebrokers` should return a "Ready" status
```
[dwhatley@precision-t olm-svcat-brokers-testing]$ oc get clusterservicebrokers
NAME                URL                                              STATUS    AGE
automation-broker   https://broker.automation-broker.svc:1338/osb/   Ready     17m
```

### DIFF
--- a/ansible_role/tasks/main.yml
+++ b/ansible_role/tasks/main.yml
@@ -74,6 +74,8 @@
     - name: broker.route.yaml
       apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-client.secret.yaml
+    - name: broker-service-ca-bundle.configmap.yaml
+      apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-auth.secret.yaml
       apply: "{{ (broker_config.broker.auth[0].enabled | default(False)) | bool }}"
     - name: broker.deployment.yaml
--- a/ansible_role/tasks/main.yml
+++ b/ansible_role/tasks/main.yml
@@ -75,7 +75,7 @@
       apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-client.secret.yaml
     - name: broker-service-ca-bundle.configmap.yaml
-      apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
+      apply: "{{ ('servicecertsigneroperatorconfigs.servicecertsigner.config.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-auth.secret.yaml
       apply: "{{ (broker_config.broker.auth[0].enabled | default(False)) | bool }}"
     - name: broker.deployment.yaml
--- a/ansible_role/tasks/main.yml
+++ b/ansible_role/tasks/main.yml
@@ -75,7 +75,7 @@
       apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-client.secret.yaml
     - name: broker-service-ca-bundle.configmap.yaml
-      apply: "{{ ('servicecertsigneroperatorconfigs.servicecertsigner.config.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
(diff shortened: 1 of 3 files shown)


ansible_role/tasks/main.yml
```diff
@@ -74,6 +74,8 @@
     - name: broker.route.yaml
       apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
     - name: broker-client.secret.yaml
+    - name: broker-service-ca-bundle.configmap.yaml
+      apply: "{{ ('route.openshift.io' in lookup('k8s', cluster_info='api_groups')) | bool }}"
```
djzager: Is there a more specific API (like one provided by the cert-signer operator) that we could look for?

ansible_role/templates/broker.servicecatalog.yaml
```diff
@@ -37,11 +37,11 @@ spec:
   caBundle: {{
     lookup(
       'k8s',
-      kind='Secret',
+      kind='ConfigMap',
```
djzager: If there is a more specific API we could find from the cert-signer operator, we should use that to determine if the `caBundle` should come from the secret or configmap.

djzager:
Removed the lgtm and approved flags so we can discuss before it gets auto-merged.

ansible_role/tasks/main.yml
djwhatle: Good point, I was hoping you would chime in on this. I will check on that.

jmontleon: Will there ever not be a cert signer operator running in new versions? I think it's always there in Openshift?

djzager: My concern is when this runs in Kubernetes or any older version of OpenShift where nothing will happen when you create the configmap and the resulting `(Cluster)ServiceBroker` we create will try and use values from a configmap that will never be filled in.

jmontleon: This version of the role is for 4.0 right? We have branches in release-1.3, release-1.2, etc. with roles appropriate to those versions, correct?

Having said that, for kubernetes, perhaps:
```
ok: [localhost] => (item=certificates.k8s.io) => {
    "certificates.k8s.io": "VARIABLE IS NOT DEFINED!",
    "item": "certificates.k8s.io"
```

jmontleon: Maybe
```
ok: [localhost] => (item=servicecertsigner.config.openshift.io) => {
    "item": "servicecertsigner.config.openshift.io",
    "servicecertsigner.config.openshift.io": "VARIABLE IS NOT DEFINED!"
}

```

djwhatle: @jmontleon @djzager updated to look for `servicecertsigner.config.openshift.io` apigroup

ansible_role/templates/broker.servicecatalog.yaml
djwhatle: updated

PR #1150, opened 2018-12-17
Update operator playbook to use IMAGE from environment

djzager:
The operator playbook will now enforce `broker_name`, `broker_namespace`, and `broker_image` to be defined with `broker_image` being captured from the operator's environment.

### DIFF
--- a/ansible_role/playbooks/operator.yml
+++ b/ansible_role/playbooks/operator.yml
@@ -5,9 +5,22 @@
   become: false
   gather_facts: false
   connection: local
-  roles:
-  - role: automation-broker
   vars:
     state: present
     broker_name: "{{ meta.name }}"
     broker_namespace: "{{ meta.namespace }}"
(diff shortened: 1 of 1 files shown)


ansible_role/playbooks/operator.yml
```diff
@@ -5,9 +5,17 @@
...
     broker_name: "{{ meta.name }}"
     broker_namespace: "{{ meta.namespace }}"
+    broker_image: "{{ lookup('env', 'IMAGE') }}"
+  tasks:
+    - assert:
+        that:
+          - broker_name is defined
+          - broker_namespace is defined
+          - broker_image is defined
+        msg: broker_name broker_namespace broker_image must be defined
```
eriknelson: I would change this to indicate that the `IMAGE` environment variable must provide the broker's image name, since `broker_image` is an internal variable and the env var is the public interface.

djzager: Done

PR #1151, opened 2018-12-17
Add ASB operator deployment manifest assets.

djwhatle:
Move ASB operator manifests from my *[stuff](https://github.com/djwhatle/stuff/tree/master/olm-svcat-brokers-testing)* repo to here, and add additional resources to create the namespace and clusterrolebinding. This _is not_ what we want the final state of the `manifests` directory to be, but it is a step towards centralizing the location of these assets.

Still need to test if `oc create -f manifests` works to get ASB running on a 4.0 cluster with OLM. I already know that the CR can't be created immediately since the CRD won't be registered until some amount of time after the ASB Operator Subscription is created.

Note that this PR separates the previous `osb-operators` configmap into individual CatalogSources/ConfigMaps for both ASB and TSB operators.

#### To test:
 - From a fresh 4.0 cluster using tagged installer release v0.7.0 or newer, (or one completely cleaned of the old `osb-operators` ASB/TSB combined CatalogSource / automation-broker namespace), run `oc create -f ansible-service-broker/manifests`.
 - The first time you run the above command, you should get a few "failed to create resource" messages. This is expected. It takes some time for OLM to create and execute the InstallPlan that will install the `automationbroker` CRD onto the cluster.
 - Wait a few seconds and run the same command `oc create -f ansible-service-broker/manifests` again, this time you should see the remaining resources create successfully, and you should get an `automation-broker` pod running after some time.


manifests/0.1.0/01-asb-catalogsource-configmap.yaml
```diff
@@ -0,0 +1,364 @@
...
+                    name: automation-broker-operator-alm-owned
+                    labels:
+                      name: automation-broker-operator-alm-owned
+                  spec:
+                    serviceAccountName: automation-broker-operator
+                    containers:
+                    - name: automation-broker-operator
+                      image: docker.io/djwhatle/asb-operator:4.0-fix
+                      imagePullPolicy: IfNotPresent
+                      env:
```
djzager: Should go ahead and add the `IMAGE` environment variable

manifests/0.1.0/01-asb-catalogsource-configmap.yaml
```diff
@@ -0,0 +1,364 @@
...
+                template:
+                  metadata:
+                    name: automation-broker-operator-alm-owned
+                    labels:
+                      name: automation-broker-operator-alm-owned
+                  spec:
+                    serviceAccountName: automation-broker-operator
+                    containers:
+                    - name: automation-broker-operator
+                      image: docker.io/djwhatle/asb-operator:4.0-fix
```
djzager: Not sure what we need to do in order to make this use `docker.io/ansibleplaybookbundle/openshift-ansible-service-broker:latest` and work.

djwhatle: Thanks for catching this, I meant to use docker.io/ansibleplaybookbundle/openshift-ansible-service-broker:latest here. Should work, I pushed an updated image earlier today with all of our cert fixes.

manifests/0.1.0/01-asb-catalogsource-configmap.yaml
djwhatle: Good catch, I will add that in.

manifests/0.1.0/01-asb-catalogsource-configmap.yaml
djwhatle: Made a change to use the official operator image, good catch

manifests/01-asb-catalogsource-configmap.yaml
```diff
@@ -0,0 +1,367 @@
...
+                    serviceAccountName: automation-broker-operator
+                    containers:
+                    - name: automation-broker-operator
+                      image: docker.io/automationbroker/automation-broker-operator:latest
+                      imagePullPolicy: IfNotPresent
+                      env:
+                      - name: IMAGE
+                        value: docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest
+                      - name: WATCH_NAMESPACE
+                        valueFrom:
```
djzager: I believe we should hardcode the `WATCH_NAMESPACE` to be the namespace where we want the broker deployed into like `openshift-ansible-service-broker`

PR #1153, opened 2018-12-19
Make it possible to run tests only

jmontleon:


ansible_role/defaults/main.yml
```diff
@@ -164,3 +164,5 @@ quay_registry:
 
 # conditionally set later based on Cluster configuration
 default_registries: []
+
+tests_deploy_broker: true
```
djzager: Looking at the failure from travis. I think it's more correct to take this out of this file and add it to the vars of the ansible_role/playbooks/test.yml

jmontleon: Understood. I'll do so in a few and see if that gets it cleaned up.

PR #1166, opened 2019-01-17
Operator SDK Alignment

djzager:
The purpose of this change is to bring the ASB Operator in line with what a typical operator-sdk project looks like. While we are at it we are abandoning the ASB-apb and only supporting the operator concept. The goal of this change is to drastically simply the execution and help ensure idempotence.

PR #1169, opened 2019-01-23
Add OLM support to new operator

eriknelson:
This PR continues #1166, adding requirements for deployment via OLM. The key differences between this and the previous operator is:
* APB support has been dropped, in favor of a role that is specifically designed for ansible operator.
* The operator's ansible is responsible for provisioning the application's SA and RBAC, rather than delegating this to OLM.
* The public interface has been simplified with the expectation that users will express their configuration via a limited set of values in their CRs. Legacy operating modes such as basic auth and an etcd based dao have been deprecated.

### DIFF
--- a/operator/deploy/olm-catalog/0.1.0/automationbroker.crd.yaml
+++ b/operator/deploy/olm-catalog/0.1.0/automationbroker.crd.yaml
@@ -1,9 +1,9 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: automationbrokers.automationbroker.io
+  name: automationbrokers.osb.openshift.io
 spec:
-  group: automationbroker.io
+  group: osb.openshift.io
   names:
     kind: AutomationBroker
     listKind: AutomationBrokerList
--- a/operator/deploy/olm-catalog/0.1.0/automationbrokeroperator.v0.1.0.clusterserviceversion.yaml
+++ b/operator/deploy/olm-catalog/0.1.0/automationbrokeroperator.v0.1.0.clusterserviceversion.yaml
@@ -1,10 +1,11 @@
+---
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   name: automationbrokeroperator.v0.1.0
   namespace: placeholder
   annotations:
-    alm-examples: '[{"apiVersion":"automationbroker.io/v1alpha1","kind":"AutomationBroker","metadata":{"name":"automation-broker","namespace":"automation-broker"},"spec":{"createBrokerNamespace":"false","waitForBroker":"false"}}]'
+    alm-examples: '[{"apiVersion":"osb.openshift.io/v1alpha1","kind":"AutomationBroker","metadata":{"name":"ansible-service-broker","namespace":"openshift-ansible-service-broker"},"spec":{"createBrokerNamespace":"false","waitForBroker":"false"}}]'
 spec:
   displayName: Automation Broker Operator
   description: |
@@ -41,13 +42,13 @@ spec:
       mediatype: image/png
   installModes:
   - type: OwnNamespace
-    supported: true
+    supported: false
   - type: SingleNamespace
     supported: true
   - type: MultiNamespace
-    supported: true
+    supported: false
   - type: AllNamespaces
-    supported: true
+    supported: false
   install:
     strategy: deployment
     spec:
@@ -66,7 +67,6 @@ spec:
           - pods
           - configmaps
           - secrets
-          - serviceaccounts
           - services
           verbs:
           - "*"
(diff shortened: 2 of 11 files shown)


operator/deploy/olm-catalog/0.1.0/automationbroker.crd.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: automationbrokers.automationbroker.io
+  name: automationbrokers.osb.openshift.io
```
djzager: I had been using `operator/deploy/crds/osb_v1alpha1_ansibleservicebroker_crd.yaml` to test this out in the beginning. The CRDs in the olm-catalog (or manifests...wherever these end up living in the end) could probably be just copies from the `operator/deploy/crds` dir. Thoughts on this and the `AutomationBroker` kind?

https://github.com/openshift/ansible-service-broker/blob/master/operator/deploy/crds/osb_v1alpha1_ansibleservicebroker_crd.yaml

operator/deploy/olm-catalog/0.1.0/automationbrokeroperator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -41,13 +42,13 @@ spec:
       mediatype: image/png
   installModes:
   - type: OwnNamespace
-    supported: true
+    supported: false
```
djzager: I'm curious what the OLM supported way of watching the `openshift-ansible-service-broker` namespace is.

operator/roles/ansible-service-broker/templates/broker-config.configmap.yaml.j2
```diff
@@ -7,10 +7,37 @@ metadata:
...
+      ca_file: ''
+      bearer_token_file: ''
+      image_pull_policy: "{{ apb_pull_policy }}"
+      sandbox_role: "{{ sandbox_role }}"
+      namespace: "{{ broker_namespace | quote }}"
+      keep_namespace: {{ keep_namespace }}
+      keep_namespace_on_error: {{ keep_namespace_on_error }}
+    broker:
+{% if dashboard_redirector_host is defined %}
+      dashboard_redirector: https://{{ dashboard_redirector_host }}
```
djzager: If you want...you could simply put all of the logic for the dashboard redirector here:

```
{% if state == present %}
    dashboard_redirector: https://{{ lookup('k8s', api_version='route.openshift.io/v1', kind='Route', namespace=broker_namespace, resource_name=dashboard_redirector_route_name).spec.host }}
{% endif %}

File: operator/deploy/olm-catalog/0.1.0/automationbrokeroperator.v0.1.0.clusterserviceversion.yaml [hunk 4]
```diff
@@ -1,10 +1,11 @@
+---
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   name: automationbrokeroperator.v0.1.0
   namespace: placeholder
   annotations:
```
djzager: Read through https://docs.google.com/document/d/10SJ_k7efVMnNxv49AT-jTdvnKF_O0R1CFdT_-KOW1Qg/edit and I want to state what I believe to be the relevant configuration (primarily CSV) for the ansible-service-broker(-operator) to be installed in such a way that the Operator is installed in `openshift-operators` (which to my understanding is the preferred method for 4.0) watching a single namespace `openshift-ansible-service-broker`. This way if I'm wrong it should be easy to point out what I'm not understanding:

1) Annotations on CSV:

```
metadata:
  name: automationbrokeroperator.v0.1.0
  namespace: placeholder
  annotations:
    olm.targetNamespaces: openshift-ansible-service-broker
    olm.operatorNamespace: openshift-operators
    olm.operatorGroup: openshift-production
```

2) Install Modes:

```
spec:
  installModes:
  - type: OwnNamespace
    supported: false
  - type: SingleNamespace
    supported: true
  - type: MultiNamespace
    supported: false
  - type: AllNamespaces
    supported: false
```

3) Setting WATCH_NAMESPACE environment variable for operator built with operator-sdk:

```
env:
- name: WATCH_NAMESPACE
  valueFrom:
      fieldRef:
          fieldPath: metadata.annotations['olm.targetNamespaces']
```

operator/deploy/olm-catalog/0.1.0/automationbroker.crd.yaml
djzager: I hadn't considered `ansible-service-broker` being an instance of `AutomationBroker`. Reading back through the PR this language makes sense. So I agree with both:

1. Get rid of the `operator/deploy/crds` directory for now to avoid confusion letting the manifest directory be the source of truth
1. Use `AutomationBroker`

operator/roles/ansible-service-broker/tasks/main.yml
```diff
@@ -5,6 +5,31 @@
   set_fact:
     api_groups: "{{ lookup('k8s', cluster_info='api_groups') }}"
 
+# reconciled_generation - the last reconciled spec generation of a given CR
+# generation - the current generation of a given CR
+# These values are useful because if the current generation value is different
+# than the reconciled_generation value, we know that the CR's spec has been
+# changed, requiring a bounce of the broker's pod to accept new configuration.
+- name: Set reconciled_generation and generation facts
```
djzager: Disclaimer: I think this comment is a good candidate for a follow-up PR.

We can put these in `vars/main.yaml` and have the first task be:

```
- name: Set important facts before execution
  set_fact:
    api_groups: "{{ lookup('k8s', cluster_info='api_groups') }}"
    pending_config_changes: "{{ generation != reconciled_generation }}"
```

operator/roles/ansible-service-broker/tasks/main.yml
```diff
@@ -5,6 +5,31 @@
...
+    }}"
+    reconciled_generation: "{{
+      lookup('k8s',
+      api_version='osb.openshift.io/v1alpha1',
+      kind='AutomationBroker',
+      namespace=broker_namespace,
+      resource_name=broker_name).status.reconciledGeneration | default('')
+    }}"
+
+- set_fact:
```
djzager: You will need a name on this to pass any linting of this we do (this will be important when we tie this operator into openshift ci).

operator/roles/ansible-service-broker/templates/broker-config.configmap.yaml.j2
eriknelson: This is ultimately conditional based on whether or not routes are available, otherwise `dashboard_redirector_host` won't be defined.

djzager: Yeah. I would like to believe that this line was blob was different when I made the comment but I can't be sure. Looks fine how it is now.

PR #1173, opened 2019-01-31
Add more operator related RBAC

eriknelson:
* Adds the CRB for OLM to grant the operator admin
* Adds the CRB for the operator to grant the broker admin
* Adds the aggregation rule for user authorization via the broker

### DIFF
--- a/operator/deploy/olm-catalog/0.2.0/admin.clusterrolebinding.yaml
+++ b/operator/deploy/olm-catalog/0.2.0/admin.clusterrolebinding.yaml
@@ -1,0 +1,14 @@
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: automation-broker-operator-admin
+roleRef:
+  name: admin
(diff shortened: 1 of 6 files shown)


operator/deploy/olm-catalog/0.2.0/admin.clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: automation-broker-operator-admin
+roleRef:
+  name: admin
+  kind: ClusterRole
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: automation-broker-operator
+  namespace: placeholder
```
djzager: personally, I think you should make this `openshift-ansible-service-broker` so that we can demonstrate this working.

eriknelson: yeah, we obviously may have to do that, hence the WIP tag. I'm just really hoping I screwed something up and this is the correct thing to do.

PR #1178, opened 2019-02-05
Fix 39 travis

jmrodri:
Travis is failing because of golint version not supporting go 1.8 anymore, also the OWNERS files were causing some of the OpenShift bots to go haywire.


jmrodri:
The APBs are still causing travis to fail so if we get to the provision step, this patch has worked.

PR #1182, opened 2019-02-11
Update operator/role, show error message on misconfigured RBAC

alaypatel07:
The operator needs admin permissions. When the operator is not run with admin permissions, the logs on operator pod should show a proper error message

### DIFF
--- a/operator/roles/ansible-service-broker/tasks/main.yml
+++ b/operator/roles/ansible-service-broker/tasks/main.yml
@@ -46,7 +46,11 @@
 
 - fail:
     msg: "Operator does not have sufficient permissions to create RBAC policy for the broker, please check the documentation."
-  when: crb_status.status is defined and crb_status.status == 403
+  when: (not crb_status.failed) and crb_status.status is defined and crb_status != 403
+
+- fail:
+    msg: "Unable to create broker-admin clusterrolebinding"
+  when: crb_status.failed
 
 - name: Set up for TLS in k8s
   import_tasks: tls_k8s.yml
--- a/operator/roles/ansible-service-broker/tasks/main.yml
+++ b/operator/roles/ansible-service-broker/tasks/main.yml
@@ -37,6 +37,17 @@
 - name:  Validate pre-requisites for {{ state }}
   import_tasks: "validate_{{ state }}.yml"
 
+- name: Set broker admin cluster rolebinding state={{ state }}
+  k8s:
+    state: "{{ state }}"
+    definition: "{{ lookup('template', 'broker-admin.clusterrolebinding.yaml.j2') | from_yaml }}"
+  register: crb_status
+  ignore_errors: yes
+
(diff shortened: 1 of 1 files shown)


operator/roles/ansible-service-broker/tasks/main.yml
```diff
@@ -3,7 +3,12 @@
 # Only want to do this once per execution of role
 - name: Get cluster api_groups
   set_fact:
-    api_groups: "{{ lookup('k8s', cluster_info='api_groups', kubeconfig=lookup('env', 'K8S_AUTH_KUBECONFIG')) }}"
+    api_groups: "{{ lookup('k8s', cluster_info='api_groups', kubeconfig=lookup('env', 'K8S_AUTH_KUBECONFIG', errors='ignore')) }}"
```
djzager: I don't believe that this one would ever fail because of RBAC. @fabianvf may know better but I don't think this every fails because of RBAC problems.

operator/roles/ansible-service-broker/tasks/main.yml
```diff
@@ -67,14 +72,24 @@
...
 
+- name: Broker admin cluster rolebinding state={{ state }}
+  k8s:
+    state: "{{ state }}"
+    definition: "{{ lookup('template', 'broker-admin.clusterrolebinding.yaml.j2') | from_yaml }}"
+  register: crb_status
+  ignore_errors: yes
+
+- fail:
+    msg: "RBAC permissions are not configured properly, please check the documentation"
```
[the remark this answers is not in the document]
alaypatel07: The way you put it LGTM.

operator/roles/ansible-service-broker/tasks/main.yml
alaypatel07: done

operator/roles/ansible-service-broker/tasks/main.yml
```diff
@@ -37,6 +37,17 @@
...
+- name: Set broker admin cluster rolebinding state={{ state }}
+  k8s:
+    state: "{{ state }}"
+    definition: "{{ lookup('template', 'broker-admin.clusterrolebinding.yaml.j2') | from_yaml }}"
+  register: crb_status
+  ignore_errors: yes
+
+- fail:
+    msg: "Operator does not have sufficient permissions to create RBAC policy for the broker, please check the documentation."
+  when: crb_status.status is defined and crb_status.status == 403
```
shawn-hurley: I would also make sure that we fail if crb_status is 50X errors. 

I think the best way to do this is `crb_status.status != 20X`.

alaypatel07: I agree that the playbook will fail if it hits 50X errors, but then again 50X errors are not related to RBAC right? So it would still be inappropriate error reporting.

PR #1185, opened 2019-02-12
Bug 1643303 - Fixed in bundle-lib 0.1.10

jmrodri:


vendor/github.com/automationbroker/bundle-lib/runtime/watch_pod_test.go
```diff
@@ -172,10 +171,6 @@ func TestWatchPod(t *testing.T) {
 					if newDescription != "" {
 						descriptions = append(descriptions, newDescription)
 					}
-
```
shawn-hurley: Should this actually be removed?

jmrodri: yeah I removed it because it wasn't used and lint was complaining. We would set it and did nothing with it.

PR #1198, opened 2019-03-21
fix path for molecule binary in container

fabianvf:
Makes it so the molecule binary is in the PATH in the molecule container

### DIFF
--- a/operator/build/molecule.Dockerfile
+++ b/operator/build/molecule.Dockerfile
@@ -3,9 +3,10 @@ FROM centos:7
 RUN mkdir -p /home/molecule
 
 ENV HOME=/home/molecule
-ENV PYTHONUSERBASE=${HOME}/.bin
-RUN mkdir -p ${PYTHONUSERBASE}
-ENV PATH="${HOME}/.bin:${PATH}"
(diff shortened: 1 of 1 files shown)


operator/build/molecule.Dockerfile
```diff
@@ -3,9 +3,10 @@ FROM centos:7
 RUN mkdir -p /home/molecule
 
 ENV HOME=/home/molecule
-ENV PYTHONUSERBASE=${HOME}/.bin
-RUN mkdir -p ${PYTHONUSERBASE}
-ENV PATH="${HOME}/.bin:${PATH}"
+ENV PYTHONUSERBASE=${HOME} \
+    PATH="${HOME}/bin:${PATH}"
```
jmrodri: You sure this is right? When I did `pip install --user molecule==2.20.0.0a2` on my machine it put it in `$HOME/.local/bin`

fabianvf: Yeah, the overridden PYTHONUSERBASE makes it just $HOME/bin

fabianvf: Yeah, the overridden PYTHONUSERBASE makes it just $HOME/bin

PR #1203, opened 2019-04-01
Update Dockerfile-openshift

sosiouxme:
This needs to be usable from both OKD and OCP.
OKD now should have same RPM sources as OCP so this should work.

@jmontleon @stevekuznetsov -- discuss?

### DIFF
--- a/build/Dockerfile-openshift
+++ b/build/Dockerfile-openshift
@@ -6,11 +6,8 @@ ENV USER_NAME=ansibleservicebroker \
     BASE_DIR=/var/lib/ansibleservicebroker
 ENV HOME=${BASE_DIR}
 
-ADD rpms/rpms/* /tmp/
-
-RUN yum -y update \
- && yum -y install epel-release centos-release-openshift-origin \
- && yum -y install origin-clients /tmp/ansible-service-broker*.rpm \
+RUN yum -y install --setopt=skip_missing_names_on_install=False \
+    atomic-openshift-clients ansible-service-broker-container-scripts ansible-service-broker \
  && yum clean all
 
 RUN mkdir -p ${BASE_DIR} ${BASE_DIR}/etc /var/run/asb-auth \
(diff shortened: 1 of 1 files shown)


sosiouxme:
The composes got updated today and also a bug in the repo sync was fixed so hopefully all necessary RPM content is now available for building.
We're not really building 4.0 any more though so perhaps this should move to master or 4.1 branch? Up to you how best to proceed.

build/Dockerfile-openshift
```diff
@@ -6,11 +6,8 @@ ENV USER_NAME=ansibleservicebroker \
     BASE_DIR=/var/lib/ansibleservicebroker
 ENV HOME=${BASE_DIR}
 
-ADD rpms/rpms/* /tmp/
-
-RUN yum -y update \
- && yum -y install epel-release centos-release-openshift-origin \
- && yum -y install origin-clients /tmp/ansible-service-broker*.rpm \
+RUN yum -y install --setopt=skip_missing_names_on_install=False \
+    atomic-openshift-clients ansible-service-broker-container-scripts ansible-service-broker \
```
sosiouxme: This could be updated to
```suggestion
    openshift-clients ansible-service-broker-container-scripts ansible-service-broker \
```
... but it shouldn't matter because the new RPM `Provides:` the old name.

sosiouxme:
prow / cluster weirdness
/retest

shawn-hurley:
No package openshift-clients available

@sosiouxme

sosiouxme:
yeah, noticed that.

@stevekuznetsov sorry I'm not familiar enough with the CI config to know how repos get configured in... any idea why `openshift-clients` RPM isn't available to this build? We definitely sync that out in the rhel-server-ose-rpms repo.

alternatively, should we consider just basing the image on ose-cli to get the client?

smarterclayton:
Base on cli if you want the CLI binary.

sosiouxme:
`No package ansible-service-broker-container-scripts available.`
Thanks, yum, that would have been nice to know back when `openshift-clients` was listed as unavailable.
So. Why aren't the RPMs that OCP builds and syncs to AWS... available for image builds?
@shawn-hurley could the remaining RPMs be omitted in favor of building from source in the container?

jmrodri:
@sosiouxme @shawn-hurley do we still need this PR?

sosiouxme:
@jmrodri obviously not on `release-4.0`, will reopen on master when it becomes a priority again.

PR #1204, opened 2019-04-01
Bug 1692281 - operator should deploy a service-monitor

fabianvf:
This makes the Ansible Service Broker Operator deploy a ServiceMonitor for the Ansible Service Broker, allowing it to expose metrics to prometheus.

https://bugzilla.redhat.com/show_bug.cgi?id=1692281


jmrodri:
Commit should have the Bug number in it. Change the commit to be **Bug 1692281 - operator should deploy a service-monitor**

PR #1234, opened 2019-07-01
Bug 1723422 - fix pagination url, bump bundle-lib

jmrodri:
When using APIV2 registries, if we hit the pagination sequence, the registry code would fail to get the next page and thus not return any images to the broker. If the set of images on the registry is less than the page size, things worked or if the images were listed individually in the configuration.

Fixes: BZ 1723422

### DIFF
--- a/vendor/github.com/automationbroker/bundle-lib/clients/kubernetes.go
+++ b/vendor/github.com/automationbroker/bundle-lib/clients/kubernetes.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 
 	log "github.com/sirupsen/logrus"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


jmrodri:
https://bugzilla.redhat.com/show_bug.cgi?id=1723422

vendor/github.com/automationbroker/bundle-lib/clients/kubernetes.go
```diff
@@ -209,8 +210,12 @@ func newKubernetes() (*KubernetesClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
+		config := os.Getenv("KUBECONFIG")
```
shawn-hurley: This is a different change than we were expecting right? 

can we make sure that this not going to mess something up?

jmrodri: That was the 0.2.18 change that never got pulled in. Looks like that was for apb tool support. I'll verify to make sure it doesn't affect the broker.

jmrodri: I verified that the above change is used ONLY when the broker is run *outside* of a cluster or if there is an error getting the `InClusterConfig`. I would argue if you get an error getting the `InClusterConfig` and running in a cluster, you have larger problems.

This code is also used when the `apb` tool uses `bundle-lib`.

PR #1237, opened 2019-07-17
Pass *_PROXY envvars on to ASB deployment

fabianvf:
OLM will be setting these variables on operator deployments and we need to pass them through to deployments created by the operator

### DIFF
--- a/operator/roles/ansible-service-broker/templates/broker.deployment.yaml.j2
+++ b/operator/roles/ansible-service-broker/templates/broker.deployment.yaml.j2
@@ -59,6 +59,12 @@ spec:
           env:
             - name: BROKER_CONFIG
               value: /etc/{{ broker_name }}/config.yaml
+{% for proxy_envvar in ["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"] %}
+{% if lookup("env", proxy_envvar) != None %}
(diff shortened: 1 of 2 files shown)


operator/roles/ansible-service-broker/templates/broker.deployment.yaml.j2
```diff
@@ -59,6 +59,12 @@ spec:
           env:
             - name: BROKER_CONFIG
               value: /etc/{{ broker_name }}/config.yaml
+{% for proxy_envvar in ["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"] %}
```
jmrodri: Will this handle lowercase proxy vars? I know we had to deal with lowercase ones in bundle-lib and when I did subscription-manager.

https://github.com/automationbroker/bundle-lib/blob/master/runtime/run_bundle.go#L225-L236
https://github.com/candlepin/subscription-manager/blob/master/test/rhsm/unit/util_tests.py#L319-L320

maybe they'll always be uppercase in this scenario but I know I had to deal with them possibly being lowercase.

shawn-hurley: They are being passed specifically as uppercase from the contract with OLM

PR #1247, opened 2019-10-21
Bug 1763005 - add registry-proxy support, bump bundle-lib

jmrodri:
The registry-proxy GET /v2 returns a WWW-Authenticate header with only a
realm: .../v2/auth the APIV2Adapter simply gets a token for /v2/auth
which is not valid for registry-proxy which seems to require a scoped
token.

In this PR, we added a new RegistryProxyAdapter that now requires the
apbs to be listed in the configuration. For each of those images listed,
we will add them to the scope when we ask for the token. This allows the
broker access to fetch the apb images.

Also, added a test to ensure registry proxy errors when no images are defined
and added a unit test for getTokenWithScope.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -56,7 +56,7 @@
   revision = "bf1c2b52e55d67b4c3a917daaadee9b86a6c66e0"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


jmrodri:
The bundle-lib PR https://github.com/automationbroker/bundle-lib/pull/184

PR #1250, opened 2019-10-31
Add alert to fire when broker is installed.

jmrodri:


operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.3/image-references
```diff
@@ -0,0 +1,12 @@
+kind: ImageStream
```
joelanford: I'm not super familiar with the OLM manifest directories, so I'm sure this is right, but just in case...

Should this file have a `.yaml` extension?

jmrodri: Honestly I don't know :) I copied it from the 4.2 directory which has the same file with no `.yaml` extension https://github.com/openshift/ansible-service-broker/tree/master/operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.2

jmrodri: 4.1 is the same https://github.com/openshift/ansible-service-broker/tree/master/operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.1

PR #1256, opened 2019-11-08
Allow ASBO metrics to be scraped.

jmrodri:


jmrodri:
image (2)

operator/roles/ansible-service-broker/templates/broker-operator.monitor-role.yaml.j2
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  # TODO this should be a clusterrole
```
djzager: I don't see why this needs to be a clusterrole. Maybe remove the comment if it's not needed?

operator/roles/ansible-service-broker/templates/broker-operator.monitor-rolebinding.yaml.j2
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
djzager: It's weird to me that `prometheus-k8s` doesn't have permission to view all of this stuff across the whole cluster.

PR #1257, opened 2019-11-14
Bug 1772392: Add olm.skipRange to all CSVs and art.yaml

fabianvf:
Adds skipRanges to the CSVs and art.yaml, which should allow for proper upgrades across versions.

https://bugzilla.redhat.com/show_bug.cgi?id=1772392

### DIFF
--- a/operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.1/openshiftansibleservicebroker.v4.1.0.clusterserviceversion.yaml
+++ b/operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.1/openshiftansibleservicebroker.v4.1.0.clusterserviceversion.yaml
@@ -9,7 +9,7 @@ metadata:
     description: "OpenShift Ansible Service Broker is an implementation of the [Open Service Broker API](https://github.com/openservicebrokerapi/servicebroker)"
     capabilities: "Seamless Upgrades"
     containerImage: "quay.io/openshift/origin-ansible-service-broker"
-    olm.skipRange: ">=4.1.0 <4.2.0"
+    olm.skipRange: "<4.1.0"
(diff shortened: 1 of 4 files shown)


operator/deploy/olm-catalog/openshift-ansible-service-broker-manifests/4.1/openshiftansibleservicebroker.v4.1.0.clusterserviceversion.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     description: "OpenShift Ansible Service Broker is an implementation of the [Open Service Broker API](https://github.com/openservicebrokerapi/servicebroker)"
     capabilities: "Seamless Upgrades"
     containerImage: "quay.io/openshift/origin-ansible-service-broker"
+    olm.skipRange: ">=4.1.0 <4.2.0"
```
cuipinghuo: For this field I think it should be "<4.1.0". 
For v4.1.0, it should skip all the version below 4.1.0. Per my understanding, ">=4.1.0 <4.2.0" will skip the version between 4.1.0 and 4.2.0. This means 4.1.0 could not upgrade to 4.1.z( z>0). @fabianvf

fabianvf: Fixed, thanks @cuipinghuo!

cuipinghuo: @fabianvf thanks for the quick respond. we also need tsb to be fixed.

PR #1261, opened 2019-12-11
Bug 1782058: Allow ASBO to access endpoints

jmrodri:
also remove TODO comment left over from previous PR
https://github.com/openshift/ansible-service-broker/pull/1256/files#r345233931

PR #1263, opened 2019-12-11
fix CI builds (#1262)

jmrodri:


jmrodri:
Using admin privileges to merge this to fix the builds since the main branch seems to be broken and PRs aren't building. I did it with master and now PR #1261 is building.

PR #1273, opened 2020-03-02
Fix permissions in the Dockrefiles of /etc/passwd [master]

camilamacedo86:
- Fix permissions in the Dockrefiles of /etc/passwd


camilamacedo86:
Closed it shows no longer required,

PR #1275, opened 2020-03-04
fix CI

jmrodri:


estroz:
`pip`/python version problems in the new parent image, might need to specify `pip3`.

jmrodri:
Replaced by PR #1278

PR #1292, opened 2020-06-03
Replace whitelist/blacklist terminology by allowlist/denylist

thiagoalessio:
I've seen [rails](https://github.com/rails/rails/pull/33681) and [elixir](https://github.com/elixir-lang/elixir/pull/10078/rails#33681) doing this, and would like to suggest the same for all our openshift repos.


adambkaplan:
The Ansible service broker - like all other OpenShift service brokers - is deprecated. With the release of OpenShift 4.5 (scheduled for GA in the coming weeks) all service brokers and the Service Catalog itself will be removed [1].

I appreciate the sentiment behind this PR, but given that this component will not exist in OpenShift moving forward it's hard to see any further action being taken.

[1] https://docs.openshift.com/container-platform/4.4/applications/service_brokers/installing-service-catalog.html

thiagoalessio:
@adambkaplan makes sense! thanks for the feedback.

